PR #52, opened 2020-06-09
Redefining `find_package` breaks vcpkg (because it does the same thing)

photex:
Howdy!

This is not a fully tested patch yet because I wanted to start the conversation before taking much more time on it.

Currently the eCAL build redefines `find_package` as a macro which then breaks if you try and build using vcpkg. This is due to the fact that vcpkg is also redefining the common cmake functions for finding dependencies and adding targets.

My proposed solution was to just rename the eCAL internal macro to `ecal_find_package`, which as you can see here, ends up touching a lot of build scripts.

Ultimately I'd like to see eCAL get added to vcpkg and I figure this would be the first step.


Kerstin-Keller:
Hi @photex, 

thanks for opening this issue. It'd be nice to have eCAL provided by vcpkg :smile:

I guess it's ok to provide a macro in ordner not to mess with the original `find_package` command, however we need to make sure to define it as a regular `find_package` in [`eCALConfig.cmake.in`](https://github.com/continental/ecal/blob/master/cmake/eCALConfig.cmake.in)

```
macro(ecal_find_package)
    find_package(${ARGV})
endmacro()
```

Such that all samples can still be build against a regular eCAL installation .

photex:
Just so I understand, I should add this macro to this file? I didn't modify the file as part of this PR so far.

Kerstin-Keller:
Yes, exactly. The idea is, that the examples can be build as part of the eCAL build. But also they should be able to build if someone has an eCAL installation on their PC (or eCAL package provided by vcpkg, conan or put your favorite package manager here) and they copy one of the samples as a starting point.

Since the samples also use `ecal_find_package` the command also has to be known.

However I just realize that this does not make sense, as `ecal_find_package(eCAL)` has to be known *before* finding eCAL, so it makes no sense to define it in `eCALConfig.cmake`.
I will think about it.

photex:
We definitely don't need to consider `ecal_find_package` to be an absolute solution either. I'm willing to take another approach too.
If you have other ideas I'd be happy to try them out.

rex-schilasky:
@photex I will close this outdated PR for now. Kerstin will return soon and we can try to solve this vcpkg issue beginning 2021.

Issue #51, opened 2020-06-08
eCAL_Sub_Receive declaration appears to be incorrect.


photex:
Howdy!

https://github.com/continental/ecal/blob/master/ecal/core/include/ecal/cimpl/ecal_subscriber_cimpl.h#L109

I believe that this should declare the `buf_` parameter as `void**` and not `void*`. There isn't a problem calling this and passing a reference to a `void*` in C or C++; but when generating the FFI interface for other languages it's going to cause troubles.
In my case, the code generated by [bindgen](https://rust-lang.github.io/rust-bindgen/) prevented me from using this function. 

I'm happy to submit a PR.

Cheers,
Chip

Kerstin-Keller:
Very cool that you're trying to create a Rust binding!

I agree, that using `void*` is not a very good interface, `char*` or `unsigned char*` would have probably been a better solution, or even `uint8_t` (are these type available in pure C?)

However, we can't just change it to `void**` in this case. First, also the implementation will have to be changed. 
Also the eCAL interface is commited to being stable, changing the type here will most likely cause backward / binary incompatibility issues.
If necessary, we can provide an additional function with the required interface change.

However, the idea is, that the user provides allocated memory of length len, and eCAL will fill that memory with data. So I am not sure if `void**` is necessary?

```
 * @param       handle_       Subscriber handle. 
 * @param [out] buf_          Buffer to store the received message content.
 * @param       buf_len_      Length of the receive buffer or ECAL_ALLOCATE_4ME if
 *                            eCAL should allocate the buffer for you (see eCAL_FreeMem).
 * @param [out] time_         Time from publisher in us.
 * @param       rcv_timeout_  Maximum time before receive operation returns (in milliseconds, -1 means infinite).
```

What exactly are the problems that `bindgen` encounters?

photex:
You're right and I think the problem then is just one particular use-case because it is incorrect to declare it as `void*` when you use the `ECAL_ALLOCATE_4ME` option. Even in C or C++ this will fail to work unless you pass a `void**` which is what the samples and README demonstrates: https://github.com/continental/ecal/blob/master/samples/c/minimal/minimal_rec/src/minimal_rec.c#L41 

The `CopyBuffer` implementation specifically short-circuits if you just pass the null pointer instead of the reference to the null pointer: https://github.com/continental/ecal/blob/master/ecal/core/src/ecalc.cpp#L32

What's the better way to address this then? Since I started the PR I'm happy to continue with another approach.

[...]

PR #53, opened 2020-06-09
buf_ parameter should be `void**`

photex:
As mentioned in #51


photex:
Closing this PR after our discussions in #51 :+1:

PR #60, opened 2020-07-13
Configuration Overwrite Keys Application Fix and VS2019 Build Fix

johngladp:
Hi!

I have been having a good look at the eCAL library and experimenting with this library on Windows 10 with Visual Studio 2019 (open folders) as the build environment.  I have made a few minor enhancements that facilitate using this impressive library in this environment and I would like to offer these for merging back upstream.

The first enhancement is a workaround for the following Visual Studio 2019 (16.6.2) cmake integration bug:
https://developercommunity.visualstudio.com/content/problem/1085862/cmake-build-reports-error-when-there-isnt-one.html
The file "thirdparty/cmake_functions/git/git_revision_information.cmake" uses the word "error" in a diagnostic message which is unconditionally emitted.  Whilst this has been reported to Microsoft, who have triaged the issue, it prevents successful builds so the workaround is necessary on current and older Visual studio releases.

The second enhancement relates to the application of configuration overwrite keys to the default eCAL configuration:
Presently the overwrite keys are only applied if an ini file is successfully opened.  In the scenario where an application is calling the "int Initialize(std::vector<std::string> args_, const char *unit_name_, unsigned int components_)" function and specifying an empty string for the "--ecal-ini-file" option, the ini file fails to load as expected and the default configuration is used as expected; however, the overwrite keys are subsequently ignored.  This use-case is for when an application configures itself and does not use an ini file on disk.  The enhancement provided in this pull request re-orders the "CConfigImpl::AddFile" function such that, regardless of the success or failure of loading an ini file, overwrite keys are always applied to the configuration.  It seems prudent and a light-touch to alter the behaviour in this way as opposed to attempting to handle a non-ini-file use-case in a different way more invasive way.

The last enhancement is a minor adjustment to the warning message emitted in the "CConfigImpl::AddFile" function.  This message has been re-directed via the eCal logging infrastructure rather than being emitted to the console (which may not be accessible or visible depending on the application).


johngladp:
Double checked the logging, the global logging isn't initialised when the ini file is loaded so the PR is incorrect so I closed the PR until I can correct this error on my part.

rex-schilasky:
Hi John, thank you for that nice issue report. Great ! If you setup a new PR we can merge this easily.

johngladp:
@rex-schilasky I have squashed the commits and tested the changes.  The new PR is: #61

rex-schilasky:
Thank you .. CI is ongoing. I reviewed your changes and it's all fine. Will merge soon ...

Issue #59, opened 2020-07-08
ECAL fails when used in seperate CMake Project, TinyXML2 not properly packaged with cpack?


JeremyBYU:
I am attempting to use ECAL inside of a separate C++ project with CMake. I have been able to successfully do this with ECAL 5.5.2.  I am now trying to update and use 5.6.0. Inside my cloned `ecal` repo I have checked out the new release (tag) version and have updated all submodule dependencies.  Here is may CMake configuration when building `ecal` for completeness:

```
cmake .. -DCMAKE_BUILD_TYPE=Release -DECAL_THIRDPARTY_BUILD_PROTOBUF=OFF -DHAS_QT5=OFF -DBUILD_DOCS=OFF -DECAL_THIRDPARTY_BUILD_CURL=OFF -DECAL_THIRDPARTY_BUILD_HDF5=OFF -DBUILD_PY_BINDING=ON
make -j20
cpack -G DEB
sudo dpkg -i _deploy/eCAL-*
```

There are **no** errors building or installing ecal. However when attempting to now use Ecal in another cmake project (toally different directory) I get this error. Note the error is about not finding tinyxml target. This is a new dependency that it seems is not being bundled or found correctly when using `cpack`.
```
CMake Error at /usr/lib/cmake/eCAL/eCALTargets-release.cmake:9 (set_property):
  set_property could not find TARGET tinyxml2::tinyxml2.  Perhaps it has not
  yet been created.
Call Stack (most recent call first):
  /usr/lib/cmake/eCAL/eCALTargets.cmake:131 (include)
  /usr/lib/cmake/eCAL/eCALConfig.cmake:60 (include)
  CMakeLists.txt:33 (find_package)


CMake Error at /usr/lib/cmake/eCAL/eCALTargets-release.cmake:10 (set_target_properties):
  set_target_properties Can not find target to add properties to:
  tinyxml2::tinyxml2
Call Stack (most recent call first):
  /usr/lib/cmake/eCAL/eCALTargets.cmake:131 (include)
  /usr/lib/cmake/eCAL/eCALConfig.cmake:60 (include)
  CMakeLists.txt:33 (find_package)

```
My guess is that something has gone wrong with "packaging" ecal using cpack. Specifically the tinyxml2 dependency. Here's an image of this directory if needed. 

![image](https://user-images.githubusercontent.com/6628085/86967006-04bc4c80-c138-11ea-88b1-5e0ccc0e84fa.png)

[...]

PR #64, opened 2020-07-22
Exclude install step of static libs (tinyxml, fineftp and googletest)…

hannemn:
… since they are only build requirements


hannemn:
This will fix issue #59 if cmake option is left to -DECAL_THIRDPARTY_BUILD_TINYXML2=ON .

Issue #77, opened 2020-09-09
Protobuf is missing in the Windows Installation


FlorianReimold:
The current MSI installer does not include Protobuf. Thus it is not possible on Windows to link agains eCAL, when using the official builds.

PR #79, opened 2020-09-10
Pack protobuf into Windows Installer (issue #77)

FlorianReimold:
Switchted Installer toolchain to Innosetup. The installer now includes Protobuf, so it is actually possible to use the installed eCAL for building and linking own software.
We should advise Users to uninstall the old WIX based .msi package first to avoid conflicts.

Note that the debug libraries are still not included in the installer. Building software in release mode however is possible, which is a huge improvement compared to the previous version.


rex-schilasky:
Hi Florian, is the debug build disabled or is this build  linking against the release sdk and finally linking corrupted binaries ?

rex-schilasky (changes requested):
Can you please adapt the windows github action (build-windows.yaml) and change line 103 from

path: ${{ runner.workspace }}/_build/_deploy/*.msi

to

path: ${{ runner.workspace }}/_build/_deploy/*.exe

and maybe rename the job to "Upload InnoSetup" :-)

rex-schilasky (approved):

FlorianReimold:
> 
> 
> Hi Florian, is the debug build disabled or is this build linking against the release sdk and finally linking corrupted binaries ?

The Setup is genereated from a Release build. Thus, the setup only contains release binaries. When linking against the eCAL Setup in a debug build, you will end up with a mixed debug/release binary. As the C++ Debug ABI and C++ Release ABI are not compatible, the applications will crash when trying to pass STL containers to eCAL:

error

Linking against the C interface (eCAL::core_c) seems to work in a debug build. I however doubt that it is possible to produce a working debug binary with that workaround, once you use protobuf.

rex-schilasky:
> > Hi Florian, is the debug build disabled or is this build linking against the release sdk and finally linking corrupted binaries ?
> 
> The Setup is genereated from a Release build. Thus, the setup only contains release binaries. When linking against the eCAL Setup in a debug build, you will end up with a mixed debug/release binary. As the C++ Debug ABI and C++ Release ABI are not compatible, the applications will crash when trying to pass STL containers to eCAL:
> 
> error
> 
> Linking against the C interface (eCAL::core_c) seems to work in a debug build. I however doubt that it is possible to produce a working debug binary with that workaround, once you use protobuf.

Okay, let's solve this in a separate issue.

PR #90, opened 2020-09-28
Add cpp class callback example

imgeorgiev:
Hello

Looking through the C++ samples you had, I found it somewhat difficult to piece together a way to get subscriber callbacks to work within a class. After some research, I found a way to with lambdas which is more efficient than std::bind and would like to share this example of how to do it with others in the eCAL community.

I tried to stick as close as possible to the other samples you had and keep mine as simple as possible :)


rex-schilasky:
Hello, first of all, thank you for your contribution ! That looks good to me. But I would not add two new sample applications person_snd_callback, person_rec_callback, but only add ONE new sample for receiving person samples using your "class/lambda callback" variant. Your person_snd_callback sample is identical with the person_snd sample, isn't it ?

So finally the new person_rec_callback sample should be organized under samples/cpp/person. Besides this, the name "person_rec_callback" is not describing your new sample app perfectly. The standard person_rec sample is using a (std::bind-) callback too .. so we need something more unique here .... maybe "person_rec_lambda" ??

What do you mean ?

imgeorgiev:
Thanks for the feedback.

I realize now that I did some unnecessary code duplication. I've renamed the example to `percon_rec_lambda_in_class` as the big deal for me was that it's a callback inside a class. I hope that's fine :)

samples/cpp/person/person_rec_lambda_in_class/src/addressbook.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+  try
+  {
+    std::cout << "Waiting for messages\n";
+    addressbook.spin();
+  }
+  catch(InterruptException& e)
+  {
+    addressbook.print();
+    eCAL::Finalize();
```
hannemn, line 110: Please move this line behind the try-catch block.

hannemn (changes requested):
Thank you very much for your contributions, @imgeorgiev ! I only have one small improvement to your code. Since eCAL::Initialize() sets global states in the process, it is best to call it exactly once in the main() function. Similarly, eCAL::Finalize() should be called exactly once at the end of the process. In your case after the try-catch block.

imgeorgiev:
@hannemn thanks, I have addressed all of your points.

Your comments got me thinking about what is the correct way to use eCAL in a class...

For sure, your comments make the code more modular in the sense that you can set up multiple classes within the same program which is cooler and more portable. However, it also introduces the issue of trying to set up a class using eCAL without having initialized eCAL beforehand. To avoid this is there any way to check if eCAL has been initialized before setting up the publishers and/or subscribers?

hannemn:
There is a global function which is called eCAL::IsInitialized(). I think this is the functionality you were looking for. It returns 1 if eCAL has been already initialized, otherweise zero.

By the way, please make sure that your code doesn't throw compiler warnings and also builds on Windows and macOS. See https://github.com/continental/ecal/runs/1208277358?check_suite_focus=true and https://github.com/continental/ecal/runs/1208277355?check_suite_focus=true

imgeorgiev:
All done :)

hannemn (approved):
Looks good to me!

PR #132, opened 2020-11-21
different static code analysis fixes

schilasky:
* std::list -> std::vector (if possible)
* new -> std::make_shared / std::make_unique

### DIFF
--- a/app/rec/rec_client_cli/src/ecal_rec_cli.cpp
+++ b/app/rec/rec_client_cli/src/ecal_rec_cli.cpp
@@ -344,7 +344,7 @@ int main(int argc, char** argv)
 
   std::cout << header_ss.str();
 
-  std::shared_ptr<eCAL::pb::rec_client::EcalRecClientService> rec_service_(new EcalRecService(ecal_rec));
+  std::shared_ptr<eCAL::pb::rec_client::EcalRecClientService> rec_service_(std::make_shared<EcalRecService>(ecal_rec));
   eCAL::protobuf::CServiceServer<eCAL::pb::rec_client::EcalRecClientService> rec_service_server_(rec_service_);
 
   while (eCAL::Ok())
@@ -424,8 +424,8 @@ void UpdateEcalState()
   else
   {
     const eCAL::rec::JobStatus*            recording_job_status_(nullptr);
-    std::list<const eCAL::rec::JobStatus*> flushing_jobs_;
-    std::list<const eCAL::rec::JobStatus*> uploading_jobs_;
+    std::vector<const eCAL::rec::JobStatus*> flushing_jobs_;
+    std::vector<const eCAL::rec::JobStatus*> uploading_jobs_;
 
     // Collect The currently recording job, the flushing jobs and the jobs that are being uploaded
     for (const auto& job_status : rec_status.job_statuses_)
--- a/ecal/core/src/ecal_globals.cpp
+++ b/ecal/core/src/ecal_globals.cpp
@@ -48,7 +48,7 @@ namespace eCAL
     /////////////////////
     if (config_instance == nullptr)
     {
-      config_instance = std::unique_ptr<CConfig>(new CConfig);
+      config_instance = std::unique_ptr<CConfig>(std::make_unique<CConfig>());
       if (config_keys_)
       {
         config_instance->OverwriteKeys(*config_keys_);
@@ -62,7 +62,7 @@ namespace eCAL
     /////////////////////
     if (entity_register_instance == nullptr)
     {
-      entity_register_instance = std::unique_ptr<CEntityRegister>(new CEntityRegister);
+      entity_register_instance = std::unique_ptr<CEntityRegister>(std::make_unique<CEntityRegister>());
       new_initialization = true;
     }
 
@@ -71,7 +71,7 @@ namespace eCAL
     /////////////////////
     if(descgate_instance == nullptr) 
     {
-      descgate_instance = std::unique_ptr<CDescGate>(new CDescGate);
+      descgate_instance = std::unique_ptr<CDescGate>(std::make_unique<CDescGate>());
       new_initialization = true;
     }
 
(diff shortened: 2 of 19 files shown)


app/rec/rec_client_cli/src/ecal_rec_cli.cpp
```diff
@@ -424,8 +424,8 @@ void UpdateEcalState()
   else
   {
     const eCAL::rec::JobStatus*            recording_job_status_(nullptr);
-    std::list<const eCAL::rec::JobStatus*> flushing_jobs_;
-    std::list<const eCAL::rec::JobStatus*> uploading_jobs_;
+    std::vector<const eCAL::rec::JobStatus*> flushing_jobs_;
```
FlorianReimold, line 427: I am curious, why would you change this to an std::vector? Why did the static code analysis show this as an issue?

A Vector may have to be reallocated multiple times when back-inserting elements like in this loop. An std::list will never have to be reallocated (it offers constant time insertion at any position). Sure, a vector will offer constant time random access of any element, but the elements aren't ever accessed, only the size of the list is used. In my optinion the old code was more performant (CPU wise, it will use more memory, though). The improvement that will be faster and use less memory would be replacing the `std::list` / `std::vector` with an `int`, as the content of the list doesn't matter anyways.

ecal/core/src/readwrite/ecal_reader.cpp
```diff
@@ -695,7 +695,7 @@ namespace eCAL
     DoRegister(false);
 
     // check connection timeouts
-    std::shared_ptr<std::list<std::string>> loc_timeouts = std::shared_ptr<std::list<std::string>>(new std::list<std::string>);
+    std::shared_ptr<std::list<std::string>> loc_timeouts = std::shared_ptr<std::list<std::string>>(std::make_shared<std::list<std::string>>());
```
FlorianReimold, line 698: Isn't the output of std::make_shared already a shared_ptr of the correct type? Why would you copy-construct a new shared_ptr from that again?

ecal/core/src/readwrite/ecal_writer.cpp
```diff
@@ -620,7 +620,7 @@ namespace eCAL
     DoRegister(false);
 
     // check connection timeouts
-    std::shared_ptr<std::list<std::string>> loc_timeouts = std::shared_ptr<std::list<std::string>>(new std::list<std::string>);
+    std::shared_ptr<std::list<std::string>> loc_timeouts = std::shared_ptr<std::list<std::string>>(std::make_shared<std::list<std::string>>());
```
FlorianReimold, line 623: Isn't the output of std::make_shared already a shared_ptr of the correct type? Why would you copy-construct a new shared_ptr from that again?

ecal/core/src/ecal_globals.cpp
```diff
@@ -171,7 +171,7 @@ namespace eCAL
     {
       if (log_instance == nullptr)
       {
-        log_instance = std::unique_ptr<CLog>(new CLog);
+        log_instance = std::unique_ptr<CLog>(std::make_unique<CLog>());
```
FlorianReimold, line 174: Isn't the output of std::make_unique already a unique_ptr of the correct type? Why would you copy-construct a new shared_ptr from that again?

ecal/core/src/ecal_globals.cpp
```diff
@@ -159,7 +159,7 @@ namespace eCAL
     {
       if (monitoring_instance == nullptr)
       {
-        monitoring_instance = std::unique_ptr<CMonitoring>(new CMonitoring);
+        monitoring_instance = std::unique_ptr<CMonitoring>(std::make_unique<CMonitoring>());
```
FlorianReimold, line 162: Isn't the output of std::make_unique already a unique_ptr of the correct type? Why would you copy-construct a new shared_ptr from that again?

ecal/core/src/ecal_globals.cpp
```diff
@@ -147,7 +147,7 @@ namespace eCAL
     {
       if (timegate_instance == nullptr)
       {
-        timegate_instance = std::unique_ptr<CTimeGate>(new CTimeGate);
+        timegate_instance = std::unique_ptr<CTimeGate>(std::make_unique<CTimeGate>());
```
FlorianReimold, line 150: Isn't the output of std::make_unique already a unique_ptr of the correct type? Why would you copy-construct a new shared_ptr from that again?

FlorianReimold (changes requested):
I don't think that I agree with the static code analysis tool (If that code really is what it suggested). Maybe I am wrong, but why would you alyways write:

```
std::shared_ptr<t> = std::sdhared_ptr<t>(std::make_shared<t>())
```

instead of 

```
std::shared_ptr<t> = std::make_shared<t>()
```
?

The first version (that's the version you got suggested from the static code analysis) includes an unnecessary copy construction and looks quite redundant.

And are you sure that you can always replace `std::list` with `std::vector` without hurting the insertion performance? To my knowledge an `std::list` offers constant time O(1) insertion and removal, while an `std::vector` is reallocated every time you hit the pre-allocated limit, which *can* hurt the performance with large lists that are dynamically growing or shrinking.

app/rec/rec_client_cli/src/ecal_rec_cli.cpp
rex-schilasky: Yes that what I expected too. But the static code analysis suggests to always use vectors instead lists if there are push_backs only. Check this article [std::list vs std::vector performance](https://dzone.com/articles/c-benchmark-%E2%80%93-stdvector-vs#:~:text=std%3A%3Avector%20always%20performs,or%20inserting%20in%20the%20front) and that [one](https://www.viva64.com/en/w/v826/print/) from PVS itself.

I mean it doesnt matter that much in our use cases, but it's faster :).

ecal/core/src/readwrite/ecal_reader.cpp
rex-schilasky: It's about optimized object allocation and avoiding memory leaks in case of an exception. That is my understanding. See [here](https://www.viva64.com/en/w/v824/print/)

rex-schilasky: Now (after our discussion) I got it. Next commit will remove the unnecessary extra casts.

FlorianReimold (approved):
Looks good 👍

Issue #202, opened 2021-03-04
remove explicit compiler toolset definition in cmake batch file


rex-schilasky:
The CMake batch generation files for windows should use the installed standard compiler.

PR #205, opened 2021-03-04
Update win_make_cmake.bat

rex-schilasky:
X64 option only #202


FlorianReimold:
Can you also fix the documentation? 👍🏻

Issue #201, opened 2021-03-04
Samples are not building with default cmake command


floriangeis:
If a user has installed eCAL from a released installer, it is not possible to cmake + build the eCAL samples that use protobuf (e.g. person send). Problem seems to be the style how protobuf library will be linked by default. 

This happens for windows only

Possible solution: 
Like it is done in the hello-world example, there shall be a cmake command, like this:
![image](https://user-images.githubusercontent.com/45308813/109978977-b2e7b880-7cfe-11eb-86f9-ed34b193ef5d.png)

KerstinKeller:
Btw, also tclap cannot be found, needed by latency_send sample
```
CMake Error at cpp/latency/latency_snd/CMakeLists.txt:22 (find_package):
  By not providing "Findtclap.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "tclap", but
  CMake did not find one.

  Could not find a package configuration file provided by "tclap" with any of
  the following names:

    tclapConfig.cmake
    tclap-config.cmake
```

KerstinKeller:
With newer CMake version, you also need a mandatory `cmake_minimum_required(VERSION 3.??)` on top of every top-level CMake file, which is not present in `samples/CMakeLists.txt`.

KerstinKeller:
When building the samples with `-DCMAKE_FIND_PACKAGE_PREFER_CONFIG=ON` it builds. I think this should be enabled, by default, and always, but I am not sure if it has implications on the eCAL build itself.
If not, we can set in in the samples CMakeLists.txt file.

PR #212, opened 2021-03-16
added cmake minumim required version and protobuf static lib link (#201)

schilasky:


FlorianReimold:
Tried to compile person_snd on Windows 10 and Ubuntu 18.04 (CMake 3.10). Both worked out of the box.

FlorianReimold:
Additional Note: CMake 3.5 (from Ubuntu 16.04) fails to compile the samples. I used eCAL 5.8.3 from the PPA for that. Maybe we should state CMake 3.10 as minimum required version instead of 3.0. After Upgrading CMake to 3.13 (also from the PPA), person_snd compiles without issues.

FlorianReimold:
Looks good. When I now try to compile person_snd with stock CMake on Ubuntu 16.04, I get a proper error message.

```
CMake Error at CMakeLists.txt:19 (cmake_minimum_required):
  CMake 3.10 or higher is required.  You are running version 3.5.1


-- Configuring incomplete, errors occurred!
```

After upgrading CMake to 3.13, CMaking works fine. However the CMakeLists.txt still lacks the C++14 compiler setting:

``` cmake
set(CMAKE_CXX_STANDARD 14)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
```

Do we want to include that as well? It is not need for Ubuntu 18.04 (gcc 7) any more, but gcc 5 will complain about the `auto` keyword, as it is set to c++98 by default. For that reason, the [Hello World example](https://continental.github.io/ecal/getting_started/hello_world.html#hello-world-publisher) states to explicitely set the C++ standard in the CMakeLists.txt.

rex-schilasky:
@FlorianReimold thanks for testing. I would not add extra lines to support Ubuntu 16.04. This can be skipped for that specific improvement to build samples standalone.

PR #216, opened 2021-03-17
Samples use PREFER CONFIG mode instead of setting the Protobuf variable.

Kerstin-Keller:


FlorianReimold:
Works perfectly on Windows 10 and Ubuntu 20.04 (tested person_snd only).

PR #233, opened 2021-03-24
Update setup.rst

rex-schilasky:
Wouldn't "Installation" fit better for this section ?


FlorianReimold:
Yes, it would. I think "Installing eCAL" would fit even better. How about that?

FlorianReimold (approved):

Issue #236, opened 2021-03-25
Feature request: add "-C" option to "ecal_mon --proto | --echo" (based on inner source issue 1565)


rex-schilasky:
**Is your feature request related to a problem? Please describe.**
I use 'ecal_mon --proto SomeTopic' often to check ecal messages, now I'd like to request an additional option of "-C" to `ecal_mon --proto`, so that when I use 
```
ecal_mon --proto SomeTopic -C 2
```
ecal_mon only receive **2** messages then exit.

This "-C" option will be similar to `ping` or `mosquitto_sub`'s "-C" options.

in `ping`:
>Usage: ping [-c count]

in `mosquitto_sub`:
>Usage: mosquitto_sub -t topic ...
                     [-C msg_count]
 -C : disconnect and exit after receiving the 'msg_count' messages.

**Additional context**
I'm writing some script which use `ecal_mon --proto`, the script will run commands in a sequenced way (synchronized), if `ecal_mon --proto` is used, I don't want to kill it manually but hope it can exit by itself, so I think a `"-C"` option will be very helpful in my case.

PR #237, opened 2021-03-25
Feature request: add "-C" option to "ecal_mon --proto | --echo" (#236)

schilasky:


FlorianReimold (approved):
Looks good to me. I first stumbled upon the exact (un) equality comparison `cnt != 0` with cnt being decremented in a different thread, but the way it is implemented in the thread it should work perfectly.

rex-schilasky:
> Looks good to me. I first stumbled upon the exact (un) equality comparison `cnt != 0` with cnt being decremented in a different thread, but the way it is implemented in the thread it should work perfectly.

Yes, maybe there are better way to do it but I test for all possible values of the new ```--count``` argument and it is working as expected in all cases.

FlorianReimold:
A topic with a high frequency will cause more messages to be received, as the cnt variable is only checked every 500ms. But I cannot image anybout caring about that.

rex-schilasky:
No, inside the receive thread (lambda) the ```cnt``` variable is checked and decreased with every message. That thread will stop printing out if the number of received messages reached ```cnt```. The main thread then will exit with a maximum delay of 500 ms checking the same variable of zero.

Issue #187, opened 2021-02-21
Project does not build with imported protoc target


pablorcum:
Crosscompiling eCAL may required importing the protoc target for the current host. Due to the recent additions to disable warnings for third-party lib in the main CMakeLists.txt, the eCAL build is broken for crosscompiling targets such as QNX.

Please edit the following code in the main CMakeLists.txt lines 143-146.

    target_compile_options(protoc PRIVATE
      $<$<CXX_COMPILER_ID:MSVC>:/W0>
      $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:-w>
    )

to include an if statement that checks the IMPORTED_LOCATION as shown below.

    if (protobuf_BUILD_PROTOC_BINARIES)
      target_compile_options(protoc PRIVATE
        $<$<CXX_COMPILER_ID:MSVC>:/W0>
        $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:-w>
      )
    endif()

FlorianReimold:
Hi Pablo,

I tried solving that issue in a more generic way, as I was afraid it would not only be an issue with protobuf, but with any library that one would import from another location. Thus I moved the warning-disabling-code into the ``if (NOT TARGET ...)`` statement. Can you check it out and give me feedback, if my chosen solution works for your crosscompile build?

Branch: https://github.com/continental/ecal/compare/hotfix/crosscompile_fix
Commit: https://github.com/continental/ecal/commit/42aeec58040efd3ed876d96b17bd3ee6b409c240

Regards
Florian

pablorcum:
Hi Florian, thanks for the quick fix. Unfortunately, it is still not working. I have the following configuration in my CMake toolchain file.

```
# Set protoc properties for cross-compiling
add_executable(protoc IMPORTED)
if(CMAKE_HOST_SYSTEM_NAME STREQUAL "Windows")
	set_target_properties(protoc PROPERTIES IMPORTED_LOCATION ${CMAKE_CURRENT_SOURCE_DIR}/../protoc-3.11.4-win64/bin/protoc.exe)
else()
	set_target_properties(protoc PROPERTIES IMPORTED_LOCATION ${CMAKE_CURRENT_SOURCE_DIR}/../protoc-3.11.4-linux-x86_64/bin/protoc)
endif()
set_target_properties(protoc PROPERTIES IMPORTED_GLOBAL TRUE)
```

FlorianReimold:
[...]

PR #257, opened 2021-04-27
Cross-compilation (arm64) example and build fixture, some CMake fixes

christopherbate:
Adds a set of files under `testing/contrib/cross` which allows one to quickly setup a cross-compilation environment example and test the building of eCAL using this environment. 

There are two Docker containers constructed: "target" and "host"

The "target" is a simple arm64 container (built using QEMU) used to emulate having an ARM64 system root with some pre-built binaries, for example libssl and hdf5. The exact libraries present can be changed in order to test different build scenarios/flags. 

The "host" is a basic C++ Debian environment with GCC8 aarch64 compiler and an updated Protobuf compiler installed.

The build script provided in the README executes a sequence of steps:

1. dump the "target" container's filesystem so it can be referenced by the cross toolchain
2. spin up the "host" and use the example CMake toolchain and the system root to perform the cross-compilation. 

This setup allows you to emulate the most straightforward cross compilation scenario.

Using this, I found a few items in the CMake lists which needed to change. 
1. find_path for internal dependencies - these changes should be compatible with native builds since you only want to search your source tree (from what I understand).
2. For handling hdf5, this is an existing issue with CMake's FindHDF5 script ,see also CMake gitlab [issue](https://gitlab.kitware.com/cmake/cmake/-/issues/20819) , FindHDF5.cmake will report host libraries when you want target libraries. In this case, I handle finding the library and header paths directly, but obviously there may be other types of HDF5 installs? Probably this should get factored out into a Module similar to the other modules present under `./cmake/`

Building HDF5 from source in cross-compile scenario is a little more complicated, you have to supply data type sizes, etc in the toolchain, and I didn't really try to explore that. A quick google search makes me think cross building HDF5 is probably a nightmare, so I'm not going to touch that.

3. For handling protobuf, there was an issue [earlier](https://github.com/continental/ecal/issues/187) for someone saying they did `add_executable` to import `protoc` in their `toolchain.cmake`, but that didn't work for me and seems unorthodox. It seems more canonical to handle overriding protoc in the main CMakelists.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,7 +145,20 @@ if (ECAL_THIRDPARTY_BUILD_PROTOBUF)
     )
   endif()
 
-  if (NOT TARGET protobuf::protoc)
+  if (NOT TARGET protoc_imported AND CMAKE_CROSSCOMPILING)
+    # In cross compile scenario, we don't build protoc, so we need
+    # to find and import it so downstream usage will use the host compiler.
+
+    # find_program will search the host paths if CMAKE_FIND_ROOT_PATH_MODE_PROGRAM 
+    # is configured correctly (see toolchain file)
+    find_program(protoc_path NAMES protoc REQUIRED)
+    add_executable(protoc_imported IMPORTED)
+    set_target_properties(protoc_imported PROPERTIES IMPORTED_LOCATION ${protoc_path})
+    set_target_properties(protoc_imported PROPERTIES IMPORTED_GLOBAL TRUE)
+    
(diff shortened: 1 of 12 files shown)


christopherbate:
I'm not familiar with the Github workflows, but it looks straightforward to add the steps I listed out in the Makefile to the CI/CD workflow steps. The only question would be whether that environment would support using Docker

FlorianReimold:
Github actions indeed support docker: https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action
But I am not sure whether we really need to add more builds there, we already have so many 😄

PR #292, opened 2021-05-31
Hotfix/freebsd

hannemn:


lib/ecal_utils/src/filesystem.cpp
```diff
@@ -311,12 +309,15 @@ namespace EcalUtils
...
-#if defined (__APPLE__) || defined(__FreeBSD__)
+#if defined (__APPLE__)
       int result = fcopyfile(input_fd, output_fd, 0, COPYFILE_ALL);
       return (result == 0);
 #elif defined(__QNXNTO__)
       // TODO: Find an alternative to copy files on QNX operating system
       return false;
+#elif defined(__FreeBSD__)
+      // TODO: Find an alternative to copy files on FreeBSD operating system
+      return false;
```
FlorianReimold, line 320: Hm, this solution isn't very satisfying... Maybe we should just implement a read-file -> write-file loop with pure posix calls? This would work for qnx and freebsd.

``` cpp
#include <fcntl.h>
#include <unistd.h>
#include <errno.h>
#include <sys/mman.h>

int cp(const char *to, const char *from)
{
  int fd_from = open(from, O_RDONLY);
  if(fd_from < 0)
    return -1;
  struct stat Stat;
  if(fstat(fd_from, &Stat)<0)
    goto out_error;

  void *mem = mmap(NULL, Stat.st_size, PROT_READ, MAP_SHARED, fd_from, 0);
  if(mem == MAP_FAILED)
    goto out_error;

  int fd_to = creat(to, 0666);
  if(fd_to < 0)
    goto out_error;

  ssize_t nwritten = write(fd_to, mem, Stat.st_size);
  if(nwritten < Stat.st_size)
    goto out_error;

  if(close(fd_to) < 0) {
    fd_to = -1;
    goto out_error;
  }
  close(fd_from);

  /* Success! */
  return 0;
}
out_error:;
  int saved_errno = errno;

  close(fd_from);
  if(fd_to >= 0)
    close(fd_to);

  errno = saved_errno;
  return -1;
}
```
taken from https://stackoverflow.com/questions/2180079/how-can-i-copy-a-file-on-unix-using-c

hannemn: Good idea! I will add your proposal as generic fallback solution which should also work on QNX. A dedicated method on freebsd has been already implemented.

Issue #285, opened 2021-05-25
Support GCC 11


Blutkoete:
**Is your feature request related to a problem? Please describe.**
Arch Linux recently switched to GCC 11 and a minor patch is required to compile eCAL with it. The patch only adds two includes to two files (see "Header dependency changes" at https://gcc.gnu.org/gcc-11/porting_to.html for more information) that were implicit before and need to be explicit with GCC 11. As this does not interfere with building on officially supported platforms, I'd like to ask if it is possible to add this patch.

**Describe the solution you'd like**
The patch is applied to the official eCAL repository.

**Describe alternatives you've considered**
I have no problem to apply this patch during the build process of the AUR package each time which clearly is an alternative; but applying the patch to the official code is convenient for me and saves you a minor headache later when an officially supported platform moves to GCC 11.

**Additional context**
I also have a patch changing one line to support Protobuf >= 3.15.6, but that breaks compiling against lower versions of Protobuf and so I'll keep applying it in the AUR package (it's just one line that changes, so no worries for the future).

**Patch**
```
Index: samples/cpp/multiple/multiple_snd/src/multiple_snd.cpp
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/samples/cpp/multiple/multiple_snd/src/multiple_snd.cpp b/samples/cpp/multiple/multiple_snd/src/multiple_snd.cpp
--- a/samples/cpp/multiple/multiple_snd/src/multiple_snd.cpp	(revision b160392e47800319a498b2ff30c85cb418bb794c)
+++ b/samples/cpp/multiple/multiple_snd/src/multiple_snd.cpp	(date 1621537793427)
@@ -25,6 +25,7 @@
 #include <atomic>
 #include <chrono>
 #include <thread>
+#include <memory>
 
 #define PAYLOAD_SIZE      1024
 #define PRINT_LOG            0
Index: app/rec/rec_addon_core/src/time_limited_queue.h
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/app/rec/rec_addon_core/src/time_limited_queue.h b/app/rec/rec_addon_core/src/time_limited_queue.h
--- a/app/rec/rec_addon_core/src/time_limited_queue.h	(revision b160392e47800319a498b2ff30c85cb418bb794c)
[...]

PR #293, opened 2021-05-31
Add some explicit includes to support building with GCC 11

Blutkoete:
As discussed in #285, here's the pull request.

It only adds two explicit includes that are no longer implicitly available with GCC 11.


Blutkoete:
macOS Catalina 10.15 already uses Protobuf 3.17 and would need this adaption here to build: https://github.com/Blutkoete/ecal/tree/support_protobuf_3_16

PR #298, opened 2021-06-10
temporary request and response should be released in server & client mode

zergling-rush:
When using the server and client pattern in eCal, there's a significant memory leak problem especially when big data chunks are present in the request sent by a client.

This pull request aims to fix this issue by deleting the request and response variables at the end of the "call" method.


rex-schilasky:
Thank you for reporting. Just checked it again and yes .. we have a leak here.

FlorianReimold:
Thats a really awfull memory leak. I will accept this PR to a branch and improve it further (We must also delete the response in the error case: `if (!request->ParseFromString(request_)) return -1;`)

PR #316, opened 2021-06-19
CMakeLists.txt: Change default option BUILD_SAMPLES to OFF

phongt:
sync with documentation cmake options in [1]

[1]
https://continental.github.io/ecal/development/ecal_cmake_options.html

Signed-off-by: Phong Tran <tranmanphong@gmail.com>


FlorianReimold:
You are absolutely right, the documentation and CMakeLists.txt file are not in sync regarding that option. But somehow adjusting the file to match the documentation feels wrong... I would rather like to correct the documentation 🤔.

Any comments on whether we should have samples enabled by default @Kerstin-Keller @rex-schilasky  @schilasky @hannemn?

rex-schilasky:
I think the samples should enabled by default. So after a "standard build" I have lots of samples to test different kind of features and the eCAL eco system tools.

phongt:
@FlorianReimold , @rex-schilasky  changed the doc for sync

PR #320, opened 2021-06-23
Give ECAL_HOME hint to find_program protoc

pablorcum:
Facilitates crosscompiling when eCAL is installed on development computer.


Kerstin-Keller:
I would suggest not to merge this PR as it feels non-generic.
What does your cross-compiling toolchain file look like? I suggest to specify a `CMAKE_PROGRAM_PATH` either in the toolchain file or when invoking CMake, you can use the env variable to eCAL home there if you like.

pablorcum:
ECAL_HOME is passed in as a hint, so I tested it works, if it's empty. eCAL is likely to be installed in the development computer too. Thus, we ensure that the same protoc version is used when crosscompiling. If there are other protoc.exe in the path, CMake may import a wrong protoc.exe instead of the one included in the eCAL local installation.

FlorianReimold:
`ECAL_HOME` is an environment variable that is set during installation (at least on Windows, not sure if we set / need it on Linux). Making the Cross-compile depend on an installation of the same software seems to like a bad idea.

There has to be a be better way to set the protoc executable from the outside. I don't see a reason why you would always want to cross-compile agains the same protobuf version you have installed on your machine. The reason for cross-compiling usually is to create binaries for a totally different CPU Architecture and OS. Having a match between Local-eCAL-installation-protoc and crosscompile-ecal-libprotobuf looks like pure chance to me.

pablorcum:
There is still an option so that the user sets the imported PROTOC target externally. This will only affect those standard use cases, in which the user wants an easy way to cross-compile, and expects an easy solution.

eCAL is already in the system path, so it will be found in most of the cases regardless of the HINT. However, I have heard of an user that was having issues to cross-compile because he had two different protoc.exe in his Windows path. CMake was selecting the wrong one, and the compilation would fail later due to mismatching protoc versions. The easy way to fix this was to remove the other path from the system path, and just leave eCAL/bin there. However, if we had defined eCAL as a HINT that would have been searched before the system path, increasing the probability of using a compatible protoc version for crosscompilation.

KerstinKeller:
Hi @pablorcum,
I know that it is very convenient if this "just work".
However, if making more and more implicit assumptions, from my experience it leads to more problems than it is worth having a little more convenience.
If you're crosscompiling eCAL, it is a requirement to have a `protoc` version matching the `protobuf` version that you are cross compiling for, in your host system. To implicitly search for protoc in an existing eCAL installation just feels wrong. You could just as well receive a missmatch there, too.

pablorcum:
Alright. It is fine, thanks for the justification. eCAL is added to the path at least of Windows during installation, so CMake will find protoc during cross-compiling anyway. My point was for those cases in which for any reason eCAL was installed but removed from path.

PR #322, opened 2021-06-25
Enable zero copy shared memory transport mode.

schilasky:
By default, the builtin shared memory layer is configured to make one memory copy
on the receiver side. That means the payload is copied by the internal eCAL memory pool manager
out of the memory file and the file is closed immediately after this.
The intention of this implementation is to free the file as fast as possible after reading
its content to allow other subscribing processes to access the content with minimal latency.
The different reading subscribers are fully decoupled and can access their memory copy
independently.

New publisher option to switch eCALs shared memory transport layer to zero copy mode.

If CPublisher::EnableZeroCopy is switched on no memory will be copied at all. The user message callback is
called right after opening the memory file. A direct pointer to the memory payload is forwarded
and can be processed with no latency. The memory file will be closed after the user callback function
returned. The advantage of this configuration is a much higher performance for large payloads (> 1024 kB).
The disadvantage of this configuration is that in the time when the callback is executed the memory file
is blocked for other subscribers and for writing publishers too. Maybe this can be eliminated
by a better memory file read/write access implementation (lock free read) in future releases.

Today, for specific scenarios (1:1 pub/sub connections with large payloads for example) this feature
can increase the performance remarkable. But please keep in mind to return from the message callback function
as fast as possible to not delay subsequent read/write access operations.


rex-schilasky:
ecal_zero_copy

First performance tests look promising. More then double performance for large payloads. Performance SND/REC sample shows 22 GB/s throughput on my Windows 10 machine !

ZhenshengLee:
Why this pr not merged into repo? @rex-schilasky

FlorianReimold:
The zerocopy implementation is far from finished and will break API, ABI, configuration and shared memory compatibility. Therefore it will continue to live in the branch until we are satisfied with the results and have decided on an upgrade path. We didn't want to have this open PR for weeks or months.
ZeroCopy *can* increase performance significantly, but it also hands additional responsibility to the user, which can also decrease the performance significantly if not used correctly. We want to implement additional measures to ease the negative effects that a zero copy mechanism can have. Always keep in mind that zerocopy is cool in academical scenarios where you only want to transport data. But usually you also want to do something with that data, not only transport it. And that's the point where you can actually block your entire system by blocking zerocopy memory. We want to implement additional measures to ease the negative effects that a zero copy mechanism can have.

PR #334, opened 2021-07-06
fix for #333

### LINKED ISSUE #333: Publisher send frequency breaks down to 0 when sending an empty protobuf message
**Describe the bug**
I have a protobuf publisher that sends out an "empty" protobuf message. The corresponding protobuf class inside of my program is only created and never filled up with anything. In the eCAL monitor I can see the publisher with the correct sending frequency. As soon as I subscribe the topic the frequency breaks down to 0 and it looks like the publisher has stopped sending. 
**Not sure if this only happens when using a protobuf syntax 2 proto file**

**To Reproduce**
1. Modify the eCAL sample "person_snd" in a way that the protobuf message only get's instantiated and send, but never filled up with anything 
2. Build this sample
3. Run this sample
4. Start the eCAL Monitor
5. See that it is sending
6. Subscribe it with the eCAL Monitor and see that it is not sending any more

**Expected behavior**
I would expect that the subscriber is still having a sending frequency and also sending out the empty protobuf message

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Version 18.04

rex-schilasky:


floriangeis (approved):

floriangeis:
Looks good. Behavior is now as expected: Publisher is still sending, when empty topic is subscribed.

Issue #115, opened 2020-10-31
Sensor message types for eCAL?


1939938853:
Hello,

eCAL  looks really simple to use comparing with ROS/ROS2. But maybe it is better to add more message types, especially the sensor message types?

For me I need to be able to send/get image type, point cloud type, and triangle-mesh data to/from  topic and services using OpenCV and Open3D.

In [ROS there are quite some sensor message types ](http://wiki.ros.org/sensor_msgs)to choose.  

Any suggestions?

Thank you!

rex-schilasky:
Hi, 

thank you for your nice feedback. Simplicity is one of the major API design decision of eCAL. Regarding the messages type support eCAL is quite different to ROS/ROS2. It's not providing any kind of message collection at all. It's middleware only, not focusing on a specific use case like robotics or AD. 

To create a google protobuf message file based on a ROS message file should be an easy task. And using google protobuf is not the only option, there are other message protocols (CapnProto, Flatbuffers) that you can utilize for your use case too. But protobuf is a good choice for many use cases, it's like a swiss army knife ;-).

If there is a general interest to have a standardized collection of messages for the robotic use case, maybe it would be a good idea to collect them in a separate repository.

1939938853:
Hi rex-schilasky, 

Thank you for your replay!

But  is the eCAL protobuf message transport possible using shared-memory (zero copy) among processes that are on the same physical machine?

rex-schilasky:
Hi,

yes this is completely independent. You can choose different kind of publisher / subscriber (raw data blobs, protobuf, capnproto, flattbuffers, strings) and how the data is transferred from a publisher to a matching subscriber is by default shared memory on the same machine and udp multicast if they are running on different hosts,

So if you publish in protobuf format then
  * the message will be serialized into a binary representation by the protobuf publisher
  * a binary publisher will check if there is a local or network connection to a matching subscriber
  * the data transfer starts by shared memory and / or udp multicast to one or more matching subscriber
  * a matching binary subscriber will receive the binary payload
  * the payload is serialized back into a  protobuf message object
  * the message object is forwarded to the protobuf subscriber

As you can see .. on the lower communication levels data will always transferred as binary representation. The kind of transport is independent from the chosen message protocol.

ReX

1939938853:
Hi Rex,

Thank you for your detailed explanation.  I think it is really a good choice of using eCAL in the case that only pub/sub and services are needed.  

[...]

PR #355, opened 2021-07-28
Add ros2 msg reflection plugin to ecal_mon_gui.

ZhenshengLee:
---
name: Pull request
about: Create a pull request for this project
title: Add ros2 msg reflection plugin to ecal_mon_gui
labels: ''
assignees: ''

---

**Pull request type**

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number:  https://github.com/continental/rmw_ecal/issues/18#issue-929783852

- The ecal app tool does not have a good support to ros2 system.
- rmw_ecal_dynamic_cpp does not support ros2cli like `ros2 topic echo`
- 

**What is the new behavior?**

- You can view some ros2 msgs through ecal_mon_gui
- It can be a tool that replaces `ros2 topic echo`

image

- [x] Yes
- [ ] No

**Other information**

As for now, it is still a demo case and may not be merged into the official repo.

Feel free to give me feedback. 

Thanks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,4 @@ doc/extensions/_autogen
 .idea
 cmake-build-*
 
+build/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,6 +33,7 @@ set(ECAL_PROJECT_ROOT "${CMAKE_CURRENT_LIST_DIR}")
 # command line build options
 # use it that way cmake .. -DBUILD_APPS=ON -DBUILD_SAMPLES=ON
 # --------------------------------------------------------
+option(HAS_ROS2                                "Platform supports ROS2 library"                                    OFF)
 option(HAS_HDF5                                "Platform supports HDF5 library"                                    ON)
 option(HAS_QT5                                 "Platform supports Qt 5 library"                                    ON)
 option(HAS_CURL                                "Build with CURL (i.e. upload support in the recorder app)"         ON)
@@ -127,7 +128,7 @@ set(possible_subprojects
   CMakeFunctions
 )
 
-# We should rename the option, but don't know how to do in in a 
+# We should rename the option, but don't know how to do in in a
 # backwards compatible way
 set(ECAL_THIRDPARTY_BUILD_CMAKEFUNCTIONS ${ECAL_THIRDPARTY_BUILD_CMAKE_FUNCTIONS})
 
--- a/app/mon/mon_plugins/CMakeLists.txt
+++ b/app/mon/mon_plugins/CMakeLists.txt
@@ -5,9 +5,9 @@
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
-# 
+#
 #      http://www.apache.org/licenses/LICENSE-2.0
-# 
+#
 # Unless required by applicable law or agreed to in writing, software
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -24,3 +24,6 @@ add_subdirectory(capnproto_reflection)
 endif (HAS_CAPNPROTO)
 add_subdirectory(raw_data_reflection)
 add_subdirectory(string_reflection)
+if(HAS_ROS2)
+add_subdirectory(ros2_reflection)
+endif()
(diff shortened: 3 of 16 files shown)


rex-schilasky:
First of all thank you for contributing. As you said it's more a prototype and needs some more work. I checked your implementation and it is specifically supporting 2 message types for now. Is there any kind of a more generic approach to dynamically reflect the content of ROS message types ? I don't think so, but wanted to be sure I did not miss something.

ZhenshengLee:
> First of all thank you for contributing. As you said it's more a prototype and needs some more work. I checked your implementation and it is specifically supporting 2 message types for now. Is there any kind of a more generic approach to dynamically reflect the content of ROS message types ? I don't think so, but wanted to be sure I did not miss something.

@rex-schilasky 

Perhaps rmw_ecal can help to get msg type info from the given parameters including type name and node name.
As for now, many `if` statements will be an ugly solution for echo ros2 msgs.
I'll push more commits, and before that this pr will be a draft.

Thanks.

brakmic-aleksandar:
Hmm, maybe rmw_ecal could provide API for deserialization of messages, but the plugin itself will have to provide filled rosidl_typesupport_introspection_cpp::MessageMembers or rosidl_typesupport_introspection_c__MessageMembers with metadata of messages.

rex-schilasky:
@ZhenshengLee what is the reason to not use rosidl_typesupport_protobuf + rmw_ecal and then to use the eCALMon builtin reflection for the protobuf message format. With that solution would have a generic reflection for all ROS messages.

ZhenshengLee:
> @ZhenshengLee what is the reason to not use rosidl_typesupport_protobuf + rmw_ecal and then to use the eCALMon builtin reflection for the protobuf message format. With that solution you would have a generic reflection for all ROS messages.

You are right, but currently there is an issue of runtime crash. https://github.com/continental/rmw_ecal/issues/19

According to https://github.com/continental/ecal/issues/115 ecal is designed as a middleware and is not domain specific.

So there are things to do to get "rosidl_typesupport_protobuf + rmw_ecal" used with comfort in ROS2 , like rviz support.

RViz will stuck now with rmw_ecal running currently.

BTW, ptotobuf serialization may makes it easier to connect apollo cyber and ros2.

Thanks.

rex-schilasky:
Okay, so we need to fix that issue in rmw_ecal first, Aleks is working on it. Regarding protobuf I fully agree. If you record ROS2 messages with eCALRec into HDF5 container, serialized as protobuf you have lots of new options:

* eCALRec can record distributed on multiple hosts using shm ipc only
* the performance is often limited by the mass storage throughput only (currently we run a project that is recording 4 cameras 8M with 30 fps with zero loss of frames, its about 900 MB/s on that host, other host doing their own measurements in parallel)
* the reuse / compatibility of that kind of HDF5 / protobuf measurements is guaranteed for a long time, even message schemas are changing slightly, this is reducing cost for transforming or rerecording dramatically

ZhenshengLee:
> @ZhenshengLee: I found an interesting discussion at ROS discourse regarding ROS2 type support introspection [here](https://discourse.ros.org/t/ros-2-type-support-introspection-and-conversion-of-c-c-messages-to-from-yaml/22295). Seems to me a generic approach to convert a ROS2 message to a YAML representation. This YAML could be shown a a simple string in an eCALMon plugin, couldn't it ?

@rex-schilasky 

Thanks for your report, I checked the [doc](https://github.com/christophebedard/dynamic_message_introspection/blob/add-cpp-support-and-refactor/dynmsg_demo/README.md) and It
> do dynamic introspection of messages using the rcl message API

As the usability of rcl api depends on rmw api, and then the implementation of rmw, due to lack of enough rmw api support of rmw_ecal, for example `rcl_get_publishers_info_by_topic` or `rmw_get_publishers_info_by_topic`, this demo will get compiled or work.

I already created an issue about lack of rmw api support, check https://github.com/continental/rmw_ecal/issues/18 , which has caused most of ros2 tools unavailable.

Currently this PR only uses msg definition hpp and helper functions and does not depend on rmw.

rex-schilasky:
We should discuss to close the PR .. to have a ROS message reflection seems to be a very specific use case to me, most of the users seems to use eCAL with protobuf or ROS and it's message system.

@ZhenshengLee can you please work on that feature isolated ? If there is a ROS concept for doing generic reflection like google protobuf is able to do we can easily create a monitor plugin out of it to support such a feature.

PR #358, opened 2021-08-09
Feature/shm performance

rex-schilasky:
**Pull request type**

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

eCAL does not support zero copy message transport. This can lead to lower data throughput for larger messages (> 512 kB).

**What is the new behavior?**

This PR implements an optional zero copy feature that can be activated globally (host wide) by setting the new parameter ```memfile_zero_copy``` in the eCAL configuration file or by using the new publisher API function ```EnableZeroCopy```. The default parameter is ```OFF```.

Because the subscriber callback function is accessing the memory file content without any extra copy (zero copy :-)), the memory is blocked for other any other read/write access. To not block publisher for writing new content a multiple buffering feature is added. By default a publisher is using ONE memory file buffer for a 1:n connection. By using the new publisher API function ```SetBufferCount``` or the configuration parameter ```memfile_buffer_count``` the number of parallel utilized memory files can be set. These files act like a kind of FIFO ring buffer. The default setting is one buffer.

- [ ] Yes
- [x] No

There are 2 new publisher API functions. The new shared memory layer is still compatible with previous implementations, with one exception. Payloads from previous version can be always received. For the local ipc communication from this version to previous versions, the new API function `SetBufferCount` will break compatibility if more then one buffer is used. Default is still one memory file buffer only. New API function `EnableZeroCopy` can be used and will be ignored by previous versions (they still make a memcpy of the received payload). The inter host (UDP) protocol is not affected by that change at all.

**Other information**

Performance comparison between eCAL Default, eCAL ZeroCopy, eCAL IceOryx.

ecal_zero_copy


ecal/core/src/readwrite/ecal_writer_shm.cpp
```diff
@@ -186,7 +186,7 @@ namespace eCAL
     return true;
   }
 
-  std::string CDataWriterSHM::GetConectionPar()
+  std::string CDataWriterSHM::GetConnectionPar()
```
Kerstin-Keller, line 189: Why don't you spell out the function name? `GetConnectionParameter`

ecal/core/src/readwrite/ecal_writer_shm.h
```diff
@@ -50,7 +50,7 @@ namespace eCAL
     bool AddLocConnection(const std::string& process_id_, const std::string& conn_par_) override;
     bool RemLocConnection(const std::string& process_id_) override;
 
-    std::string GetConectionPar() override;
+    std::string GetConnectionPar() override;
```
Kerstin-Keller, line 54: Same here 😉

Kerstin-Keller (commented):
Let's not use these half-abbreviations if things are getting changed anyways.

Issue #371, opened 2021-08-25
CServiceClient CallAsync() may silently fail


bsirang:
**Describe the bug**
There are cases where `CallAsync()` may not actually initiate a request due to some "early-exit" conditions. In these cases the user's callback is never invoked.

Relevant code snippet from `CServiceClientImpl::CallAsync`:
```
    if (!g_servgate()) return;
    if (!m_created)    return;

    if (m_service_name.empty()
      || method_name_.empty()
      )
      return;
```

**To Reproduce**
1. Invoke `CallAsync` on a service which does not exist.
2. Observe the callback is never called

**Expected behavior**
If the async call fails immediately, it needs to be signaled to the user somehow. Either by calling the user's callback with a flag indicating failure, or by return value since the error is immediate.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Version 20.04

PR #375, opened 2021-08-30
Added callback calls on CallAsync service errors

brakmic-aleksandar:
**Pull request type**

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

When client calls service which doesn't exist/isn't initiated yet it fails silently.

Issue Number: #371

**What is the new behavior?**
Client will let user know that it couldn't connect to service through callback.

- [ ] Yes
- [x] No

### DIFF
--- a/ecal/core/src/service/ecal_service_client_impl.cpp
+++ b/ecal/core/src/service/ecal_service_client_impl.cpp
@@ -147,13 +147,23 @@ namespace eCAL
 
   void CServiceClientImpl::CallAsync(const std::string& host_name_, const std::string& method_name_, const std::string& request_)
   {
-    if (!g_servgate()) return;
-    if (!m_created)    return;
+    if (!g_servgate())
+    {
+      ErrorCallback(method_name_, "Servgate error.");
+      return;
+    }
+    if (!m_created)
+    {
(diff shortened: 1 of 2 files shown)


FlorianReimold:
This adds a new API function right? Wich would mean it breaks ABI compatibility :/

brakmic-aleksandar:
@FlorianReimold It adds private function to impl class which should prevent breakage of ABI compatibility, if i'm not mistaken.

rex-schilasky:
For the new ` CServiceClientImpl::CallAsync` methods I would prefer the return type `bool` too. Like there counterpart signatures `bool CServiceClientImpl::Call`.

ecal/core/src/service/ecal_service_client_impl.cpp
```diff
@@ -147,13 +147,23 @@ namespace eCAL
 
   void CServiceClientImpl::CallAsync(const std::string& host_name_, const std::string& method_name_, const std::string& request_)
```
rex-schilasky, line 148: `bool CServiceClientImpl::CallAsync`

rex-schilasky (changes requested):

Issue #396, opened 2021-09-09
Support for dynamic protobuf publisher


bsirang:
**Is your feature request related to a problem? Please describe.**
This request is for use cases where the publishing protobuf message type is not known at compile time, but rather at runtime. 

One such use case is creating a general purpose utility that publishes a particular message given as a JSON string to a particular topic with a given proto message

**Describe the solution you'd like**
A sibling class to `eCAL::protobuf::CPublisher<T>` which instead of using a template type `T` uses the abstract `google::protobuf::Message` instead.

**Describe alternatives you've considered**
I can't think of an alternative

**Additional context**
Here's how protobuf APIs can be used to create a `Message` object from a `std::string` that represents the type
```
  const google::protobuf::Descriptor* desc =
      google::protobuf::DescriptorPool::generated_pool()->FindMessageTypeByName(type_name);

  std::shared_ptr<google::protobuf::Message> message{
      google::protobuf::MessageFactory::generated_factory()->GetPrototype(desc)->New()};

  ```
In the example above `type_name` is a std::string representing the proto type such as `my.proto.package.MyMessage`

The created `Message` object has all the metadata needed for eCAL. I validated this myself with an implementation in my project, but it would be good to have it built in.

rex-schilasky:
@bsirang Can you share a minimal sample how you use that method today ? The basic `eCAL::CPublisher` class needs to setup 'topic_name', 'topic_type' and the 'topic_descriptor' [(API here)](https://continental.github.io/ecal/_api/classeCAL_1_1CPublisher.html#_CPPv4N4eCAL10CPublisher10CPublisherERKNSt6stringERKNSt6stringERKNSt6stringE).

Did you initiate such a class instance and could you see your message data in the ecal monitor protobuf reflection view ? What we would need is a kind of "person_snd_dynamic" sample. That would be great.

bsirang:
@rex-schilasky Here's the quick proof-of-concept I put together. The key is that it is a publisher that works on a `google::protobuf::Message`. The message type still gets fixed during construction of the class instance since we need `GetTypeName()` and `GetDescription()`, but the type doesn't need to be known at compile time.

When I used `ecal_mon_gui` I could see the correct message type and it parsed all the fields and metadata correctly.

```c++
class DynamicPublisher : public eCAL::CMsgPublisher<google::protobuf::Message> {
 public:
  DynamicPublisher(const std::string& topic_name, std::shared_ptr<google::protobuf::Message> msg)
[...]

PR #404, opened 2021-09-29
first draft for the new protobuf dynamic publisher (#396) (Thx bsiran…

rex-schilasky:
**Pull request type**

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Introduces dynamic protobuf publisher (#396)


samples/cpp/person/person_snd_dyn/src/person_snd_dyn.cpp
```diff
@@ -0,0 +1,71 @@
...
+    // set person object content
+    person->set_id(++cnt);
+    person->set_name("Max");
+    person->set_stype(pb::People::Person_SType_MALE);
+    person->set_email("max@mail.net");
+    person->mutable_dog()->set_name("Brandy");
+    person->mutable_house()->set_rooms(4);
+
+    // send the person object
+    pub.Send();
```
bsirang, line 52: I understand why you don't allow the user to pass in a message here. You want to prevent the user from attempting to send a message structure that's different than what the publisher was constructed with.

The downside I see is that the user has to maintain a reference to the `person` message (in this example) alongside the `pub` object. Looks like there's no way to query the `pub` object for the underlying message.

For my use case providing a getter method that simply returns a `std::shared_ptr<google::protobuf::Message>` would be helpful. This way if you use this constructor `CDynamicPublisher(const std::string& topic_name_, const std::string& proto_type_name_)` you have a way to access the proto message instance.

samples/cpp/person/person_snd_dyn/src/person_snd_dyn.cpp
```diff
@@ -0,0 +1,71 @@
...
+{
+  // initialize eCAL API
+  eCAL::Initialize(argc, argv, "person publisher dynamic");
+
+  // set process state
+  eCAL::Process::SetState(proc_sev_healthy, proc_sev_level1, "I feel good !");
+
+  // create a dynamic publisher (topic name "person")
+  std::shared_ptr<pb::People::Person> person{ new pb::People::Person() };
+  eCAL::protobuf::CDynamicPublisher pub("person", person);
```
bsirang, line 37: Curious did you try using the other constructor in your samples? I believe the string `pb.People.Person` should work.

bsirang (commented):
Thanks for including this.

samples/cpp/person/person_snd_dyn/src/person_snd_dyn.cpp
rex-schilasky: Just tried out the second constructor and yes, it's working that way. But in this case I can't access the message object :-). That is true.

rex-schilasky: If I would have a getter like
`
    std::shared_ptr<google::protobuf::Message> GetMessage()
    {
      return m_msg;
    }
`
how to convert that `std::shared_ptr<google::protobuf::Message>` into the specialized message type like `std::shared_ptr<pb::People::Person>` ?

bsirang: In my use case, I actually wouldn't care to reinterpret the message as another type.

I am using this: `google::protobuf::util::JsonStringToMessage(body, message.get(), json_options);`

Where `message` is `std::shared_ptr<google::protobuf::Message>`. So even that simple getter would satisfy my use case.

However, if you wanted to, you could use this to reinterpret the shared_ptr as a `std::shared_ptr<pb::People::Person>`: https://en.cppreference.com/w/cpp/memory/shared_ptr/pointer_cast

For convenience you could have a templated method `GetAs()` or something. For example:
`std::shared_ptr<pb::People::Person> person = pub.GetAs<pb::People::Person>();`

Then `GetAs()` would just do the pointer reinterpret casting.

rex-schilasky: Sounds good to me, please check my last commit.

ecal/core/include/ecal/msg/protobuf/dynamic_publisher.h
```diff
@@ -0,0 +1,191 @@
...
+
+      /**
+       * @brief  Get underlying protobuf message object as T.
+       *
+       * @return  Message object.
+      **/
+      template<typename T>
+      std::shared_ptr<T> GetAs()
+      {
+        return std::static_pointer_cast<T>(m_msg);
```
bsirang, line 104: I think you want `dynamic_cast` here since it's a base to derived class conversion

bsirang (commented):
Looks reasonable!

Issue #420, opened 2021-11-22
Event callback for client/server service?


jtschea:
Dear developers,

thanks a lot for this excellent product :smiley: 

For classes [CSubscriber](https://continental.github.io/ecal/_api/classeCAL_1_1CSubscriber.html#) and [CPublisher](https://continental.github.io/ecal/_api/classeCAL_1_1CPublisher.html) there's a method `AddEventCallback` to get notified when connection state has changed.

Is there something similar for both [CServiceClient](https://continental.github.io/ecal/_api/classeCAL_1_1CServiceClient.html) and [CServiceServer](https://continental.github.io/ecal/_api/classeCAL_1_1CServiceServer.html)?

There's the chance that a RPC call got lost if the server is not yet up and running and we wouldn't want to rely on `eCallState` for checking connection state.

Do you maybe have a suggestion for a workaround?

Thanks for your help and best regards,
Jupp

rex-schilasky:
Hi Jupp,

thank you for the nice feedback. Currently we do not have an `AddEventCallback` method similar to `CSubscriber` and `CPublisher` for the services. There was no user request until today :).

But in general internally service server and client both do have methods that are called at least when a new matching service connects. So for at least a the events `service_server_connected` and `service_client_connected` could be implemented easily.

Let me check that and implement a first draft in a branch for reviewing.

Best, ReX

rex-schilasky:
As a workaround there is another API method in `eCAL::Process`. With `eCAL::Process::AddRegistrationCallback` you can register a method that informs you if a new publisher, subscriber, service or process is registered.

```cpp
#include <ecal/ecal.h>
#include "ecal/pb/ecal.pb.h"

#include <iostream>

void OnServiceRegistration(const char* sample_, int sample_size_)
{
  eCAL::pb::Sample sample;
  if (sample.ParseFromArray(sample_, sample_size_))
  {
    std::cout << "service host : " << sample.service().hname() << std::endl;  // host name
    std::cout << "service uname: " << sample.service().uname() << std::endl;  // process name
    std::cout << "service name: "  << sample.service().sname() << std::endl << std::endl;  // service name
  }
}

int main(int argc, char **argv)
{
  // initialize eCAL API
  eCAL::Initialize(argc, argv, "monitoring registrations", eCAL::Init::None);

  // add service register callback function
  eCAL::Process::AddRegistrationCallback(reg_event_service, std::bind(OnServiceRegistration, std::placeholders::_1, std::placeholders::_2));

[...]

Issue #372, opened 2021-08-26
Service client `Call` and `CallAsync` APIs do not have a timeout argument


bsirang:
**Describe the bug**
The APIs in `ecal_client.h` for `CServiceClient::Call` and `CServiceClient::CallAsync` do not have a timeout option. It doesn't appear the user has any way of bounding the time that the operation takes. This makes it more challenging for application code to be robust against missing or unresponsive services.

**To Reproduce**
1. Initiate a `Call` to a very long running service handler

**Expected behavior**
The client call fails after a user-specified timeout and signals the timeout to the application.

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Version 20.04

**Additional context**

rex-schilasky:
@bsirang Hi, let us check what we can do here. We are working on a few other issues we will hopefully close soon. I fully agree that a timeout would be a nice option here. The services are handled by underlaying ASIO sockets and I remember that there was no option for timeout offered by the API. But let's check again.

bsirang:
Hi @rex-schilasky I actually have a fair bit of ASIO experience and I have implemented timeouts as well, so I'm happy to provide guidance.

The pattern I've used before was to kick off a timer with the IO request.

If the timer expires first, then it aborts the pending IO operation, and then the IO callback will get an "operation aborted" error, which you can treat as a timeout condition.

If the IO operation completes first, you can cancel the timer. You would have to couple a timer operation with each IO operation.

bsirang:
When I say "It aborts the pending IO operation", I mean invoking `cancel()` method such as from: https://think-async.com/Asio/asio-1.18.2/doc/asio/reference/basic_stream_socket/cancel.html

bsirang:
Oh, and I should mention that the timer and the socket operation should happen on the same event loop (same io_context) so that the timer expiration and the IO operation can't truly simultaneously complete. One always comes first. This avoids potential race conditions.

rex-schilasky:
Hi @bsirang, I will start to work on that issue now. It will be part of an eCAL service / client API redesign. See PR #427. I will change the signature of the blocking service call API first and add a timeout event to the new EvenCallback logic. Let us discuss here how to implement the timeout logic the right way.

bsirang:
Hi @rex-schilasky I appreciate the update! I'm happy to look at any code you point me toward and/or discuss implementation details. Cheers.

rex-schilasky:
[...]

PR #427, opened 2021-11-30
Add event callback API (#420) and Call and CallAsync API (#372)

rex-schilasky:
new service event callback API added (connect, disconnect)
redesign of service gate logic (separate gate added for client entities)
clients registration added (clients can be visualized in eCALMon later in extra fix)
client blocking service call improved to allow broadcast call of all services and return all responses as service info vector
client and service implementation cleaned up

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

No callback logic for service events.

Issue Number: #420

**What is the new behavior?**
New service and client events for connecting and disconnecting as callback logic available (how to use - see samples math_service and math_client).

- [X] Yes
- [ ] No

New API methods for eCAL::CServiceServer and eCAL::CServiceClient added. Protobuf registration sample extended with client type (downwards compatible).


ecal/core/include/ecal/cimpl/ecal_publisher_cimpl.h
```diff
@@ -214,7 +214,7 @@ extern "C"
    *
    * @return  None zero if succeeded.
   **/
-  ECALC_API int eCAL_Pub_AddEventCallbackC(ECAL_HANDLE handle_, enum eCAL_Publisher_Event type_, PubEventCallbackCT callback_, void* par_);
+  ECALC_API int eCAL_Pub_AddEventCallback(ECAL_HANDLE handle_, enum eCAL_Publisher_Event type_, PubEventCallbackCT callback_, void* par_);
```
Kerstin-Keller, line 218: @rex-schilasky aren't these (C-) API breaking changes?
Why have they been renamed?

rex-schilasky: Yes, these changes are breaking the (C-) API. For the new services there are some new C and C++ functions including an improved API logic so that we may take into consideration to cleanup (break) the whole services API.

This specific renaming is because of inconsistent naming for add and remove event callback methods for publisher / subscriber and services. Some end with "C" some not. In general all other C API functions do not use a trailing "C" to declare them additional as C API.

Issue #447, opened 2022-01-04
eCAL Monitor protobuf reflection should show default message values.


Kerstin-Keller:
**Is your feature request related to a problem? Please describe.**
When opening eCAL Monitor, default values of protobuf messages are not shown at all.
This is frustrating for the user.
Also it leads to the problem, that field disappear and re-appear, if a value toggles between a non-default and the default value.

**Describe the solution you'd like**
Show all fields, even default ones.

**We should:**
- evaluate if behavior changes in newer protobuf versions
- find a way to use reflection differently?

Kerstin-Keller:
Maybe https://github.com/protocolbuffers/protobuf/issues/9365 will tell us more :)

PR #452, opened 2022-01-05
Feature/default values

Kerstin-Keller:
**Pull request type**

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Solves #447

- [ ] Yes
- [x] No


contrib/ecalproto/src/ecal_proto_decoder.cpp
```diff
@@ -42,22 +42,25 @@ namespace protobuf
...
     if (!is_Array_)
     {
-      visitor->ScalarMessageStart({ name_, prefix_, (int)index_, msg_.GetTypeName()}, GetProtoMessageFieldNames(field_v));
+      visitor->ScalarMessageStart({ name_, prefix_, (int)index_, msg_.GetTypeName()}, GetProtoMessageFieldNames(descriptor));
     }
     else
     {
-      visitor->ArrayMessageStart({ name_, prefix_, (int)index_, msg_.GetTypeName()}, GetProtoMessageFieldNames(field_v));
+      visitor->ArrayMessageStart({ name_, prefix_, (int)index_, msg_.GetTypeName()}, GetProtoMessageFieldNames(descriptor));
     }
```
Kerstin-Keller, line 56: @FlorianReimold  this is the reasons why even submessages are displayed (though empty).
I guess we can just leave it like this?

cmake/helper_functions/ecal_add_functions.cmake
```diff
@@ -79,6 +79,10 @@ function(ecal_add_mon_plugin TARGET_NAME)
     SOVERSION $<NOT:$<CXX_COMPILER_ID:Clang,AppleClang>:${${TARGET_NAME}_VERSION_MAJOR}>>
     LIBRARY_OUTPUT_DIRECTORY $<IF:$<BOOL:${WIN32}>,${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/$<CONFIG>/ecalmon_plugins,${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/ecal/plugins/mon>
   )
+  target_compile_definitions(${TARGET_NAME}
+    PRIVATE
+      $<$<CONFIG:Release,RelWithDebInfo,MinSizeRel>:QT_NO_DEBUG>
+  )
```
Kerstin-Keller, line 86: I wasn't able to load the mon plugins in Release / RelWithDebug mode unless `QT_NO_DEBUG` is set. I guess it shouldn't hurt to set it explicitly for non Debug builds, also see https://stackoverflow.com/questions/16780797/incompatible-qt-library-between-app-and-plugin-compiled-same-way

FlorianReimold (approved):

PR #481, opened 2022-01-24
Feature/unicode support

### LINKED ISSUE #417: eCAL Player cannot load measurements from paths with non-ASCII characters
**Describe the bug**
The eCAL Player fails to load measurements from paths with non-ASCII characters, at least on Windows.

**Screenshots**
grafik

**Additional context**
I should probably also check whether the **eCAL Recorder** can handle non-ascii characters

FlorianReimold:
Fixes #417 

Changed hopefully all Windows calls to ASCII-fuctions (like `CreateDirectoryA` or `open()`) by WideString variants. eCAL will from now on assume everything to be UTF-8 encoded, which should make perfect sense in 2022.


rex-schilasky (approved):
So far okay from my side. It's a large PR, so just a rough review file by file ..

contrib/ecalhdf5/CMakeLists.txt
```diff
@@ -43,7 +43,6 @@ set(ecalhdf5_src
     src/eh5_meas_file_v5.cpp
     src/eh5_meas_file_v5.h
     src/eh5_meas_impl.h
-    src/eh5_util.h
     src/escape.cpp
```
Kerstin-Keller, line 46: does this line have anything to do with the rest?

Kerstin-Keller (commented):
Except my comment I don't have any remarks.
Except that I hate all those `#ifdef WIN32`, though I don't think that there is a solution for that.

contrib/ecalhdf5/CMakeLists.txt
FlorianReimold: The eh5-util.h was a utily header that wasn't used any more. So I didn't bother with unicode-porting and just removed it. It all uses the EcalUtils now, which supports unicode paths.

Issue #436, opened 2021-12-03
Add generic description field to CSubscriber/CPublisher


brakmic-aleksandar:
**Is your feature request related to a problem? Please describe.**
Currently CSubscriber/CPublisher classes have topic_desc_ argument in their constructors, but it is used mainly for their type description, it would be nice if we had generic description field as well which could be used to have some metadata related to specific subs/pubs or just to add plain text descriptions for them.

**Describe the solution you'd like**
Rename topic_desc_ cosntructor parameters to type_desc_ and add topic_desc_ optional string argument to them, this way it shouldn't break API.

brakmic-aleksandar:
Additional info:
This feature is needed by rmw_ecal to implement [get_topic_endpoint_info functions](https://docs.ros2.org/foxy/api/rmw/get__topic__endpoint__info_8h.html) which provide node name/namespace for specific pub/sub (which is required by ros tools such as rosbag), the best way to do this would be to specify additional field in monitoring layer which should contain info needed for those functions, to make this feature not specific to rmw_ecal i propose we name new monitoring field either generic_desc or endpoint_desc and make it a string field, so that it can be applied to different use cases.

FlorianReimold:
So we would basically extend the topic info
https://github.com/continental/ecal/blob/master/ecal/pb/src/ecal/pb/topic.proto#L45

which is used in the big monitoring information struct:
https://github.com/continental/ecal/blob/master/ecal/pb/src/ecal/pb/monitoring.proto#L45

and expose setting this user-defined field to anything.

PR #483, opened 2022-01-25
Add generic description field to monitoring data

brakmic-aleksandar:
**Pull request type**

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

eCAL monitoring data only contains description for topic type.

Issue Number: #436 

**What is the new behavior?**
eCAL monitoring data will contain generic description for publisher and subscriber endpoints.

- [ ] Yes
- [x] No

**Other information**


Kerstin-Keller:
I have a few question:
- This is something per topic, e.g. "Calibration information from Sensor XYZ"?
- Does the user set this information (Maybe `SetUserDescription`, `SetTopicComment` or similar will be a more appropriate naming?
- Only publishers set this, or also subscribers?
- Do they need to be identical on pub / sub side? Will eCAL issue warnings (like with Descriptor)?

brakmic-aleksandar:
> I have a few question:
> 
>     * This is something per topic, e.g. "Calibration information from Sensor XYZ"?
> 
>     * Does the user set this information (Maybe `SetUserDescription`, `SetTopicComment` or similar will be a more appropriate naming?
> 
>     * Only publishers set this, or also subscribers?
> 
>     * Do they need to be identical on pub / sub side? Will eCAL issue warnings (like with Descriptor)?

Yeah, this is something for each topic endpoint, so every subscriber and publisher can have its own unique data in it, user should set that info, and they can use it for whatever they want.

I plan on using this to store parent node information (name and namespace) for pubs/subs in rmw_ecal. 

Not sure about naming, GenericDescription sounded clear enough to me (but i'm very bad at naming things, so that it with grain of salt).

ecal/core/include/ecal/cimpl/ecal_publisher_cimpl.h
```diff
@@ -74,6 +74,17 @@ extern "C"
   **/
   ECALC_API int eCAL_Pub_SetDescription(ECAL_HANDLE handle_, const char* topic_desc_, int topic_desc_len_);
 
+  /**
+   * @brief Setup generic publisher description. 
+   *
+   * @param handle_          Publisher handle. 
+   * @param topic_desc_      Generic description. 
+   * @param topic_desc_len_  Generic description length. 
+   *
```
rex-schilasky, line 83: `topic_desc_` and `topic_desc_len_` should be renamed

ecal/core/include/ecal/cimpl/ecal_subscriber_cimpl.h
```diff
@@ -93,7 +93,18 @@ extern "C"
...
   ECALC_API int eCAL_Sub_SetID(ECAL_HANDLE handle_, const long long* id_array_, const int id_num_);
-  
+
+  /**
+   * @brief Setup generic subscriber description. 
+   *
+   * @param handle_          Publisher handle. 
+   * @param topic_desc_      Generic description. 
+   * @param topic_desc_len_  Generic description length. 
+   *
```
rex-schilasky, line 103: `topic_desc_` and `topic_desc_len_` should be renamed

rex-schilasky:
I'm still not a 100 percent happy with this new API. This new description is a new additional information provided by a publisher and received by all connected subscriber right ?

So it is not like a topic description or topic name. These information are shared between all matching publisher and subscriber and a kind of independent entities in the eCAL network.

So could we maybe rename it to something like

```cpp
    bool CPublisher::SetAttribute(const std::string& attr_name_, const std::string& attr_, );

    bool CSubscriber::GetAttribute(const std::string& attr_name_, std::string& attr_, );
```

and use a protobuf map

```proto
  map<string, string> attr;
```

to encode this ?

brakmic-aleksandar:
Well, no, this info isn't received by matching subs, all of subs and pubs can have unique info (description) attached to them, it would be used to describe (in whatever format you want) one specific publisher or subscriber, not the topic itself, it could be used to either describe the purpose of that specific endpoint in plain text, or to add some kind of encoded info for it that could be later retrieved through monitoring data and parsed.

I like the idea of naming it attribute and using map for it.

rex-schilasky:
Got it ... but to name it as Set/GetAttribute and using a key, value signature would make it real flexible for future extensions.

brakmic-aleksandar:
> Got it ... but to name it as Set/GetAttribute and using a key, value signature would make it real flexible for future extensions.

I like that idea! :)

brakmic-aleksandar:
Although i think 
`    bool CSubscriber::GetAttribute(const std::string& attr_name_, std::string& attr_);`
Should be changed to 
`    std::string CSubscriber::GetAttribute(const std::string& attr_name_);`
Because rest of the getters in CPublisher and CSubscriber classes follow this pattern.

FlorianReimold:
> ` std::string CSubscriber::GetAttribute(const std::string& attr_name_);` 

Hm, you will loose the information about wether the key hasn't been set or if it has been set to an empty string. But I guess if there are other functions as well that hide such an information i would still vote for that, for keeping it consistent.

I am overall very happy with the idea of a generic attribute map. This looks very clean to me and already leaves room for expansions in the future. 👌

rex-schilasky:
> Although i think ` bool CSubscriber::GetAttribute(const std::string& attr_name_, std::string& attr_);` Should be changed to ` std::string CSubscriber::GetAttribute(const std::string& attr_name_);` Because rest of the getters in CPublisher and CSubscriber classes follow this pattern.

I would keep the boolean return typed version for the reasons Florian mentioned. Even other existing API functions do not follow this design today, I had some headache to redesign the service API to get exactly that level of control. Is the key set or not ? Was it set to an empty string ?? Maybe a default protocol buffer with length 0 ?

FlorianReimold:
Yes, an optional would be a proper modern way for a return value. But at the moment Visual Studio 2015 ist still way too much used to give up C++14 compatiblity.

The Idea here is to set attributes to both publishers and subscribers. I think you imagine that an attribute gets set on the publisher and read on the subscriber side, but that is not the case. You set on either publishers and subscribers and read again through the monitoring API directly from the protobuf object. I guess the get option is only there to see what you have set previously.

rex-schilasky:
Looks good to me. The ```GetAttribute``` still returns a string value instead a boolean one but we can keep that as it is and if needed we can add a ```HasAttribute``` function to check for existing attributes.

The C API is missing the ```GetAttribute``` variants. Do we need that ? Don't think so for now. Wrapped API's (like python) that are based on the C API can keep that information in their own wrapper layer.

So if no one votes for an additional ```HasAttribute``` I would approve that PR.

Thank you Aleks !! Great job.

rex-schilasky (approved):

KerstinKeller:
Sorry for being so picky. But what we decide here will become part of the public API, and then we're stuck with the decision, at least until eCAL6 😄 

Is there any real reason why `SetAttribute` may fail (return false). Can anything go wrong? Does it have to do with the `IsCreated()`?

I guess I am just not a big fan of this whole `Create()` logic with the publishes, because it's very much as non RAII as it can get 😉 

Otherwise, I think it's ok.

FlorianReimold:
As I see it in the current state, we have the following limitations:

- An attribute can be set to any string, including a string containing null-bytes (this is good and not a limitation 😉)
- One can set an attribute to an empty string, but this will cause eCAL to transport that empty string (but this is what you would expect)
- An attribute can never be unset again
- The API can never ever determine again whether an attribute has been set to an empty string or hasn't been set at all
- There is no way to iterate over attributes, you would have to brute-force names

As Kerstin mentioned this PR is public API and therefore worth to discuss. We should at try to make this as future-proof as we can get 👍

brakmic-aleksandar:
I agree with first 2 points, we should let users set empty strings and ones containing zero_bytes. Other 3 functions should be easy to implement. Regarding iteration we could just return const reference to underlying map or maybe just a iterator?

rex-schilasky:
If I now understand the idea right, then those attributes will finally read out over the monitoring API. The logic is NOT to make a `SetAttribute` on the publisher side and a `GetAttribute` on the subscribers side. But from the users viewpoint it seems that the logic is more or less the same like `SetDescription` and `GetDescription`, but it's not.

So maybe we do not provide the `GetAttribute` interface at all ? Instead we could have a

```cpp
bool SetAttribute(const std::string& key, const std::string& value);
// and
bool ClearAttribues();

// or

bool SetAttributes(std::map<std::string, std::string> attr); 
```

So if you lost control what is set or not, just call `SettAtributes` (who you gonna call ??? ;-)).

FlorianReimold:
>Regarding iteration we could just return const reference to underlying map or maybe just a iterator?

> bool SetAttributes(std::map<std::string, std::string> attr); 

Hm, both functions are quite hard to wrap to C, i guess. We could also provide a function to return the keys

```cpp

void SetAttribute(const std::string& key, const str::string& value);

bool UnsetAttribute(const std::string& key)

std::vector<std::string> GetAttributeKeys(); // Still a pain to wrap to C, I know...

std::string GetAttribute(const std::string& key);

```

I actually don't like `bool SetAttributes(std::map<std::string, std::string> attr);`, as it would automatically unset all attributes that were set before. This contradicts the idea of having multiple key-value-pairs. If it is 1 monolithic value after all, we could just as well have a bytes or string field.

Kerstin-Keller:
So actually, the longer I think about it, why not go back to square one, how this all started
 
``` 
    bool SetAttribute(const std::string& attr_name_, const std::string& attr_value_);
    //bool GetAttribute(const std::string& attr_name_, std::string& output) const;
```
I'd even argue, let's omit the `GetAttribute`. Only `SetAttribute`. What's the usecase anyways?

What I guess we should think about, though, does this info need to be preserved in measurements? Like the descriptor is?

Then for 6.0 we should strive for a different kind of API for Publishers / Subscribers alike.
```
  std::shared_ptr<PublisherAttributes> GetAttributes();
  std::shared_ptr<PublisherSettings> GetSettings(); // can modify QOS Settings, ...
```

rex-schilasky:
So maybe a good first compromise would be just this minimalistic interface ?

```cpp
void SetAttribute(const std::string& key, const str::string& value);
bool UnsetAttribute(const std::string& key)
```

brakmic-aleksandar:
I've updated API according to our previous meeting, i've also marked new api methods as experimental.

rex-schilasky (approved):
Perfect

PR #486, opened 2022-01-26
[Build]: Split CONFIG expressino to support legacy cmake <3.19

FlorianReimold:
Pre-3.19 cmake did not accept multiple values in a $<CONFIG:xxxx> expression.


Kerstin-Keller (approved):
😢 I really liked the new syntax!

Issue #493, opened 2022-01-27
Protobuf service name could not be set independently


rex-schilasky:
**Describe the bug**
Protobuf server / client service instance names can not be set by the user.

**To Reproduce**
Currently the server side is implemented like this:

```cpp
      /**
       * @brief Create service.
       *
       * @param service_  Google protobuf service instance.
       *
       * @return  True if successful.
      **/
      bool Create(std::shared_ptr<T> service_)
      {
        if (service_ == nullptr) return false;
        m_service = service_;

        const google::protobuf::ServiceDescriptor* service_descriptor = service_->GetDescriptor();
        Create(service_descriptor->full_name());  << the protobuf service name is used to name the server
```

And for the client side it looks like this:

```cpp
    /**
     * @brief Google Protobuf Client wrapper class.
    **/
    template <typename T>
    class CServiceClient : public eCAL::CServiceClient
    {
    public:
      /**
       * @brief Constructor.
       *
      **/
      CServiceClient()
      {
        Create(T::descriptor()->full_name());  << the protobuf service name is used to name the client
      }
```

**Expected behavior**

We could add a ```SetServiceName``` function to overwrite the service name set by the protobuf service. That would not break compatibility.

PR #495, opened 2022-01-27
Fix: setting service name independently from protobuf service name

rex-schilasky:
… name)

**Pull request type**

 

 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Protobuf service instance name can not be set independently from the service name given by the protobuf service description.

Issue Number: #493

**What is the new behavior?**
Protobuf service instance name can be set independently.

- [ ] Yes
- [X] No

### DIFF
--- a/ecal/core/include/ecal/msg/protobuf/server.h
+++ b/ecal/core/include/ecal/msg/protobuf/server.h
@@ -71,6 +71,17 @@ namespace eCAL
       }
 
       /**
+       * @brief Constructor.
+       *
+       * @param service_       Google protobuf service instance.
+       * @param service_name_  Unique service name.
+      **/
+      CServiceServer(std::shared_ptr<T> service_, const std::string& service_name_) : m_service(nullptr)
+      {
+        Create(service_, service_name_);
+      }
+
+      /**
        * @brief Destructor.
       **/
       ~CServiceServer()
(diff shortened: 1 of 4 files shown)


ecal/core/include/ecal/msg/protobuf/server.h
```diff
@@ -92,16 +103,24 @@ namespace eCAL
        * @brief Create service.
        *
        * @param service_  Google protobuf service instance.
+       * @param service_name_  Unique service name.
        *
        * @return  True if successful.
       **/
-      bool Create(std::shared_ptr<T> service_)
+      bool Create(std::shared_ptr<T> service_, const std::string& service_name_ = "")
```
Kerstin-Keller, line 110: @rex-schilasky technically, this is API but not ABI backwards compatible. It would be better to have two functions overloads of `Create`, where the old version calls the new one (which is nicer than to check for `empty()`, too.

```
bool Create(std::shared_ptr<T> service_)
{
  const google::protobuf::ServiceDescriptor* service_descriptor = service_->GetDescriptor();
  return Create(service_, service_descriptor->full_name());
}
```
and 
```
bool Create(std::shared_ptr<T> service_, const std::string& service_name_)
{
  //...
}
```

Kerstin-Keller (changes requested):

ecal/core/include/ecal/msg/protobuf/server.h
rex-schilasky: I mean, service API is changed anyway in 5.10. Do we need to particular stay here ABI compatible ?

Kerstin-Keller: Maybe we don't. But I like explicit overloads better thatn  implicit default `""` string. But the others can give their votes, too 😄

PR #499, opened 2022-01-27
CMake: group files like they appear in the filesystem 

FlorianReimold:
The grouping has previously only been enabled for MSVC. But modern QtCreator now also displays all files in a flat list (which is the proper way for CMake projects that don't use source groups, to be fair).

As I really like my files to be displayed in the same way I initially organized them in the filesystem, the source groups are now also enabled on Linux. For that I had ro remove a bunch of if(MSVC) directives.


Kerstin-Keller (commented):
Actually, CMake can already do it out of the box https://cmake.org/cmake/help/latest/command/source_group.html with the `TREE` subcommand.
Don't know why we never used it, maybe we had to support older cmake versions back then

FlorianReimold:
That TREE option totally looks like what I implemented manually a long time ago. Cmake supports that since 3.8. Looks like this is legacy stuff now and we should completely remove it.

Issue #500, opened 2022-01-28
Remove C wrapped C++ Interfaces


rex-schilasky:
**Is your feature request related to a problem? Please describe.**
Currently many C++ API's are wrapped additional around the C API to guarantee ABI compatibility to older Visual Studio versions (2013 and before).

This additional "C API implementation" is covered by the preprocessor switch `ECAL_C_DLL`. For a publisher this looks like

Native C++ API:
```cpp
    /**
     * @brief Creates this object. 
     *
     * @param topic_name_   Unique topic name. 
     * @param topic_type_   Type name (optional). 
     * @param topic_desc_   Type description (optional). 
     *
     * @return  True if it succeeds, false if it fails. 
    **/
    bool Create(const std::string& topic_name_, const std::string& topic_type_ = "", const std::string& topic_desc_ = "");
```

C++ / C wrapped API:
```cpp
    bool Create(const std::string& topic_name_, const std::string& topic_type_ = "", const std::string& topic_desc_ = "")
    {
      if(m_created) return(false);

      // create new instance
      m_publisher = eCAL_Pub_New();
      if (!m_publisher) return(false);

      // set layer mode if changed class method function
      if (m_tlayer.sm_udp_mc != TLayer::smode_none) eCAL_Pub_SetLayerMode(m_publisher, static_cast<eTransportLayerC>(TLayer::tlayer_udp_mc), static_cast<eSendModeC>(m_tlayer.sm_udp_mc));
      if (m_tlayer.sm_shm    != TLayer::smode_none) eCAL_Pub_SetLayerMode(m_publisher, static_cast<eTransportLayerC>(TLayer::tlayer_shm),    static_cast<eSendModeC>(m_tlayer.sm_shm));

      // set qos
      struct SWriterQOSC qos;
      qos.history_kind       = static_cast<eQOSPolicy_HistoryKindC>(m_qos.history_kind);
      qos.history_kind_depth = m_qos.history_kind_depth;;
      qos.reliability        = static_cast<eQOSPolicy_ReliabilityC>(m_qos.reliability);
      eCAL_Pub_SetQOS(m_publisher, qos);

      // create it
      eCAL_Pub_Create(m_publisher, topic_name_.c_str(), topic_type_.c_str(), topic_desc_.c_str(), static_cast<int>(topic_desc_.size()));

      m_topic_name = topic_name_;
      m_topic_type = topic_type_;
      m_topic_desc = topic_desc_;

      m_created = true;
      return(true);
    }
```

[...]

PR #501, opened 2022-01-28
Feature: remove all c++ / c component wrapper (ECAL_C_DLL)

rex-schilasky:
**Pull request type**

 

 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Currently many C++ API's are wrapped additional around the C API to guarantee ABI compatibility to older Visual Studio versions (2013 and before).

Issue Number: #500

**What is the new behavior?**
Removed them all.

- [X] Yes
- [ ] No

For windows platform only - VS versions before 2015.


FlorianReimold:
Do we want to have this "Feature" in eCAL 5.10? Or do we want to wait until the 5.10 support branch was created

rex-schilasky:
> Do we want to have this "Feature" in eCAL 5.10? Or do we want to wait until the 5.10 support branch was created

Because it's not a "Feature" as you mentioned but cleaning up the API I would like to have this in 5.10.

PR #502, opened 2022-01-28
Added dark mode to GUI Applications

FlorianReimold:
Added a dark mode that one can enable via `View/Theme` in all of our main GUI Applications.

grafik


rex-schilasky (approved):
Looks good to me.

Issue #519, opened 2022-02-15
eCAL Sys: Keep ID-list of groups sorted


FlorianReimold:
The IDs of tasks that are assigned to a group are not properly sorted before being save to the XML file. This causes difficult to read diffs when comparing .ecalsys files.

PR #527, opened 2022-02-22
eCAL Sys now sorts tasks of groups by ID  before saving

FlorianReimold:
eCAL Sys now sorts tasks of groups by ID  before saving


FlorianReimold:
Fixes #519

Kerstin-Keller (approved):

PR #534, opened 2022-02-28
[Mon]: Refactoring to prefer std::find_if

tcoyvwac:
**Pull request type**

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

**What is the new behavior?**

Refactored to use C++ standard library algorithms, which abstracts previous code to a more universal offical-standard abstraction. As the code is now using standard algorithms, there are many great benefits which accompany it.

- [ ] Yes
- [x] No

**Other information**

OS: `Debian Sid`
Compiler: `GCC (Debian 11.2.0-16) 11.2.0`

Very happy to co-author to merge.


FlorianReimold (commented):
I am curious, what is the reason for the std::function object created from the lambda? What I usually do is create a `static` member function inside the class. I assume that is better understandable for most users...

Nevertheless, extracting that code to a function is a good thing.

tcoyvwac:
An audience for the bunny show? Greetings! :smile: :rabbit2: 

Slight caution / Beware though @Kerstin-Keller , as I'm sure you know, a "`static` member function inside the class"  is very much not the same technically as a "a regular, free function, which takes the `selected_items` as an argument".

I'm sure you resonate this, but just want to be clear we see the same technical "definitions" in the quoted texts.

tcoyvwac:
Preview of follow-up PR changeset is here: [tcoyvwac:fix/refactor-duplication-for-external-predicate](https://github.com/continental/ecal/compare/master...tcoyvwac:fix/refactor-duplication-for-external-predicate?diff=unified). :rabbit2:

Special write-up is en-route. :hole:

tcoyvwac:
```cpp
#ifndef RABBITHOLE
// int main() {}            // [1]
auto main() -> int {}    // [2]
#else
// /*const*/ auto main = [](){};      // [3]
/*const*/ auto main = [](){}();    // [4] Incomplete Riddle! :D Challenge: Fix it so it compiles! ;)
#endif
```
https://godbolt.org/z/1WzsWM7jj :rabbit2: :hole: 

Can you all answer the rabbithole's riddle, to explain `[1]` `[2]` `[3]` `[4]`? 

Work together and post your explaination here! :wink:

rex-schilasky:
What is this discussion about ? Kind of Jeopardy ?
Maybe there are better places for rabbitholes ..

tcoyvwac:
:wink: Excellent curiosity and questions lead to the rabbithole! :rabbit2: :hole: 

...So, I understand that the style of answer (in the PR) is indirect / unorthodox. Please understand I know everyone here has a strong knowledge & command in C++, so I do not want to dictate to anyone here their abilities, this is why the purpose & my style of answer is written in this form (a "bunny-show" :rabbit2: ).

For maybe seven+ years now, because of the features of C++(14), two forms of "code design" have been somewhat superseded (in this current PR scenario) with this PR's code-form. There is a lot of help from the compiler to assist in code cleanup, simplicity and layout (and only being added to in further C++ future versions).

From this PR, I am happy there is an agreement among us; _it is in the right direction that the "predicate" code should be "abstracted" into a "function" (of some sort / type)._
As this C++ code has been standardised for years, all of this code builds cleanly and passes CI, so _"technically"_ there are no problems at all with this PR. :heavy_check_mark: 

_If I am understanding right(?),_ these curiosity questions from @FlorianReimold and @Kerstin-Keller seem to be only about the intriguing C++14 style.

In summary:

Why was the code in this PR written in its current form, over some others such as:
* > a `static` member function inside the class 
* > a regular, free function, which takes the `selected_items` as an argument.

---

These are really great questions, but the difficulty (for me), is that this answer could be quite lengthy, depending on background technical knowledge, without a lot of written explanation of how C++ works.

So, personally for me, to skip this possibly misunderstood-offensively background explanation, the only way to keep my answer brief (in this PR), is to actively answer in a form of a puzzle / riddles (a "bunny-show" :rabbit2: ).

"A picture is worth a thousand words" approach.

By gaining the answer to the riddles, we would all be in agreement, seeing-through a common "C++ technical background knowledge lens" and it will (hopefully) also explain why this code-form was chosen. (And in effect, answer the two great curious style questions at once).

_This PR was specifically crafted to catch curious discussion, and talk about its acceptance into this project._ :wink: 

I am happy to add more to this (code-style answer) in a "code-design" sense, but as long as it not misunderstood as _"an offence"_ in some way (in background technical knowledge) to any other contributors (in this open-source project).

FlorianReimold:
Hey tcoyvwac,

This projects aims to provide usable software, so we prefer actual explanations over riddles. At the moment we are at:

Disadvantages:
- Doesn't fit the current code style
- Less readable in the eyes of the people maintaining the code
- Breaks Intellisense and probably other Code Completers as well

Advantages:
- ???

So unless there are some very good reasons to do it this way, we are not going to merge this PR. We may still use the code duplications you found and refactor it in a way that fits to the rest of the code.

Regards
Florian

tcoyvwac:
> Disadvantages:
> 
>     * Doesn't fit the current code style
> 
>     * Less readable in the eyes of the people maintaining the code
> 
>     * Breaks Intellisense and probably other Code Completers as well

Thanks, this is very helpful! I do not know what the side-effects of this C++14 code has for other contributors so before posting further PRs with other similar-styled changes, this clarity really helps. This what I would have hoped the PR would discuss.

```cpp
const auto get_first_non_group_item = [] (auto && items_list)
```

This code can be rewritten as
```cpp
template<typename T1>
auto get_first_non_group_item(T1 && items_list)
```

This will then be more of a match to:
> to a regular, free function, which takes the `selected_items` as an argument.

However:
* it will have to be templated to get the same benefits of `auto`.
* Maybe I misunderstood the question from @Kerstin-Keller, but `selected_items` cannot be "an argument", as `std::find_if()` requires a `callable` as a predicate. As the "datatype" of `selected_items` (which then calls a signature member function called `type()`) is slightly "vague", the easiest way to resolve the situation would be `auto` (a.k.a `template<typename Tx>`). The compiler will only allow compilation with a valid "datatype" with a correct signature member-function called `type()`.


Would all three disadvantages disappear if the code was rewritten in a: `template<typename Tx>` free-function style?

FlorianReimold:
I still don't see a reason why not to just create a static member function and place the code in there. This would fit to the current style perfectly. Doing templates and auto&& seems unnecessary and may even lead to errors, as the compiler would accept data as input that cannot be handled by the function.

tcoyvwac:
Hi @FlorianReimold , so I have rewritten the PR in many different forms. 

| Style (C++03) | Code Form |
|-|-|
| Classic | [[Static mem-fn]][1]<br/>[[Free-function]][5] |

| **Style (C++11)** | **-** |
|-|-|
| Predicate as an external function | [[Static mem-fn]][2] |
| Predicate as a local-named closure object | [[Static mem-fn]][3]<br/>[[Free-function]][6]  *|

| ~~**Bonus Style?**~~ | **-** |
|-|-|
| ~~auto returns & templated parameter for fake auto~~ | ~~[[Static mem-fn]][4]~~ |

---

\* = My personal choice is: [C++11 style, free-function with the find-if predicate as a local-named closure object.][6] 

This is over the static-member-function form, because the static-member-function form affects the class (header) API.

This personal choice would be quite close to @Kerstin-Keller form of: 
> "...a regular, free function, which takes the `selected_items` as an argument".

---

I am quite flexible in styles so please feel free to leave adjust-suggestions!

[1]: https://github.com/continental/ecal/compare/master...tcoyvwac:fixx/draft/refactor-duplication-with-std-algorithm-find_if-static-memfun-cpp03-classic

[2]: https://github.com/continental/ecal/compare/master...tcoyvwac:fixx/draft/refactor-duplication-with-std-algorithm-find_if-static-memfun-cpp11-classic-auto-returns

[3]: https://github.com/continental/ecal/compare/master...tcoyvwac:fixx/draft/refactor-duplication-with-std-algorithm-find_if-static-memfun-inline-lambda-pred-auto-returns

[4]: https://github.com/continental/ecal/compare/master...tcoyvwac:fixx/draft/refactor-duplication-with-std-algorithm-find_if-static-memfun-inline-lambda-pred-auto-returns-auto-parameters

[5]: https://github.com/continental/ecal/compare/master...tcoyvwac:fixx/draft/refactor-duplication-with-std-algorithm-find_if-free-function-cpp03-classic

[6]: https://github.com/continental/ecal/compare/master...tcoyvwac:fixx/draft/refactor-duplication-with-std-algorithm-find_if-free-function-inline-lambda-pred-auto-returns

PR #547, opened 2022-03-09
minor doc fixes for building from sources (ubuntu)

rex-schilasky:
Fix some minor documentation issues for Ubuntu build

**Pull request type**

 

 

- [X] Documentation content changes

**What is the new behavior?**

Some minor fixes - ubuntu only. Moved installing python dependencies in Ubuntu16/18/20 section. Add some comments regarding python CMake options.

### DIFF
--- a/doc/rst/development/building_ecal_from_source.rst
+++ b/doc/rst/development/building_ecal_from_source.rst
@@ -154,13 +154,6 @@ First check out the eCAL repository and all of the submodules:
 
       sudo apt-get install git cmake doxygen graphviz build-essential zlib1g-dev qt5-default libhdf5-dev libprotobuf-dev libprotoc-dev protobuf-compiler libcurl4-openssl-dev
 
-#. If you plan to create the eCAL python language extension (here as an example for the python 3.6 version):
-
-   .. code-block:: bash
-
-      sudo apt-get install python3.6-dev python3-pip
-      python3 -m pip install setuptools
-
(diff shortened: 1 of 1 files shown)


doc/rst/development/building_ecal_from_source.rst
```diff
@@ -173,7 +166,14 @@ First check out the eCAL repository and all of the submodules:
 |fa-ubuntu| Ubuntu 16/18/20 build
 ---------------------------------
 
-#. Compile eCAL with the following options:
+#. If you plan to create the eCAL python language extension (here as an example for the python 3.8 version):
+
+   .. code-block:: bash
+
+      sudo apt-get install python3.8-dev python3-pip
```
FlorianReimold, line 173: Hm, moving the dependencies from the "Install dependencies" Section into the "Build" Section doesn't really make sense. Additionally, those dependencies will only work For Ubnutu 20.04, das it specifies python 3.8.

Maybe we should add an entire "Build python language binding" section?

FlorianReimold (changes requested):

doc/rst/development/building_ecal_from_source.rst
rex-schilasky: Yes maybe. In the build section it's not fitting that is true. But for 20.04 it was missing in general. We describe it for 18.04 (as now) with the 3.6 version and additional for 20.04 with 3.8 under "install dependencies". What do you mean ?

FlorianReimold (approved):

Issue #542, opened 2022-03-04
Dockable eCAL Mon Plugins


FlorianReimold:
Currently, eCAL Mon Plugins are always displayed in their own Window, that pops up when double-clicking a channel, while this is nice for introspection of a specific topic, it doesn't really allow to create a configurable view except from manually aligning many windows afterwards.

As we already have the Monitor Plugin API and there are quite sophisticated third party plugins, it is proably a good idea to reuse those plugins to form some kind of single-window visualization that the user can easily configure

A frame for the plugins would have to expose the important Monitor Plugin API parameters in a GUI:
- plugin
- topic name

The tool should then also be able to save and load different configurations. It is probably the easiest to integrate everything directly into the eCAL Monitor, as the user can then also use the topic list, log windows etc. just like his monitor plugins.

Maybe it is worth the effort to also switch from the QDockingWidget that is currently used to [Qt-Advanced-Docking-System](https://github.com/githubuser0xFFFF/Qt-Advanced-Docking-System), as sometimes it is really cumbersome to place the Docking Widgets in their container.

PR #550, opened 2022-03-09
The monitor can now be very very small.

FlorianReimold:
It can now be unreadably small. I did it to support #542, as it always bothers me when I cannot at least temporarily resize docked widgets to a very small state in order to move them around.

grafik


hannemn (approved):
Awesome feature ;-)

Issue #553, opened 2022-03-11
cmake_functions: not able to use qt_add_windeployqt_postbuild 


oanadumitruc:
**Current behavior:**
We use eCAL as a dependency and to deploy some QT related dependencies we use qt_add_windeployqt_postbuild. 
Unfortunately `qt_add_windeployqt_postbuild  `function does not find path for` qt_windeployqt_threadsafe_cmake.bat.in`.
Here is the error:
```
CMake Error: File /qt_windeployqt_threadsafe_cmake.bat.in does not exist.
CMake Error at C:/eCAL/cmake/qt/qt_windeployqt.cmake:49 (configure_file):
  configure_file Problem configuring file
Call Stack (most recent call first):
  CMakeLists.txt:161 (qt_add_windeployqt_postbuild)
```
See link below:

    https://github.com/continental/ecal/blob/master/thirdparty/cmake_functions/qt/qt_windeployqt.cmake#L49

**Proposal:**

- remove `PARENT_SCOPE `from line below:
   https://github.com/continental/ecal/blob/master/thirdparty/cmake_functions/qt/qt_windeployqt.cmake#L20

**Desktop:**
 - OS: Windows
 - Version 5.9.5

MrMoose:
I have run into the same issue trying to build it. To me, ecal is becoming part of a (primarily windows) script based build process. All dependencies are built as packages, installed individually, and then CMAKE_PREFIX_PATH is set to find them.

My solution to this was to explicitly build CMakeFunctions which looks like a submodule but isn't.
You can build from  CMake source path `thirdparty/cmakefunctions/cmake_functions` into any tmp build dir, run cmake and build there, then install CMakeFunctions. Before you proceed to build ecal, include the install dir where you installed CMakeFunctions to CMAKE_PREFIX_PATH. Then it should work.

PR #560, opened 2022-03-15
Update qt_windeployqt.cmake

oanadumitruc:
**Pull request type**

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #553

**What is the new behavior?**

-
-
-

- [ ] Yes
- [x] No

**Other information**

### DIFF
--- a/thirdparty/cmake_functions/qt/qt_windeployqt.cmake
+++ b/thirdparty/cmake_functions/qt/qt_windeployqt.cmake
@@ -17,7 +17,7 @@
 # ========================= eCAL LICENSE =================================
 
 
-set (qt_windeployqt_cmake_path ${CMAKE_CURRENT_LIST_DIR} PARENT_SCOPE)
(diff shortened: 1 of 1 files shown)


FlorianReimold:
The issue here is that qt_windeployqt was actually meant for internal use, where the `parent_scope` is required due to the cmake layout. That was why I said that we may just not install it at all.

If we want the file to appear in the installation and also be usable, we must find a universal way, probably by either making the parent_scope work in the installation or by moving the file so we don't need parent scope internally any more.

oanadumitruc:
Hi Florian, 

I guess I was too quick and over confident about the change. :)

Nevertheless, I agree, whatever is available/public, the user should have the possibility to use it. I did not know you also plan not to have it available in the installer, that's also a way to solve it.

Then I think we can close this pr. :) 

Best regards,
Oana

PR #568, opened 2022-03-21
Created eCAL Get Config API functions

FlorianReimold:
I created 3 API functions that can access some parts of the eCAL Configuration.
The plan is to create more, but I wanted to get a review first. I had to move the ecal_config.h to the public include directory.

Does anything speak against this approach?


Kerstin-Keller (commented):
I am not 100% sure.
 I would at least split `ecal_config.h` file, put your functions in `ecal/ecal_config.h`, but the `CConfig` (e.g. it's very implementation specific) should stay in an internal implementation. (maybe `src/ecal_config_reader.h`).
 
 But also I am not sure if I would make the functions `ECAL_API` functions. Are users supposed to use them? E.g. if only applications want to use them, we can make them part of an internal library, not necessarily part of the core API.

FlorianReimold:
> But also I am not sure if I would make the functions `ECAL_API` functions. Are users supposed to use them? E.g. if only applications want to use them, we can make them part of an internal library, not necessarily part of the core API.

The purpose of this PR is to make the functions user-accessible. For eCAL-Core-internal use we already have functions to access them. The goal is to access the exact configuration that the eCAL Core is currently using. And the only way to do that is by exposing them in the API.

Kerstin-Keller:
Ok. But then still, i would split the `ecal_config.h` file into two files, as mentioned above.

rex-schilasky (approved):

ecal/core/include/ecal/ecal_config.h
```diff
@@ -0,0 +1,107 @@
...
+
+    //ECAL_API std::string GetEcalSysFilterExcludeList          ();
+
+    /////////////////////////////////////
+    // publisher
+    /////////////////////////////////////
+
+    ECAL_API int         GetPublisherInprocMode               ();
+    ECAL_API int         GetPublisherShmMode                  ();
+    ECAL_API int         GetPublisherTcpMode                  ();
```
Kerstin-Keller, line 97: If we are modifying this. The `ints` are magic numbers. Can we introduce an enum?
```
enum class LayerMode
{
   OFF = 0;
   ON = 1;
   AUTO = 2;
}
```

Kerstin-Keller (changes requested):
I scanned over the functions and they seem fine except for the Publisher Modes. Here I would really introduce an enum to make the magic numbers more clear.

ecal/core/include/ecal/ecal_config.h
Kerstin-Keller: @rex-schilasky @FlorianReimold what do you think?

FlorianReimold: I think it would create more consise code.

@rex-schilasky: What happenes, when someone inputs an invalid number there? Will `eCALPAR()` return the invalid number, or will it return the default value?

Kerstin-Keller:
I merged the master into this branch and resolved the merge errors regarding ecal_config.cpp

ecal/core/include/ecal/ecal_config.h
rex-schilasky: I changed the API of the GetPublisherXYMode functions that way

```cpp
    ECAL_API TLayer::eSendMode GetPublisherInprocMode               ();
    ECAL_API TLayer::eSendMode GetPublisherShmMode                  ();
    ECAL_API TLayer::eSendMode GetPublisherTcpMode                  ();
    ECAL_API TLayer::eSendMode GetPublisherUdpMulticastMode         ();
```

with

```cpp
    enum eSendMode
    {
      smode_none = -1,
      smode_off  = 0,
      smode_on,
      smode_auto
    };
```
@Kerstin-Keller does that fits ?

Kerstin-Keller: @rex-schilasky 
Any reason against using an `enum class?` -> type safety? 

What does `smode_none` mean?

I can see that the functions have not been added to the `C` interface. Are they supposed to be? What about Python?
(Maybe those can be different issues).

FlorianReimold:
OK, I checked out the latest commit.

What I don't like: When somebody writes an invalid value in the `ecal.ini`, that value is not interpreted but simply casted to the enum leading to values that aren't even in the Enum. There is no error checking.
What I like: I didn't know that there already was an enum for that in the public API, so it is nice to reuse it. As the eCAL Core handles it the same way (also without any error checking) it is at least consistent, which is even more important than the error checking in my opinion. The error checking would have to be done in the core anyways, not in the Config API. So in my opinion this is fine.

_Do we want to keep the behaviour in the eCAL Core? Or do we want to interpret invalid values as `none` or `off`?_

**Regarding other Language Bindings**:
I think this eCAL Config API should be in all language bindings, both C and Python.

ecal/core/include/ecal/ecal_config.h
rex-schilasky: Wouldn't that be an API break ? `smode_none` means undefined. In the past we were forced to use that "classic enums" because of that additional C API variant for `CPublisher`.
If we want to change that, then better in a new issue and for all public enum types.

FlorianReimold: Yes, scoping the enum with `enum class` would break the API. The benefits of doing that now are negligible and don't justify an API break. Especially when it's not compatible with the C interface. Beside from that I agree though that scoped enums are usually the proper way.

PR #582, opened 2022-03-29
warns if NPCAP is set via configuration (ecal.ini) but not linked in …

rex-schilasky:
…by CMake

**Pull request type**

 

 

- [X] Bugfix

If the user wants to use NPCAP (configure npcap_enabled=true) but did not link the core against NPCAP by activating the CMake option eCAL will fallback on default behavour silently.

Issue Number: N/A

**What is the new behavior?**
A warning is shown.


rex-schilasky:
Just a minor fix, I checked that on my machine and it's working fine.

PR #586, opened 2022-03-30
Doc: Disabled Python apidoc on macOS

### LINKED ISSUE #584: Fix documentation build on macOS
At the moment, the documentation build on macOS is broken. It build process appears to crash with a segfault.

grafik

This issue persists since we added the python documentation. So maybe we should simply disable that one for macOS.

FlorianReimold:
The python apidoc caused sphinx-build to crash on macOS.

Fixes #584

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -11,8 +11,17 @@ jobs:
     runs-on: macos-latest
 
     steps:
+    - name: Downgrade Python version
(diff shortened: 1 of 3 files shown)


Kerstin-Keller (commented):
Hmhm I don't understand what implication it has to exclude the `autodoc` and `apidoc` on Mac. Is the documentation then incomplete. However, nice that it doesn't crash anymore 😄

PR #587, opened 2022-03-30
Config: Set network_mode=false by default

FlorianReimold:
- Set `network_mode = false` by default in `ecal.ini`
- Documented that since eCAL 5.10 you will have to also set the network mode to `true` again for it to work properly

### DIFF
--- a/ecal/core/src/ecal_def.h
+++ b/ecal/core/src/ecal_def.h
@@ -56,21 +56,13 @@
 /*                                     network settings                                       */
 /**********************************************************************************************/
 /* network switch */
-#ifdef CFG_FORCE_DEFAULT_LOCAL
 #define NET_ENABLED                                 false
-#else /* CFG_FORCE_DEFAULT_LOCAL */
-#define NET_ENABLED                                 true
-#endif /* CFG_FORCE_DEFAULT_LOCAL */
 
 /* eCAL udp multicast defines */
 #define NET_UDP_MULTICAST_GROUP                    "239.0.0.1"
 #define NET_UDP_MULTICAST_MASK                     "0.0.0.15"
 #define NET_UDP_MULTICAST_PORT                     14000
-#ifdef CFG_FORCE_DEFAULT_LOCAL
-#define NET_UDP_MULTICAST_TTL                          0
-#else /* CFG_FORCE_DEFAULT_LOCAL */
 #define NET_UDP_MULTICAST_TTL                          3
-#endif /* CFG_FORCE_DEFAULT_LOCAL */
 #define NET_UDP_MULTICAST_PORT_REG_OFF                 0
 #define NET_UDP_MULTICAST_PORT_LOG_OFF                 1
 #define NET_UDP_MULTICAST_PORT_SAMPLE_OFF              2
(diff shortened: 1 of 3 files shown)


Kerstin-Keller (changes requested):
@FlorianReimold probably you also need to modify https://github.com/continental/ecal/blob/b8a681d7504828d9de7dc592a388a628ffe34a4c/ecal/core/src/ecal_def.h#L62
Also I don't know whats up with the `CFG_FORCE_DEFAULT_LOCAL` variable, if we need to set something in CMake (ok, its only reference is in that `ecal_def.h` file, so I would remove it altogether).

FlorianReimold:
Here is what CFG_FORCE_DEFAULT_LOCAL does (refactored):

```cpp
#ifdef CFG_FORCE_DEFAULT_LOCAL
  #define NET_ENABLED                  false
  #define NET_UDP_MULTICAST_TTL        0
  #define NET_UDP_MC_REC_ENABLED       false
  #define PUB_USE_UDP_MC               0
#else
  #define NET_ENABLED                  true
  #define NET_UDP_MULTICAST_TTL        3
  #define NET_UDP_MC_REC_ENABLED       true
  #define PUB_USE_UDP_MC               2
#endif
```

so basically it
- sets the multicast ttl to 0 (which is not required when network mode is turned off. I also decided to leave it at 2 in the `ecal.ini` because it always is a case for errors when someone sets network to ON but forgets to increase the multicast ttl)
- disables sending and receiving UDP traffic

it does not disable sending and receiving TCP traffic though, which is inconsistent.

I would vote for either removing the `CFG_FORCE_DEFAULT_LOCAL` entirely or restricting it to only modify the `NET_ENABLED` define

Kerstin-Keller:
I would vote to remove it completely. At somepoint we used to have two different ini files, which one could be chosen to install.
I  mean this is the fallback in case no ini file is found, or the value is not present in the ini file.
Let's remove it, and at some point, we can revisit it. We could then also consider to have a single source of truth (e.g. to generate (CMake) the ini file & the def.h file)

ecal/core/src/ecal_def.h
```diff
@@ -87,11 +79,7 @@
 #define NET_TCP_REC_ENABLED                         true
 #define NET_SHM_REC_ENABLED                         true
 
-#ifdef CFG_FORCE_DEFAULT_LOCAL
-#define NET_UDP_MC_REC_ENABLED                      false
-#else /* CFG_FORCE_DEFAULT_LOCAL */
 #define NET_UDP_MC_REC_ENABLED                      true
```
Kerstin-Keller, line 82: @FlorianReimold but shouldn't we have kept the `false` value?

FlorianReimold: I wanted it to be the same as the default `ecal.ini`. When `network` is `false`, the multicast receiver is should not be used anyways, so it is unnecessary to turn it off separetely

Kerstin-Keller: ah sorry I confused it with the `NET_ENABLED` parameter....

Kerstin-Keller (approved):

rex-schilasky (approved):

PR #590, opened 2022-04-01
Rec GUI: Fixed "..." convention in "Open File"

FlorianReimold:
Changed "Open File" to "Open File..."
It is SO MUCH BETTER NOW!!!!1111

grafik


Kerstin-Keller (approved):
😉

PR #595, opened 2022-04-07
Refactoring:

Kerstin-Keller:
Split udp_receiver into a base class and asio & npcap implementations


Kerstin-Keller:
The question is:
why is there *another* layer of indirection 😄  
`CReceiver` -> `CUDReceiver` -> `CUDPReceiverImpl / Pcap / Asio`
(I guess before there was UDP and TCP receivers?) 
I guess, maybe we could remove one of these layers, but maybe not atm.
But I guess I can make them override.

As for the CMake thing, I wanted to restructure the whole thing and use the tree functionality, but in a different PR. 
But as long, I guess I can do it the way it is done now.

FlorianReimold (approved):

ecal/core/src/io/udp_receiver_asio.cpp
```diff
@@ -0,0 +1,179 @@
...
+    {
+      std::cerr << "CUDPReceiver: Setting broadcast and unicast option true is not allowed." << std::endl;
+      return;
+    }
+
+    // create socket
+    asio::ip::udp::endpoint listen_endpoint(asio::ip::udp::v4(), static_cast<unsigned short>(attr_.port));
+    {
+      asio::error_code ec;
+      m_socket.open(listen_endpoint.protocol());
```
rhd, line 41: I think this should be:

```cpp
       m_socket.open(listen_endpoint.protocol(), ec);
```

ecal/core/src/io/udp_receiver_asio.cpp
```diff
@@ -0,0 +1,179 @@
...
+  bool CUDPReceiverAsio::AddMultiCastGroup(const char* ipaddr_)
+  {
+    if (!m_broadcast && !m_unicast)
+    {
+      // join multicast group
+      {
+        asio::error_code ec;
+        m_socket.set_option(asio::ip::multicast::join_group(asio::ip::make_address(ipaddr_)), ec);
+        if (ec)
+          std::cerr << "CUDPReceiver: Unable to join multicast group: " << ec.message() << std::endl;
```
rhd, line 96: Hi, sorry to comment on a merged PR - but why is this error not being returned to the caller? This pattern happens multiple times in this file.

I think it should be something like:

```cpp
        if (ec)
        {
          std::cerr << "CUDPReceiver: Unable to join multicast group: " << ec.message() << std::endl;
          return(false);
        }
```

Kerstin-Keller: @rex-schilasky @FlorianReimold can you take a closer look at both comments? I only did refactoring here 😉

rex-schilasky: I will do so.

rex-schilasky:
> 1. Can we make udp_receiver.h `override`-correct (when we are already modifying it 😄)
> 2. Can we add udp_receiver_npcap.h/cpp to the appropriate source group? currently all files are sorted like they appear in the filesystem except these two.

I will take care for 2.

PR #596, opened 2022-04-07
Reference ECAL_PROJECT_ROOT instead of CMAKE_SOURCE_DIR...

Kerstin-Keller:
to allow eCAL main CMakeLists.txt to be included by other CMakeFiles.

**Pull request type**
- [x] Build related changes

We cannot do `add_subdirctory(ecal)` because of said variables

**What is the new behavior?**
It will be possible to do  `add_subdirctory(ecal)`.


FlorianReimold (approved):
Looks fine to me. I only briefly scanned the code, but as long as the CI still successfully builds and packs everything this PR should be OK.

PR #597, opened 2022-04-07
Fixed almost all MSVC 2019 Warnings

FlorianReimold:
Fixed almost all MSVC 2019 Warnings:
- Rec
- Sys
- Mon
- Doc Core
- Samples

### DIFF
--- a/lib/ecal_utils/src/win_cp_changer.cpp
+++ b/lib/ecal_utils/src/win_cp_changer.cpp
@@ -31,7 +31,7 @@ namespace EcalUtils
   WinCpChanger::WinCpChanger(const int target_cp)
   {
     initial_cp = GetConsoleOutputCP();
-    SetConsoleOutputCP(CP_UTF8);
+    SetConsoleOutputCP(target_cp);
   }
 
   // Destructor
--- a/app/mon/mon_gui/src/widgets/visualisation_widget/visualisation_widget.h
+++ b/app/mon/mon_gui/src/widgets/visualisation_widget/visualisation_widget.h
@@ -37,7 +37,7 @@
 
 #ifdef _MSC_VER
 #pragma warning(push)
-#pragma warning(disable: 4146)
+#pragma warning(disable: 4100 4146 4505)
 #endif
 #include "ecal/pb/monitoring.pb.h"
 #ifdef _MSC_VER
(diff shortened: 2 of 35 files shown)


FlorianReimold:
Last thing to fix is this one:

https://github.com/continental/ecal/blob/cb9956d52449e4a04c728652b6da08df53ab7939/ecal/core/src/service/ecal_tcpclient.cpp#L299

The function seems very sketchy and doesn't appear to do what it is expected to do. The timeout parameter is unused.
@rex-schilasky / @brakmic-aleksandar : How shall I fix this? What is the function supposed to do?

rex-schilasky:
> Last thing to fix is this one:
> 
> https://github.com/continental/ecal/blob/cb9956d52449e4a04c728652b6da08df53ab7939/ecal/core/src/service/ecal_tcpclient.cpp#L299
> 
> The function seems very sketchy and doesn't appear to do what it is expected to do. The timeout parameter is unused. @rex-schilasky / @brakmic-aleksandar : How shall I fix this? What is the function supposed to do?

This function is not implemented fully. It's doing (as the name says) a receive for a server response asynchronously (returns immediately). The timeout parameter is not supported yet, see API doc

https://github.com/continental/ecal/blob/87fefd756b0542da3c40a1d26799d5f8a7d3906c/ecal/core/include/ecal/ecal_client.h#L134

FlorianReimold:
eCAL now is _basically_ warning free in Visual Studio and GCC11

grafik

Kerstin-Keller:
I guess we could check the python version and call only if python <= 3.6? (via `#if`)?

lib/ecal_utils/src/win_cp_changer.cpp
```diff
@@ -31,7 +31,7 @@ namespace EcalUtils
   WinCpChanger::WinCpChanger(const int target_cp)
   {
     initial_cp = GetConsoleOutputCP();
-    SetConsoleOutputCP(CP_UTF8);
+    SetConsoleOutputCP(target_cp);
   }
```
Kerstin-Keller, line 36: This is actually a functional change? If you want to fix the warnings only I would do `const int /*target_cp*/` and do the behavioral change in a different PR? What does thins change imply (e.g. what is passed as parameter in it's usage).?

FlorianReimold: It is a functional change, but it was always supposed to function that way. I just didn't realize the function completely ignores the parameter because I only use it to switch to UTF-8 anyways.

FlorianReimold:
I fixed the Python Warnings in a separete Branch / PR: #606 

The reason is that I want to Squash-merge this commit but cherry-pick the Python changes into 5.9 and 5.8. That way we are safe to also compile 5.9 against python 3.11 in the future.

rex-schilasky (approved):

Issue #456, opened 2022-01-11
Dynamic Service Client


bsirang:
**Is your feature request related to a problem? Please describe.**
I'd like to develop a generic utility to issue service calls to arbitrary service servers since it is a useful tool for testing and development.

I did something similar with publishing by using this API `eCAL::Util::GetDescription(topic);` to get the protobuf description during runtime so I can create a dynamic publisher.

**Describe the solution you'd like**
Ideally I'd like some APIs available to be able to query the service definitions at runtime to be able to call methods dynamically.

**Additional context**
I'm trying to write a command line utility that would take a JSON string for the request message and a method name to then convert into the request message and call the service with.

rex-schilasky:
Hi @bsirang,

that sounds like a nice generic idea. Currently the eCAL registration layer is transporting the service descriptor only, without the descriptors for request and response types. You check this by running the `monitoring_reg` sample.

![image](https://user-images.githubusercontent.com/49162693/150140305-ad254345-4795-4513-9b79-c6160a88276b.png)

So before we can implement this feature we need to add the missing information here I think. What do you mean ?

Best, ReX

bsirang:
Hi @rex-schilasky yes that sounds correct. I think it would be a good idea to include the missing information there first, and then that would enable what I'm describing to be possible.

rex-schilasky:
Hi @bsirang, I created a primitive prototype now that is extracting the type information out of a service description. You can check the demo repository [here](https://github.com/rex-schilasky/ecal_service_dyntypes).

The output looks like this

![ecal_service_dyntypes_log](https://user-images.githubusercontent.com/49162693/161216225-9c4c9639-2fbe-4f31-b3e2-381f719e1d5f.png)

So the service method types are _human readable_ protocol buffer definition. Does that fit to your needs or do you needs protocol buffer descriptors like we do it for topic type descriptors ?

Next step would be to add these information to the monitoring layer and finally to have a `eCAL::Util` function retrieving that information. Currently I have no idea how to signature of such a function could look like. Maybe you have an idea :-). We would need the service name, the method name and get the request and response type descriptors ?

Kerstin-Keller:
@rex-schilasky wouldn't it be more in line with the rest of our information if they were actual descriptors instead of the human readable buffer definitions?
I guess it would be helpful if we would want to reflect on these types?

[...]

PR #598, opened 2022-04-07
dynamic service client

rex-schilasky:
**Pull request type**

 

 

- [X] Feature

Service method message type descriptions are not part of the registration and are not available by any kind of user API.

Issue Number: #456

**What is the new behavior?**
Service method message type descriptions can be set using a new CServiceServer API method and are added to the registration information of a eCAL server entity. A new eCAL::Util API function will make this information available.


Kerstin-Keller (commented):
In general, this looks quite good!
However, constructing the decoder and the prototypes are in general quite expensive operations from what I know (haven't debugged it though).
It might make sense to construct an object (with `msg_desc_` and `msg_type`), and then have the two conversion functions as member functions.
I guess it depends on how frequently you will want to convert between JSON and serialized messages.

contrib/ecalproto/include/ecal/protobuf/ecal_proto_dyn.h
```diff
@@ -105,6 +105,15 @@ namespace protobuf
...
     google::protobuf::Message* GetProtoMessageFromDescriptor(const google::protobuf::FileDescriptorProto& proto_desc_, const std::string& msg_type_, std::string& error_s_);
 
+    /**
+      * @brief Create message from serialized proto descriptor string.
+      *
+      * @param       msg_desc_    Serialized message descriptor.
+      * @param       msg_type_    Type name.
+      * @param [out] error_s_     Error string.
+    **/
+    google::protobuf::Message* GetProtoMessageFromDescriptor(const std::string& msg_desc_, const std::string& msg_type_, std::string& error_s_);
```
FlorianReimold, line 115: This looks like it returns a pointer to some heap-allocated memory that the user has to manage. We should document that the user has to free that memory!

Like here:
https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.message#Message.New.details

We should also document that this function can return a `nullptr` in certain cases.

A better way would probably be to return a smart pointer to prevent the user from leaking memory.

ecal/core/include/ecal/ecal_util.h
```diff
@@ -136,41 +136,107 @@ namespace eCAL
...
+
+    /**
+     * @brief Gets type name of the specified topic.
+     *
+     * @param topic_name_   Topic name.
+     * @param topic_type_   String to store type name.
+     *
+     * @return  True if succeeded.
+    **/
+    [[deprecated]]
```
FlorianReimold, line 208: We should tell the user which function to use now instead of this deprecated one

ecal/core/include/ecal/msg/protobuf/server.h
```diff
@@ -192,6 +209,34 @@ namespace eCAL
...
+
+        std::string file_desc_s = file_desc->DebugString();
+        google::protobuf::FileDescriptorProto file_desc_proto;
+        if (!msg_decoder.GetFileDescriptorFromString(file_desc_s, &file_desc_proto, error_s_)) return false;
+
+        google::protobuf::FileDescriptorSet pset;
+        google::protobuf::FileDescriptorProto* pdesc = pset.add_file();
+        pdesc->CopyFrom(file_desc_proto);
+
+        auto req_msg = msg_decoder.GetProtoMessageFromDescriptorSet(pset, type_name_, error_s_);
```
FlorianReimold, line 227: I guess this causes a memory leak, right? Or is it deleted somewhere?

FlorianReimold (changes requested):

ecal/core/include/ecal/ecal_util.h
FlorianReimold: Oh, and it would be really cool to have something like "Deprecated since 5.10" somewhere

contrib/ecalproto/include/ecal/protobuf/ecal_proto_dyn.h
rex-schilasky: I can improve the documentation here. I just added one more member function to `eCAL::CProtoDynDecoder`. Some of these functions return this kind of raw message pointer to a `google::protobuf::message`. The memory is handled inside `CProtoDynDecoder` and is valid as long the decoder instance exists (it's using an instance of `google::protobuf::DescriptorPool` internally).

We could make a copy using suggested `Message::New()`, but maybe it's better to just better document the right usage of `eCAL::CProtoDynDecoder`.

ecal/core/include/ecal/msg/protobuf/server.h
rex-schilasky: It should be deleted with the destruction of `pset`.

ecal/core/include/ecal/ecal_util.h
rex-schilasky: Done, I add some more documentation in ecal_util.h

ecal/core/include/ecal/msg/protobuf/server.h
FlorianReimold: The pointer goes out of scope without calling delete.

FlorianReimold (approved):

PR #601, opened 2022-04-11
udp receiver error handling fixed

rex-schilasky:
**Pull request type**

 

- [X] Bugfix

Error states are not handled correctly in the udp low level receiver implementation. See this [discussion](https://github.com/continental/ecal/pull/595#discussion_r847012314).

**What is the new behavior?**
Error states are checked and functions return correctly (thanks @rhd)

### DIFF
--- a/ecal/core/src/io/udp_receiver_asio.cpp
+++ b/ecal/core/src/io/udp_receiver_asio.cpp
@@ -21,7 +21,9 @@
 #include <io/udp_receiver_asio.h>
 
 namespace eCAL {
-
+  ////////////////////////////////////////////////////////
+  // Default ASIO based receiver class implementation
+  ////////////////////////////////////////////////////////
   CUDPReceiverAsio::CUDPReceiverAsio(const SReceiverAttr& attr_) :
     CUDPReceiverBase(attr_),
     m_broadcast(attr_.broadcast),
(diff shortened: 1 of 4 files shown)


Kerstin-Keller (commented):
I am a bit sceptical, especially with regards to the changes in `CUDPReceiverAsio`.
Now the function returns from the constructor.
What happends if it is not constructed? Can anything be done with the object?

Wouldn't it be `safer` to `throw` in the constructor, than to have a partially initialized object?
What implications does it have, it e.g. loopback cannot be set on the object?

rex-schilasky:
> I am a bit sceptical, especially with regards to the changes in `CUDPReceiverAsio`. Now the function returns from the constructor. What happends if it is not constructed? Can anything be done with the object?
> 
> Wouldn't it be "safer" to `throw` in the constructor, than to have a partially initialized object? What implications does it have, it e.g. loopback cannot be set on the object?

I use exceptions now when construction ends in an invalid socket. In other cases (setting an option) standard cerr is used.

FlorianReimold:
If the constructor can throw, we must surround it with `try...catch`. Otherwise the whole app will crash. It should never be the desired behaviour to have everything crashing.

Kerstin-Keller:
Or we need to document, that we throw. But more likely it needs to be handled inside eCAL.
I guess the questions are what the implications are. What does it mean for eCAL and the communication if the sockets cannot be created? 

Also my comment was not meant that this is directly implement, but to encourage to think about it at least.

rex-schilasky:
Seems to me that we need to create a more general strategy how to handle errors inside constructors. For now I will remove the exceptions again and just log an error.

FlorianReimold (approved):

Issue #602, opened 2022-04-11
Replace all eCALPar instances with new Config functions


Kerstin-Keller:
#568 introduced new functions.
All `eCALPar` call should be replaced with new function calls.

PR #603, opened 2022-04-11
Feature/use config api

Kerstin-Keller:
**Pull request type**

Solves #602

Please review carefully, as containes many copy & replace refactoring, which is always error prone.

- One function was renamed, as the function name was missleading.
- One function implemenatation was switched, as it was a  copy&paste error of #602


FlorianReimold (approved):

FlorianReimold:
Approved the current changes as I didn't finde any issues. Do we also want to use this PR to include the new API in eCAL Apps? there are a few occurences in Sys, Mon, Rec where the apps either read the ecal.ini themselfes or parse the System Information Dump.

Kerstin-Keller:
Actually, I would make a separate issue / PR for incorporating the changes into the apps. They might be a bit more complicated / less copy&replace?

But first we should finalize #604.

rex-schilasky (approved):

PR #605, opened 2022-04-11
Remove hostid check

hannemn:
**Pull request type**

 

 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

In order to exchange data payload via shared memory the host id has to match on the respective nodes. Unfortunately this check will lead to problems once containerized eCAL nodes (e.g. in docker) would like to communicate on shared memory, but does not have the same host id, although they are on the same host.

Issue Number: N/A

**What is the new behavior?**

The host id check will be removed, so the nodes can already communicate via shared memory in case of a matching hostname. Furthermore the eCAL::Process::GetHostID() API function has been tagged as deprecated and can be removed completely in a future version of eCAL. In order to avoid a "deprected warning" by internal functions like CDataReader that make use of eCAL::Process::GetHostID(), the temporary function eCAL::Process::internal::GetHostID() has been introduced.

- [ ] Yes
- [X] No

**Other information**

### DIFF
--- a/ecal/core/include/ecal/ecal_process.h
+++ b/ecal/core/include/ecal/ecal_process.h
@@ -59,6 +59,7 @@ namespace eCAL
      *
      * @return  Host id or zero if failed.
     **/
+    [[deprecated]]
     ECAL_API int GetHostID();
 
     /**
--- a/ecal/core/src/ecal_process.cpp
+++ b/ecal/core/src/ecal_process.cpp
@@ -28,6 +28,7 @@
 #include "ecal_register.h"
 #include "ecal_reggate.h"
 #include "ecal_globals.h"
+#include "ecal_process.h"
 
 #include <array>
 #include <chrono>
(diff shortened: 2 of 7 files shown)


Kerstin-Keller (commented):
I am not sure if I would yet deprecate the `HostID` function, even though it's not used anymore to distinguish or not for local_host.
I have the feeling that it might come in handy for other computations regarding docker images / communications.

How do services behave? Do we need to distinguish in certain scenarios, even though not necessarily for SHM communication?

hannemn:
From my point of there is no reason why this function has to be longer part of the public api. We introduced this function as a ugly workaround on a misconfigured network with two identical hostnames. And the really bad thing is that it breaks as we figured out in a POC all modern concepts that uses containerization like Docker. All services that we are using at the moment are completely independent from the hostid function.

FlorianReimold:
I have to agree with Kristof here. The only reason why eCAL would have the HostID function at all was that it was used internally. Now that it isn't used anywhere any more it should be deprecated.

PR #610, opened 2022-04-12
Fixed monitor plugin template

hannemn:
**Pull request type**

 

 

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

Monitor plugin template wasn't part of the default build tree and was broken anyways.

**What is the new behavior?**

The monitor plugin template has been fixed and added to the default build tree. However, it is not included in the cmake install step.
-
-
-

- [ ] Yes
- [X] No

**Other information**


hannemn:
Well, the previous plugin template target was neither built nor installed with ecal. The subproject was commented out in the Cmakelists.txt file and it was broken anyways (wrong files in the include header). With the new version I fixed everthing and added the plugin template to Cmake, so that it gets built but not installed as part of the debian package and the windows installer.

FlorianReimold:
Do you think it makes sense to install it with eCAL, so people can use it to develop eCAL Mon Plugins?

Kerstin-Keller (approved):
I think it's good. Doesn't make a whole lot of sense to install it, unless we just place them in the samples folder.

PR #620, opened 2022-04-20
Visualisation dock widget

### LINKED ISSUE #542: Dockable eCAL Mon Plugins

hannemn:
Fixes #542

**Pull request type**

 

 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

Issue Number: N/A

**What is the new behavior?**

-
-
-

- [ ] Yes
- [X] No

**Other information**


FlorianReimold (approved):

hannemn:
@FlorianReimold Thanks for the approval, but I will probably need one or two further commits ;-)

PR #630, opened 2022-04-22
Removed CMakeFunctions

### LINKED ISSUE #553: cmake_functions: not able to use qt_add_windeployqt_postbuild

FlorianReimold:
Moved almost everything from CMakeFunctions to the private cmake helpers, so the files don't get installed and depended on any more. They are only used for building eCAL itself anyways.

What is left of it are only the protobuf helper functions. I renamed it to "ProtocFunctions" for that reason.

Fixes #553


Kerstin-Keller (commented):
Hi @FlorianReimold ,

in general I agree with your changes. However, I would go even further then. (So there may be no side effects if someone uses eCAL together with the original CMakeFunctions repo)

- No `-DECAL_THIRDPARTY_BUILD_PROTOCFUNCTIONS` options, the respective files are always included with a CMake build.
- No `find_package(CMakeFunctions)` etc calls
- Rename functions to be prefixed with `ecal_` ... (`ecal_protoc_generate_xxx`)

Do you agree? If so, I can make the required changes when I get back to office next week so they can still be part of 5.10.

FlorianReimold:
Yes, I totally agree, but I was scared to do that on my own with the limited time I have left before the next release (need time for testing 😉)

rex-schilasky (approved):

PR #646, opened 2022-05-03
CMake: Setting BUILD_QWT Option based on whether HAS_QT5 is set

FlorianReimold:


FlorianReimold:
Yea, that happens when you test locally and then think "I didn't change that much, I will just re-type it in the GitHub Editor" 😄

But I am doing it again (retyping in GH Editor), so let's see how Attempt 2 goes 😉

Issue #671, opened 2022-05-17
eCAL 5.10 broke compatibility with eCAL 5.9 and lower in C API


FlorianReimold:
Some functions were renamed without any need to rename them.

Commit:
https://github.com/continental/ecal/commit/17f576bc749461be597de5afb1993102593fe102#diff-d689389f1a48931b2e6f5dce34cff0ff5b8ac0a8efe2a99879faca9b01742a67

Files (probably more):
ecal/core/include/ecal/cimpl/ecal_proto_dyn_json_subscriber_cimpl.h
ecal/core/include/ecal/cimpl/ecal_client_cimpl.h
ecal/core/include/ecal/cimpl/ecal_publisher_cimpl.h
ecal/core/include/ecal/cimpl/ecal_subscriber_cimpl.h

rex-schilasky:
Yes, this was done by accident. The idea was to have the same naming convention for all C API functions. Some of them end with a 'C' some not. This C postfix makes no sense but we can rename then back for the minority of users that are using the raw C API.

Kerstin-Keller:
I will make the fix. We will have both and deprecate the old `C` postfix. We can then remove the deprecated functions with eCAL 6.

The deprecation may be compiler specific, and might not work on all platforms.
However it will be marked as deprecated in the documentation.

Should we wait with the deprecation for 5.11. and have both (none deprecated) in 5.10? Because then we can mention it in the release notes.

FlorianReimold:
The deprecation notice is not really consistent anyways. But please add some "Deperecated since 5.11" or whatever to the documentation, so we can add more consistent deprecation notes to the release notes in the future

PR #674, opened 2022-05-18
Reintroduce functions that were removed with 5.10 (accidentially)

Kerstin-Keller:
**Pull request type**

- [x] Other (please describe): 

Revert function renaming introduce by 17f576b

Issue Number: #671 

**What is the new behavior?**
Bring back C-Api  function that were renamed in 5.10, but mark as deprecated

- [ ] Yes
- [x] No

### DIFF
--- a/ecal/core/include/ecal/cimpl/ecal_client_cimpl.h
+++ b/ecal/core/include/ecal/cimpl/ecal_client_cimpl.h
@@ -108,6 +108,7 @@ extern "C"
 
  /**
    * @brief Add server response callback. 
+   * @since eCAL 5.10.0
    *
    * @param handle_    Client handle. 
    * @param callback_  Callback function for server response.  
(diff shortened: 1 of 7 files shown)


FlorianReimold:
clang-tidy complained about `ECALC_API`, `ECAL_HANDLE` etc. not being known. Guess what, it is right 😄
Currently this only compiles because `ecalc.h` includes the headers 
We should include add `#include <ecal/ecal_defs.h>` to all of those files and also propagate the ´ECAL_HANDLE´ properly, whcih is currently hardcoded in `ecalc.h`:
https://github.com/continental/ecal/blob/cd65b54e5022f7681232840a083fb9c2b2c3a2b7/ecal/core/include/ecal/ecalc.h#L40

This can however be done in a separate PR, if we don't want those changes on the 5.10 support branch.

Kerstin-Keller:
Ah ok, yeah I'd prefer to solve this in a separate PR, where I would also remove all the `ECALC_API` definitions from the function definitions.
And this one can be easily cherry picked to the 5.10 support branch (just without deprecating the functions).

Kerstin-Keller:
Also I agree that something is not quite right with the includes. My build problems went away after a clean build. Which is never a good sign.

FlorianReimold (approved):
@Kerstin-Keller will fix the clang-tidy issues in a separate pr 😊

PR #684, opened 2022-05-30
Win Setup: Added support for non-admin installations

FlorianReimold:
Admin: install to `C:\Program Files\eCAL`
Normal User: install to `C:\Users\USERNAME\AppData\Local\Programs\eCAL\`

Environment variables (`PATH`) are set for all users or for the current user only if in non-admin mode.

ecal.ini will be in `C:\Program Data\eCAL\ecal.ini` in both situations.


FlorianReimold:
I made this a draft, as I would like to get some feedback regarding the install path, first. In "normal" installation mode (as admin), the installation path is changed from `C:\eCAL` to `C:\Program Files\eCAL`. The latter one seems more reasonable, but I would like to know if the former (old) path was chosen for a reason.

The line is:
https://github.com/continental/ecal/blob/c70f7891fef177e4211ccdd598f3f92d7250becc/cpack/innosetup/ecal_setup.iss.in#L40

rex-schilasky:
> Actually I do not know. I think `C:\Program Files\eCAL` is totally fine 😄
> 
> But in general any install location would (should) work.

rex-schilasky:
> I made this a draft, as I would like to get some feedback regarding the install path, first. In "normal" installation mode (as admin), the installation path is changed from `C:\eCAL` to `C:\Program Files\eCAL`. The latter one seems more reasonable, but I would like to know if the former (old) path was chosen for a reason.
> 
> The line is:
> 
> https://github.com/continental/ecal/blob/c70f7891fef177e4211ccdd598f3f92d7250becc/cpack/innosetup/ecal_setup.iss.in#L40

The old path `C:\eCAL` was chosen to allow to edit the configuration file `ecal.ini` that was this time in the `ini` subfolder of the installation. But I would prefer to use `C:\eCAL` for all variants of installation. It's easier in case of documentation or any kind of support.

(Sorry for closing and reopening, just back from vacation ;-)).

Issue #687, opened 2022-06-07
compile ecal::pb as .so/dll file on Linux/Windows


FlorianReimold:
**Describe the bug**
Currently eCAL::pb is compiled as static `libecal_pb.a` file on Linux. This leads to issues with the protobuf description database.

The goal of this issue is having eCAL::pb compile as a shared object (`libecal_pb.so`) by default on Linux.

Kerstin-Keller:
We identified https://github.com/continental/ecal/blob/c590671491adefd72a1ef5450fef6be4ce08e7cf/cmake/helper_functions/ecal_add_functions.cmake#L168
as the problem / missunderstanding on my side.

As on Linux, shared libraries are the default on `add_library` commands, my assumption was that `BUILD_SHARED_LIBS=ON` on Linux, which apparently is not the case. 

We now have to options:
- add `option(BUILD_SHARED_LIBS)` in main CMakeLists.txt and provide sensible defaults for both win / linux to reflect the default CMake behavior
- remove the two different functions (static, no static) to create ecal libraries. However then can't force windows shared lib on lib by lib basis (which should still be the default for ecal core).

StaubSaugerRoboter:
Enabling `ecaltime-simtime` and starting the eCAL Player chrashes on Arch Linux.

```
[libprotobuf ERROR /build/protobuf/src/protobuf-21.9/src/google/protobuf/descriptor_database.cc:642] File already exists in database: ecal/pb/sim_time.proto
[libprotobuf FATAL /build/protobuf/src/protobuf-21.9/src/google/protobuf/descriptor.cc:1986] CHECK failed: GeneratedDatabase()->Add(encoded_file_descriptor, size):
terminate called after throwing an instance of 'google::protobuf::FatalException'
  what():  CHECK failed: GeneratedDatabase()->Add(encoded_file_descriptor, size):
[1]    71947 IOT instruction (core dumped)  ecal_play_gui
```

The reason might be, that the protobuf description library is loaded twice.

FlorianReimold:
Maybe a solution to this issue would also be to link `libprotobuf` statically. We already do that by default on Windows, where we build it from source. On Ubuntu (when using system packages) we currently don't specify how Protobuf should be linked against, so it defaults to dynamic linkage. However, it looks like `libprotobuf-dev` copntains a static version as well:

```bash
florian@florian-ubuntu22:/usr/lib/x86_64-linux-gnu$ ll | grep libprotobuf.a
[...]

PR #688, opened 2022-06-09
Add `BUILD_SHARED_LIBS` option (which is already CMake variable).

Kerstin-Keller:
**Pull request type**
I'd suggest to solve #687 like this until we do a complete CMake cleanup.
It is in line with the expected build behavior, which (by default) builds shared object on unix and static libs on Windows.


FlorianReimold (approved):
I think the solution is absolutely acceptable and will solve the problem.

Kerstin-Keller:
it seems it didn't help. they are still `.a` on ubuntu 
image

PR #691, opened 2022-06-10
Play: increase file descriptor limit for ecal player on linux

FlorianReimold:


hannemn:
Handled in another issue.

Issue #701, opened 2022-07-01
eCAL Crashes when Subscriber is destructed while in callback


Kerstin-Keller:
When an eCAL subscriber is being destructed while its callback is called, the application crashes.

Can be reproduced with person_snd / person_rec in the following branch https://github.com/continental/ecal/tree/feature/ecal-bug-destructor

Branch tries already to change the order of destruction but does not seem to help.

Wonder if the shared mutexes changed behavior since 5.9, see https://github.com/continental/ecal/commit/9a6e1642ff80b933abf968317a2f554f057d9813

PR #703, opened 2022-07-07
Bugfix/Datarace in CMsgSubscriber Destruction

rex-schilasky:
simplified logic in message subscriber to avoid data races in destruction phase (proven with applied test from Kerstin)

**Pull request type**

 

 

- [X] Bugfix

Message subscriber crash while destruction based on data race when message callback is under heavy use.

Issue Number: #701

**What is the new behavior?**
No more crashing by simplified callback deactivation logic based on one class member only instead checking additional "active" boolean.

### DIFF
--- a/ecal/core/include/ecal/msg/subscriber.h
+++ b/ecal/core/include/ecal/msg/subscriber.h
@@ -79,16 +79,9 @@ namespace eCAL
     **/
     CMsgSubscriber(CMsgSubscriber&& rhs)
       : CSubscriber(std::move(rhs))
+      , m_cb_callback(std::move(rhs.m_cb_callback))
     {
-      bool has_callback(false);
-
-      {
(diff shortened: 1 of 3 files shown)


Kerstin-Keller (commented):
I wonder if this m_cb_callback pointer should either be atomic (i guess not possible for a pointer) or locked,  because to me it seems that RemReceiveCallback and the actual callback may be called from different threads? or am I missunderstanding something.

However I will try the modified version!

Kerstin-Keller:
@rex-schilasky please take a look. It's the threadsafe way to protect all accesses to the member callback function with a `std::mutex`. However, I kept the lock to only accessing the pointer, e.g. to prevent deadlock situations.

Also I test was added that causes a crash in the current master implementation.
If we decide to merge, it should be sherry-picked to release branches.

PR #716, opened 2022-07-21
Restart observer if observing was interrupted.

### LINKED ISSUE #714: eCAL SHM communication broken for Process after hibernate
**Describe the bug**
After returning from hibernate, eCAL Subscribers are not working anymore.

**Analysis**
After a hibernate, the `CMemFileObserver::Observe` times out, the state is set to not obeserving.
https://github.com/continental/ecal/blob/0e714c6342d97f8a1c8c91bf299dbf7208e46176/ecal/core/src/io/ecal_memfile_pool.cpp#L274

This condition can never be rehabilitated.
The observer object for the memfile is global, and ist only deleted upon closing the proces..
Thus once the communication has timed out, it will never start again.

Kerstin-Keller:
Fixes #714

- [x] Bugfix

### DIFF
--- a/ecal/core/src/io/ecal_memfile_pool.cpp
+++ b/ecal/core/src/io/ecal_memfile_pool.cpp
@@ -347,7 +347,17 @@ namespace eCAL
     auto observer_it = m_observer_pool.find(memfile_name_);
     if(observer_it != m_observer_pool.end())
     {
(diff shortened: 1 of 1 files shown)


rex-schilasky (approved):
Pragmatic solution. LGTM

PR #737, opened 2022-08-01
Allow binary reading of the measurements in Python API.

KerstinKeller:
**Pull request type**

Allow the possibility to read a measurement in binary mode (and access info like descriptor, schema type, ...)

- [x] Feature

No access to binary channel data from high level python interface

**What is the new behavior?**
Two measurement modes for accessing data -> OBJECT (default) and BINARY
The user may decide how they want to access the data.

- [x] No

The current API is extended to accept a new value for measurement access mode.
Per default the access mode is `OBJECT`, which is the same behavior as before the change.

### DIFF
--- a/lang/python/ecalhdf5/ecal/measurement/measurement.py
+++ b/lang/python/ecalhdf5/ecal/measurement/measurement.py
@@ -49,7 +49,7 @@ class BinaryChannel(object):
 
     channel_type = measurement._reader.get_channel_type(channel_name)
     try:
-      self._type_encoding, self._type_name = channel_type.split(":")
+      self._type_encoding, self._type_name = channel_type.split(":", 1)
     except:
       self._type_name = channel_type
       self._type_encoding = ""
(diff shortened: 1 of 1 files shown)


KerstinKeller:
The properties are now defined the following way:
```
class BinaryChannelInfo(NamedTuple):
    topic_name: str
    type_encoding: str
    type_name: str
    type_descriptor: bytes
```
Respective as properties of a channel.

lang/python/ecalhdf5/ecal/measurement/measurement.py
```diff
@@ -26,12 +26,100 @@
...
+class BinaryChannel(object):
+
+  def __init__(self, measurement, channel_name):
+    self._measurement = measurement
+    self._channel_name = channel_name
+    self._entries = self._measurement._reader.get_entries_info(channel_name)
+
+    channel_type = measurement._reader.get_channel_type(channel_name)
+    try:
+      self._type_encoding, self._type_name = channel_type.split(":")
```
FlorianReimold, line 52: This will fail for channel types with more than 1 `:`. Maybe it's a good idea to only split at the first `:`? What do you think?

KerstinKeller: Ok, splitting now at first occurence.

FlorianReimold (approved):

Issue #611, opened 2022-04-12
Compile eCAL core standalone


Kerstin-Keller:
This issue describes the steps that need to be taken to be able to compile the eCAL core component as a standalone project.

These steps will probably be evolved as separate issues, as handling everything in one PR will be too big.

- [x] Separate .proto definitions: Currently all .proto definitions for eCAL are contained in ecal/pb and linked into one library. This is problematic if we want to build the core as standalone. We should split these files and compile them alongside the approriate project (core, player, recorder, sys, ...)
- [x]  Other libs: `eCAL::utils` -> need to see what parts are needed, maybe extract those relevant for eCAL Core
- [ ]  `eCAL::proto`: This library contains helper functionality, used mainly for proto publishers / subscribers / reflections.  First, we move all these files out of the core. Then we can actually create language packs, which build upon the pure `CSubscriber`, .. Layer. These serialization packs can include different components: Pub/sub, Service/Client, Reflection, Monitoring, samples, ...
- [ ] Make eCAL Core installable by itself. E.g. Create `eCALCoreConfig.cmake` -> `eCALConfig.cmake` includes, (depending on requested component), the `eCALCoreConfig.cmake`. (Qt5 does it very similarly). Modelling interdependencies may be a bit tough. `find_package(eCAL REQUIRED COMPONENT core)` will be equivalent to `find_package(eCALCore REQUIRED)`.

rex-schilasky:
Current state:

- protobuf files splitted into app, time, ecal
- npcap supporting library (udpcap) will be part of the new ecal-core the same way as today (maybe we make a standalone library later)
- CustomTclap dependency cutted (2 file duplicates now part of the core)
- ecal_utils needs to be extracted as standalone library, proposed name "ecal-util"

rex-schilasky:
ecal-utils moved in separate repository https://github.com/eclipse-ecal/ecal-utils

PR #741, opened 2022-08-04
Feature/split protobuf interfaces

rex-schilasky:
**Pull request type**

 

 

- [X] Refactoring (no functional changes, no api changes)

Protobuf interfaces handled in common subproject for all eCAL projects

Issue Number: #611 (part1)

**What is the new behavior?**
Protobuf interfaces split into ecal::pb (core), app::pb (applications), ecaltime::pb (timeplugin)


KerstinKeller (commented):
Sorry to be picky here, but I have  a few remarks to this PR.

## Include paths

Especially for monitoring, but also sys, they are part of the public API headers. (more or less, the user can also run protoc on the proto files himself.).
So, in all cases, the include of any of these **should** start with `ecal`, no matter what.

In my opinion we could do
```
#include <ecal/core/pb/monitoring.pb.h>
```
or 
```
#include <ecal/pb/core/monitoring.pb.h>
 ```
 Though I think I like the first one better. (and if we are moving the core, we could consider moving all relevant header files also into `<eCAL/core/...> (however, keeping the old ones and deprecating them for compatibility).
 
 
 ## Targets
 From my point of view, there is no reason, that the pb files have a separate target, like `core_pb`. They could just be part of `eCAL.dll / eCAL.so`.
 However, I am not 100% sure of the implications. Probably (on Windows) they are not part of `dllexport`, hence the symbols are most probably not exported from the dll.
 
 This means that the user would have to run protoc on all those proto files which he wants to use. And then I see the risk (in case of monitoring) that we run into this error where the same proto file cannot be added to the protobuf database twice.

So maybe for now we do keep the different targets (which somehow does brake "API", because old build scripts that link `eCAL::pb` won't work any more, but we can work around that, by keeping `eCAL::pb` as an interface target and deprecating it).

But for the future, at least for the monitoring, we might consider offering an API where we pack the Protobuf data in structs, so we are not depending on Protobuf as part of the eCAL API.

KerstinKeller (commented):
Also a good way of seeing if you are breaking API, is to (at first) not making any changes to the samples, and then see what is necessary to get them to work again (with deprecation warnings). (Introducing dummy header files (deprecated) and dummy targets).

Then you can go ahead and fix the warnings. You can then be sure you didn't break anything for the users.

rex-schilasky:
> Sorry to be picky here, but I have a few remarks to this PR.
> 
> ## Include paths
> Especially for monitoring, but also sys, they are part of the public API headers. (more or less, the user can also run protoc on the proto files himself.). So, in all cases, the include of any of these **should** start with `ecal`, no matter what.
> 
> In my opinion we could do
> 
> ```
> #include <ecal/core/pb/monitoring.pb.h>
> ```
> 
> or
> 
> ```
> #include <ecal/pb/core/monitoring.pb.h>
> ```
> 
> Though I think I like the first one better. (and if we are moving the core, we could consider moving all relevant header files also into `<eCAL/core/...> (however, keeping the old ones and deprecating them for compatibility).
> 

I moved/renamed

- ```core/pb/.../.pb.h``` to ```ecal/core/pb/.../.pb.h```
- ```app/pb/.../.pb.h``` to ```ecal/app/pb/.../.pb.h```
- ```ecaltime/pb/.../.pb.h``` to ```ecal/ecaltime/pb/.../.pb.h```

rex-schilasky:
> Also a good way of seeing if you are breaking API, is to (at first) not making any changes to the samples, and then see what is necessary to get them to work again (with deprecation warnings). (Introducing dummy header files (deprecated) and dummy targets).
> 
> Then you can go ahead and fix the warnings. You can then be sure you didn't break anything for the users.

I would not go for full downward compatibility here. For now just include paths needs to be fixed like 
- ```<ecal/pb/monitoring.pb.h>``` to ```<ecal/core/pb/monitoring.pb.h>``` or 
- ```<ecal/pb/play/service.pb.h>``` to ```<ecal/app/pb/play/service.pb.h>```.

Additional target names needs to be changed from 
- ```eCAL::pb``` to ```eCAL::core_pb``` and
-  ```eCAL::pb``` to ```eCAL::app_pb```.

In a separate issue we need to remove the precompiled protobuf header at all and provide the protocol buffer definition files instead.

rex-schilasky:
Old target eCAL::pb added for full downwards compatibility. Old protobuf header files installed with redirections to the new file structure and deprecate warnings.

PR #749, opened 2022-08-08
FIX: Multiply Defined Symbols

patrick-fleischmann-conti:
Fix multiply defined symbols linker error that occurs when IMeasurement or OMeasurement is used in two different translation units as the present method definition in the two header files violates the One Definition Rule.

**Pull request type**

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

As method implementations for IMeasurement and OMeasurement are contained in the header file, multiply defined symbols errors will be thrown by the linker if the classes are used in two different translation units.

Issue Number: N/A

**What is the new behavior?**
Using the inline keyword, this error is fixed.

- [ ] Yes
- [x] No

**Other information**

Alternative solutions would be implementing the method in the header definition directly or move them to cpp files instead.

### DIFF
--- a/contrib/ecalhdf5/include/ecal/measurement/imeasurement.h
+++ b/contrib/ecalhdf5/include/ecal/measurement/imeasurement.h
@@ -228,12 +228,12 @@ namespace eCAL
       std::shared_ptr<eh5::HDF5Meas> meas;
     };
 
-    IMeasurement::IMeasurement(const std::string& path)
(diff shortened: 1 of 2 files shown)


FlorianReimold:
@KerstinKeller: As I see it, this should not harm any API (but probably the ABI?), so it should be safe to merge to master. Can you confirm this?

KerstinKeller (approved):
This is fine, since definition is only in header files, there is no ABI to be concerned about 😉

PR #752, opened 2022-08-09
Steady clock thread sleep implementation for windows

emilbic28:
**Pull request type**

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

For windows, std::this_thread::sleep_for does not use a steady clock and is affected by system clock changes.

Issue Number: N/A

**What is the new behavior?**

- Added a new templated function which can be used similarly to the classic sleep_for, which takes a std::chrono::duration argument
- Using Sleep method from synchapi.h header, thread sleep works as supposed and is not affected by system clock changes

- [ ] Yes
- [x] No

**Other information**

### DIFF
--- a/ecal/core/src/ecal_process.cpp
+++ b/ecal/core/src/ecal_process.cpp
@@ -366,7 +366,25 @@ namespace eCAL
 
     void SleepMS(const long time_ms_)
     {
-      std::this_thread::sleep_for(std::chrono::milliseconds(time_ms_));
+      #ifdef ECAL_OS_WINDOWS
+      {
+        Sleep(time_ms_);
+      }
+      #else
+        std::this_thread::sleep_for(std::chrono::milliseconds(time_ms_));
+      #endif
+    }
+
+    void SleepNS(const long long time_ns_)
+    {
+      #ifdef ECAL_OS_WINDOWS
+      {
+        auto milliseconds = time_ns_ / 1000000 + ((time_ns_ / 1000000) != 0);
+        Sleep(milliseconds);
+      }
+      #else
+        std::this_thread::sleep_for(std::chrono::nanoseconds(time_ns_));
+      #endif
     }
 
     float GetProcessCpuUsage()
--- a/ecal/core/src/ecal_process.cpp
+++ b/ecal/core/src/ecal_process.cpp
@@ -379,7 +379,7 @@ namespace eCAL
     {
       #ifdef ECAL_OS_WINDOWS
       {
-        auto milliseconds = time_ns_ / 1000000 + ((time_ns_ / 1000000) != 0);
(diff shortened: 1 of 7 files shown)


ecal/core/src/ecal_process.cpp
```diff
@@ -366,7 +366,25 @@ namespace eCAL
...
+      #else
+        std::this_thread::sleep_for(std::chrono::milliseconds(time_ms_));
+      #endif
+    }
+
+    void SleepNS(const long long time_ns_)
+    {
+      #ifdef ECAL_OS_WINDOWS
+      {
+        auto milliseconds = time_ns_ / 1000000 + ((time_ns_ / 1000000) != 0);
```
FlorianReimold, line 382: I assume the idea is to round up by checking if the division has a rest?

```suggestion
        auto milliseconds = time_ns_ / 1000000 + ((time_ns_ % 1000000) != 0);
```

FlorianReimold (changes requested):

ecal/core/src/ecal_process.cpp
KerstinKeller: I am not 100% sure I agree with this function. We shouldn't offer a `SleepNS` function, if we cannot sleep to that precision.
And in the above case I think a `auto milliseconds = time_ns / 1000000` is fine, since it will automatically round down? or do you want to potentially round up? then I would look at `std` functions like `floor`, `ceil`, ....

FlorianReimold: > I am not 100% sure I agree with this function. We shouldn't offer a `SleepNS` function, if we cannot sleep to that precision. 

It is common practice to provide those of functions in order to have a common API. They also exist in STL, even though many operating systems fail to reach that kind of precision. In this case, Win32 will not reach nanosecond precision, but Linux systems maybe will.

> And in the above case I think a `auto milliseconds = time_ns / 1000000` is fine, since it will automatically round down? or do you want to potentially round up?

Sleep functions are usually defined as sleeping "at least" for the given duration. Therefore, you need to round up, not down.

> then I would look at `std` functions like `floor`, `ceil`, ....

For those functions you need to convert to a floating point number, first.  I don't think that would be a good idea when you can just add a 1 as integer.

emilbic28: @FlorianReimold you are right, it was a typo

KerstinKeller (commented):
ok, so if we're concerned about speed, we can also do:
```
auto milliseconds = (time_ns_ + 999999) / 1000000;
```
Which is just one addition, and one division, compared to one division, one remainder, one comparison and one addition 😄 
Not that it matters in a function like this 😉

PR #755, opened 2022-08-10
Support one file per topic feature 

bkhalifaconti:
- [x] Feature

Currently all topics are written to one HDF5 file. 

Issue Number: 543

**What is the new behavior?**

- User can choose to seperate different topics in different files. A file per each topic

- [x] No


testing/contrib/ecalhdf5/hdf5_test/src/hdf5_test.cpp
```diff
@@ -90,7 +90,7 @@ TEST(HDF5, WriteReadIntegrity)
   const long long   t2_id            = 2LL;
   const long long   t2_clock         = 12LL;
 
-  const std::string t3_name          = " ASCII and beyond!\a\b\t\n\v\f\r\"#$%&\'()*+,-./0123456789:;<=>\?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~������������������";
+  const std::string t3_name          = " ASCII and beyond!\a\b\t\n\v\f\r\"#$%&\'()*+,-./0123456789:;<=>\?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~������������������";
```
FlorianReimold, line 93: Looks like the encoding of this file broke

FlorianReimold (changes requested):

testing/contrib/ecalhdf5/hdf5_test/src/hdf5_test.cpp
bkhalifaconti: Not sure what happened to miss up with encoding, I checked encoding in my editor and it is UTF-8, I think that is correct? 
anyway I retrieved back the umlaut characters and pushed the code

FlorianReimold:
@bkhalifaconti: The following checkbox fails to update, when I load a config:
image

Even though this is probably a bug I introduced myself, can you fix it? I think you will have to connect to a signal emitted by `QEcalRec` and update the checkbox accoringly.

Also it looks like there is a merge conflict, as the proto files haven been moved to: `app\app_pb\src\ecal\app\pb\rec`. Please merge the master in your branch and try to fix the merge conflict

bkhalifaconti:
I merged master into my branch and pushed it, but didn't see any merge conflicts, anyway commit is pushed so you can have a look.

FlorianReimold:
> @FlorianReimold I merged master into my branch and pushed it, but didn't see any merge conflicts, anyway commit is pushed so you can have a look.

you probably merged the master of your fork into that branch. You will have to sync the master of your fork with the eclipse-ecal master before doing that. There should be a button in the web interface of your forked master to sync it to the upstream.

FlorianReimold:
I noticed another small bug (also introduced by me 😳)

the ecal_recorder_client_service_gui (from the service samples) doesn't activate / deactivate the input for 1-file-per-topic properly. Also the Tab order (when pressing the TAB key on your keyboard) is not correct. And the function is totally un-implemented 😀

image

Activating and de-activating the fields is done from the qt designer_
image

Tab order is also done from the qt designer:
image

bkhalifaconti:
> app\app_pb\src\ecal\app\pb\rec

@FlorianReimold  I synced master with my fork and merged still merge took place smoothly, anyway again the new merge is pushed

bkhalifaconti:
> @bkhalifaconti: The following checkbox fails to update, when I load a config: image
> 
> Even though this is probably a bug I introduced myself, can you fix it? I think you will have to connect to a signal emitted by `QEcalRec` and update the checkbox accoringly.
> 
> Also it looks like there is a merge conflict, as the proto files haven been moved to: `app\app_pb\src\ecal\app\pb\rec`. Please merge the master in your branch and try to fix the merge conflict

Checkbox issue fixed. Not only signal needed to be connected but also the signal wasn't emitted properly and the loaded config wasn't applied correctly.

FlorianReimold:
@bkhalifaconti: So this PR is ready for review? 👍

bkhalifaconti:
> @bkhalifaconti: So this PR is ready for review? 👍

Not yet, there are still the last to items you mentioned. I am looking on them now

PR #756, opened 2022-08-10
Fixed internal CMemoryFile structure compatibility

hannemn:
**Pull request type**

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: N/A

**What is the new behavior?**

-
-
-

- [ ] Yes
- [x] No

**Other information**

### DIFF
--- a/ecal/core/src/io/ecal_memfile.cpp
+++ b/ecal/core/src/io/ecal_memfile.cpp
@@ -29,8 +29,9 @@
 #include <cassert>
 #include <cstdint>
 #include <cstring>
+#include <algorithm>
 
-#define SIZEOF_PARTIAL_STRUCT(_STRUCT_NAME_, _FIELD_NAME_) reinterpret_cast<std::size_t>(&(reinterpret_cast<_STRUCT_NAME_*>(0)->_FIELD_NAME_)) + sizeof(_STRUCT_NAME_::_FIELD_NAME_)
+#define SIZEOF_PARTIAL_STRUCT(_STRUCT_NAME_, _FIELD_NAME_) (reinterpret_cast<std::size_t>(&(reinterpret_cast<_STRUCT_NAME_*>(0)->_FIELD_NAME_)) + sizeof(_STRUCT_NAME_::_FIELD_NAME_))
 
 namespace eCAL
 {
@@ -123,8 +124,7 @@ namespace eCAL
         memfile::db::CheckFileSize(name_, header_size, m_memfile_info);
 
         // copy compatible header part into m_header
-        const auto compatible_header_size = sizeof(SInternalHeader) > header_size ? header_size : sizeof(SInternalHeader);
-        memcpy(&m_header, m_memfile_info.mem_address, compatible_header_size);
+        memcpy(&m_header, m_memfile_info.mem_address, std::min(sizeof(SInternalHeader), static_cast<std::size_t>(header_size)));
 
         // unlock mutex
         UnlockMtx(&m_memfile_info.mutex);
@@ -309,8 +309,7 @@ namespace eCAL
     }
 
     // update compatible header part of m_header
-    const auto compatible_header_size = sizeof(SInternalHeader) > m_header.int_hdr_size ? m_header.int_hdr_size : sizeof(SInternalHeader);
-    memcpy(&m_header, m_memfile_info.mem_address, compatible_header_size);
+    memcpy(&m_header, m_memfile_info.mem_address, std::min(sizeof(SInternalHeader), static_cast<std::size_t>(m_header.int_hdr_size)));
 
     // check size again
     size_t len = static_cast<size_t>(m_header.int_hdr_size) + static_cast<size_t>(m_header.max_data_size);
--- a/ecal/core/src/io/ecal_memfile.cpp
+++ b/ecal/core/src/io/ecal_memfile.cpp
@@ -27,7 +27,10 @@
 #include "ecal_memfile_db.h"
 
 #include <cassert>
+#include <cstdint>
 #include <cstring>
+
+#define SIZEOF_PARTIAL_STRUCT(_STRUCT_NAME_, _FIELD_NAME_) reinterpret_cast<std::size_t>(&(reinterpret_cast<_STRUCT_NAME_*>(0)->_FIELD_NAME_)) + sizeof(_STRUCT_NAME_::_FIELD_NAME_)
 
 namespace eCAL
 {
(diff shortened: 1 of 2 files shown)


ecal/core/src/io/ecal_memfile.cpp
```diff
@@ -73,7 +73,7 @@ namespace eCAL
       m_memfile_info = SMemFileInfo();
 
       // create memory file
-      if (!memfile::db::AddFile(name_, create_, len_ + sizeof(SInternalHeader), m_memfile_info))
+      if (!memfile::db::AddFile(name_, create_, len_ + m_header.int_hdr_size, m_memfile_info))
```
hannemn, line 76: I think the compatibility problem can still occur if create_ = false. Since the memory file is always mapped with len_ + sizeof(SInternalHeader), which may be the wrong length if the file already exists. (In the worst case the end of the memory file gets just truncated).

FlorianReimold: Obviously, eCAL must check the content of a memory file, if it exists. For that you have to:

- read the header size
- zero-memory a header struct
- copy min(sizeof(header), header_size_from_memfile) bytes into the created header struct

FlorianReimold (changes requested):

ecal/core/src/io/ecal_memfile.h
```diff
@@ -161,12 +162,29 @@ namespace eCAL
...
-      unsigned long  cur_data_size = 0;
-      unsigned long  max_data_size = 0;
+      std::uint16_t int_hdr_size = sizeof(SInternalHeader);
+#if _WIN32 || _WIN64 || (INTPTR_MAX == INT32_MAX) // Use of standard 32 bit data types on windows and other 32 bit platforms to remain compatible with built-in "long" data type
+                                                  // of previous struct layout. For some reason 64-bit msvc uses 4 bytes alignment as well.
+      std::uint8_t   __reserved_0[2];             // Add 2 bytes padding for 4 bytes alignment on Windows  
+      std::uint32_t  cur_data_size = 0;
+      std::uint32_t  max_data_size = 0;
+#else                                             // Use of standard 64 bit data types on all 64 bit platforms to remain compatible with built-in "long" data type of previous struct layout
+      std::uint8_t  __reserved_0[6];              // Add 6 bytes padding for 8 bytes alignment on 64-bit Linux
```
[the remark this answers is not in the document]
FlorianReimold: Apparently this is a valid warning.

> The C standard additionally reserves names beginning with a double underscore, while the C++ standard strengthens this to reserve names with a double underscore occurring anywhere.
>
> Violating the naming rules above results in undefined behavior.

https://releases.llvm.org/13.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/bugprone-reserved-identifier.html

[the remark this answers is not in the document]
hannemn: To be honest I don't know if this is really a good idea to use an std::array instead of c-style array in that case. Does the c++ standard guarantee that a c++ array like std::array<std::uint8_t, x> foo occupies x bytes in the memory layout like std::uint8_t foo[x]?

ecal/core/src/io/ecal_memfile.h
```diff
@@ -161,12 +163,31 @@ namespace eCAL
...
     {
-      unsigned short int_hdr_size  = sizeof(SInternalHeader);
-      unsigned long  cur_data_size = 0;
-      unsigned long  max_data_size = 0;
+      static_assert(sizeof(std::array<std::uint8_t, 1>) == 1, "Memory layout of std::array is different from C-style array.");
+      
+      std::uint16_t int_hdr_size = sizeof(SInternalHeader);
+#if _WIN32 || _WIN64 || (INTPTR_MAX == INT32_MAX)    // Use of standard 32 bit data types on windows and other 32 bit platforms to remain compatible with built-in "long" data type
+                                                     // of previous struct layout. For some reason 64-bit msvc uses 4 bytes alignment as well.
+      std::array<std::uint8_t, 2> _reserved_0;       // Add 2 bytes padding for 4 bytes alignment on Windows
```
FlorianReimold, line 178: Those should be initialized to 0 as well. While this is not relevant now, this may be relevant when we later use one of those fields. Then the legacy applications should always send a 0 there to not confuse newer applications that read an actual value from those fields.

```suggestion
      std::array<std::uint8_t, 2> _reserved_0 = {};       // Add 2 bytes padding for 4 bytes alignment on Windows  
```

ecal/core/src/io/ecal_memfile.h
```diff
@@ -161,12 +163,31 @@ namespace eCAL
...
+      static_assert(sizeof(std::array<std::uint8_t, 1>) == 1, "Memory layout of std::array is different from C-style array.");
+      
+      std::uint16_t int_hdr_size = sizeof(SInternalHeader);
+#if _WIN32 || _WIN64 || (INTPTR_MAX == INT32_MAX)    // Use of standard 32 bit data types on windows and other 32 bit platforms to remain compatible with built-in "long" data type
+                                                     // of previous struct layout. For some reason 64-bit msvc uses 4 bytes alignment as well.
+      std::array<std::uint8_t, 2> _reserved_0;       // Add 2 bytes padding for 4 bytes alignment on Windows  
+      std::uint32_t               cur_data_size = 0;
+      std::uint32_t               max_data_size = 0;
+#else                                                // Use of standard 64 bit data types on all 64 bit platforms to remain compatible with built-in "long" data type of previous struct layout
+      std::array<std::uint8_t, 6> _reserved_0;       // Add 6 bytes padding for 8 bytes alignment on 64-bit Linux
```
FlorianReimold, line 182:
```suggestion
      std::array<std::uint8_t, 6> _reserved_0 = {};       // Add 6 bytes padding for 8 bytes alignment on 64-bit Linux
```

ecal/core/src/io/ecal_memfile.h
```diff
@@ -161,12 +163,31 @@ namespace eCAL
...
+      std::uint64_t               cur_data_size = 0;
+      std::uint64_t               max_data_size = 0;
+#endif
+      // New fields should only declare well defined data types and be aligned to 8 bytes
+      // std::uint8_t                 _new_field = 0;
+      // std::array<std::uint8_t, 7>  _reserved_1;
```
FlorianReimold, line 188:
```suggestion
      // std::array<std::uint8_t, 7>  _reserved_1 = {};
```

FlorianReimold (changes requested):

ecal/core/src/io/ecal_memfile.h
hannemn: In case of _reserved_0 i don't think that it is necessary to zero the memory, since the field will never be used for future applications and is only intended to be compatible to the padding of the previous header layout (which maybe already contains non-zero values).

In the next version of the header I will introduce an additional version field to query which fields or rather features are available for the memory file. Then it can be easily extended for future use cases.

FlorianReimold: Hm, OK, but it still feels wrong (to me) to write uninitialized memory to a file. But should be OK, if you prefer it like that.

hannemn: I can change it if you feel better with zeros in the reserved field. The overhead of initialization should be acceptable. ;-)

FlorianReimold (approved):

Issue #723, opened 2022-07-27
Linux: Named SHM Files are never unlinked or closed


FlorianReimold:
# Issue
**Unlink issue**: Even when gracefully shutting down eCAL Applications, `/dev/shm` will never be cleaned up and contain many zombie memory files.
1. Memory files should be closed on gracefull shutdown
2. Memory files that are not supposed to be opened again, can be unlinked right after being opened (See https://stackoverflow.com/questions/65860154/behaviour-of-shm-unlink)

**Close Issue**: File Descriptors to SHM are never closed. According to https://man7.org/linux/man-pages/man3/shm_open.3.html the file descriptor may be directly closed after mmap has been called, without affecting the memory segment.

This issue may be related to #592. If File Descriptors would be closed properly, we wouldn't need to increase the number of open file handles at all.

# Solution:
- When gracefully shutting down eCAL, all memory files should be cleaned up
- after `mmap` the file descriptor should be closed in any case
- If possible, the shm file should be `unlinked` directly after it has been opened 

Optimal behavior:
- Even when killing an application through SIGINT, SIGTERM etc. the memory files should be cleaned up.

# To Reproduce
1. Open an eCAL App
3. Close the App
4. Check `/dev/shm`

**Desktop (please complete the following information):**
 - OS: Linux
 - Version: All

FlorianReimold:
According to Stackoverflow, we can `unlink` the files right after opening them, when we know that no other process will have to open it later:

> `shm_unlink()` removes the _name_ of the shared memory segment. The segment itself persists as long as any process has it open, but after it is unlinked, processes can no longer open it

https://stackoverflow.com/questions/65860154/behaviour-of-shm-unlink

FlorianReimold:
The following Code will get rid of zombie event files. I have added an `::shm_unlink` call right after the event has been opened. Events in eCAL have a 1:1 relation (at least if it is an event for triggering shared-memory communication), so it is possible to directly delete the name for it after the publisher has opened it.

https://github.com/eclipse-ecal/ecal/blob/1ac72079b97a98aa9601b6ff451046a27ccee7c9/ecal/core/src/ecal_event.cpp#L165

```cpp
  named_event_t* named_event_open(const char* event_name_)
  {
    // try to open existing shared memory file
    int fd = ::shm_open(event_name_, O_RDWR, S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH);
[...]

PR #765, opened 2022-08-17
Properly closing file descriptors on linux after mmap

FlorianReimold:
This fix is Linux-only. It saves 1 open file descriptor per published topic.
This relates to #723 but is not enough to solve #592


rex-schilasky (approved):
LGTM

PR #776, opened 2022-08-23
Add option to join multicast groups on all interfaces in linux

### LINKED ISSUE #695: Multicast groups are not updated when network interfaces change on linux
**Describe the bug**

Starting ecal nodes, on a machine with the main network interface down and just the loopback available, will not
start network communication when the main network interface goes up. Nodes need to manually restarted when network interfaces are up and running to get network communication going.

**To Reproduce**

***Setup***

2 Devices. Both Linux. Multicast routes for lo and main interface (enp2s0f0) are set on both devices.

```
239.0.0.0/24 dev enp2s0f0 proto static metric 1
239.0.0.0/24 dev lo proto static metric 1000
```

***Test***

Start ecal_sample_minimal_snd on one machine and ecal_sample_minimal_rec on the other with the ethernet cable unplugged.

On the machine running ecal_sample_minimal_rec. You will see the following in 'ip maddr show'. Multicast group is joined on
loopback device. Makes sense. enp2s0f0 is down.

```
ip maddr show | grep "lo\|enp2s0f0\|239.0.0"

1:      lo
        inet  239.0.0.7
        inet  239.0.0.1 users 2
2:      enp2s0f0
```

Plug cable in enp2s0f0.

```
ip maddr show | grep "lo\|enp2s0f0\|239.0.0"

1:      lo
        inet  239.0.0.7
        inet  239.0.0.1 users 2
2:      enp2s0f0
```

The subscription doesn't start. The group is still just joined on lo.

Same thing again with cable plugged in at the beginning. The group is joined on enp2s0f0
because it was up at the time and has a higher metric(?).

```
ip maddr show | grep "lo\|enp2s0f0\|239.0.0"

1:      lo
2:      enp2s0f0
        inet  239.0.0.7
        inet  239.0.0.1 users 2
```

Unplugging the cable will stop packets getting through and it resumes after plugging the cable
back in. Also ecal_mon_gui will complain about getting ecal monitoring information when it was started before the cable was unplugged.

**Expected behavior**
Network communication should start when the main network interface comes up and ecal_mon_gui shouldn't be reporting monitoring errors. Also nodes should not need to be restarted to recover communication.

**Desktop:**
 - OS: Ubuntu
 - Version 22.04

**Additional context**

I've created a workaround. So far for linux only. 

The idea was to add thread to CUDPReceiverAsio which keeps track of multicast groups and up running and running network interfaces. When groups are joined/left or network interfaces come up/go down the groups get updated.

I am guessing that 'm_socket.set_option(asio::ip::multicast::join_group(asio::ip::make_address(ipaddr_)), ec);' in  UDPReceiverAsio::AddMultiCastGroup just joins the group on one interface. I think it's the one up with the lowest metric in this case 'lo'. So when the network interfaces change the group has to be re-joined to make the communication recover.

The workaround can be found here:
https://github.com/tekniqueltd/ecal/commit/b5282a79075499992bf0747ca95511b4e91eb593

But maybe there is an easier way to deal with this?

floroeske:
**Pull request type**

- [x] Feature

Ecal nodes, which are started before the main network interface is up and running, can't start network communication when the main network interface becomes available. Ecal nodes need to be restarted after the main network interface is up and running to start network communication. This behaviour is caused by ecal not being able to subscribe to multicast groups on network interfaces which are down at node startup.

Fixes #695

**What is the new behavior?**
This pull request makes sure multicast groups are subscribed to on all network interfaces, independent of their current state. This makes sure ecal nodes can start network communication once network interfaces are up and running without restarting them.

- [ ] Yes
- [x] No

**Other information**

This feature is for Linux only and is currently optional. It can be enabled wiht the ECAL_JOIN_MULTICAST_ON_ALL_INTERFACES build flag.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,7 +88,6 @@ option(ECAL_INCLUDE_PY_SAMPLES                 "Include python language sample p
 option(ECAL_INSTALL_SAMPLE_SOURCES             "Install the sources of eCAL samples"                               ON)
 
 option(ECAL_JOIN_MULTICAST_TWICE               "Specific Multicast Network Bug Workaround"                        OFF)
-option(ECAL_JOIN_MULTICAST_ON_ALL_INTERFACES   "Enabling Joining Multicast Groups On All Interfaces"              OFF)
 option(ECAL_NPCAP_SUPPORT                      "Enable the eCAL Npcap Receiver (i.e. the Win10 performance fix)"  OFF)
 
 # Set option regarding third party library builds
@@ -561,7 +560,6 @@ message(STATUS "ECAL_LAYER_ICEORYX                             : ${ECAL_LAYER_IC
 message(STATUS "ECAL_INCLUDE_PY_SAMPLES                        : ${ECAL_INCLUDE_PY_SAMPLES}")
 message(STATUS "ECAL_INSTALL_SAMPLE_SOURCES                    : ${ECAL_INSTALL_SAMPLE_SOURCES}")
 message(STATUS "ECAL_JOIN_MULTICAST_TWICE                      : ${ECAL_JOIN_MULTICAST_TWICE}")
-message(STATUS "ECAL_JOIN_MULTICAST_ON_ALL_INTERFACES          : ${ECAL_JOIN_MULTICAST_ON_ALL_INTERFACES}")
 message(STATUS "ECAL_NPCAP_SUPPORT                             : ${ECAL_NPCAP_SUPPORT}")
 message(STATUS "ECAL_THIRDPARTY_BUILD_CMAKE_FUNCTIONS          : ${ECAL_THIRDPARTY_BUILD_CMAKE_FUNCTIONS}")
 message(STATUS "ECAL_THIRDPARTY_BUILD_SPDLOG                   : ${ECAL_THIRDPARTY_BUILD_SPDLOG}")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,6 +88,7 @@ option(ECAL_INCLUDE_PY_SAMPLES                 "Include python language sample p
 option(ECAL_INSTALL_SAMPLE_SOURCES             "Install the sources of eCAL samples"                               ON)
 
 option(ECAL_JOIN_MULTICAST_TWICE               "Specific Multicast Network Bug Workaround"                        OFF)
+option(ECAL_JOIN_MULTICAST_ON_ALL_INTERFACES   "Enabling Joining Multicast Groups On All Interfaces"              OFF)
 option(ECAL_NPCAP_SUPPORT                      "Enable the eCAL Npcap Receiver (i.e. the Win10 performance fix)"  OFF)
 
 # Set option regarding third party library builds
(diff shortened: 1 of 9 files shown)


FlorianReimold:
@floroeske: Thanks for the work you are putting in this! Tell me when it's done, so I we can review and merge it! 🙌

At the moment, unfortunately the Eclipse Contributor Agreement check still complains. It is a check by the Eclipse Foundation whether a contributor has signed their mandatory contributor agreement:

> The purpose of the Eclipse Contributor Agreement (ECA) is to provide a written record that contributors have agreed to provide their contributions of code and documentation under the licenses used by the Eclipse project(s) they’re contributing to.

Maybe you could create an Eclipse Account and sign it? It should then become green ✅! 

1. Create an Eclipse Account: https://accounts.eclipse.org/user/register?destination=user/edit
2. Go to your account settings, scroll down and add your GitHub Username in the "Social Media Links" Section:

    image

4. Go to http://www.eclipse.org/contribute/cla and complete the form to electronically sign the Contributor Agreement

FlorianReimold:
@floroeske: Could you sign the Eclipse Contributors Agreement (see above)?

floroeske:
Sorry currently on leave and travelling. I'll look into it when I find some time.

floroeske:
Not sure if it's a downside but it joins the multicast on all linux network devices. Including virtual devices (like tun and VPN). I am not sure if it creates any unwanted behaviour. I guess it depends on the netowrk setup.

I haven't noticed anything weird yet. I would have liked to just join the group on physical interfaces. But I couldn't find anything in "ifaddrs" to filter for them.

Another option would have been to white/blacklist interfaces/networks in ecal.ini. I think dds (cyclonedds) does has some options for it.

FlorianReimold:
> I would have liked to just join the group on physical interfaces. But I couldn't find anything in "ifaddrs" to filter for them.

I think that would have had the downside that it would prevent running eCAL e.g. on a virtual VLAN interface. I would like to discuss the implications with the other eCAL developers, who are all on vacation at the moment. So I am leaving this open for some more time...

FlorianReimold:
We talked about the feature and decided that it would be shame to remove it with a compile time flag. I think you oriented on the `ECAL_JOIN_MULTICAST_TWICE` option, but that was added for a very specific proprietary hardware setup that had a (kernel?) issue which made it necessary to always join multicast groups twice for receiving multicast data. It is an awful piece of code for an awful issue that nobody should ever use on any other system.

Your join-multicast-on-all-devices option on the other hand may be really useful for anybody. Thus we think it should be added as a runtime parameter that the user can enable or disable without recompiling eCAL.

So here is how that works:

- There must be an option in the `ecal.ini` configuration file, probably below this line:
    https://github.com/eclipse-ecal/ecal/blob/b4875ce42932db0ba31c9ce45b2f358f192dad3c/ecal/core/cfg/ecal.ini#L42
    We think the default should be `false` for now, in order to mimic the current behavior.

- The `ecal_def_ini.h` must be expanded with the name of that new option
    https://github.com/eclipse-ecal/ecal/blob/master/ecal/core/src/ecal_def_ini.h

- The `eCAL::Config` API must provide a function for getting the value:
    https://github.com/eclipse-ecal/ecal/blob/master/ecal/core/include/ecal/ecal_config.h
    https://github.com/eclipse-ecal/ecal/blob/master/ecal/core/src/ecal_config.cpp

- Then instead of the `#ifdef` compile flag you would check the newly introduced `eCAL::Config` function and join on all interfaces, if it is set to `true`

Do you feel comfortable to implement that?

floroeske:
Looking into this again.

floroeske:
Let me know if I you want any more changes. I'm happy to rename things if required.

FlorianReimold:
Thank you very much for the implementation! Looks good to me so far. Last thing I am going to do is to test it out in a VM and then merge it! 😃

FlorianReimold:
Merged with #843 👍

Issue #787, opened 2022-09-08
Meas Cutter misses find_project(yaml-cpp) call


FlorianReimold:
eCAL Meas Cutter currently missses the mentioned `find_package` call. When compiling on Linux without having yamlcpp installed, the user will be prompted with the following message:

![image](https://user-images.githubusercontent.com/11774314/189165925-26473820-f4e7-441d-a402-11e29ad29c63.png)

I think it is beneficial to have a proper find_package call.

PR #791, opened 2022-09-13
Feature/meas cutter add find

adriana-chirila:
**Pull request type**

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

No find_package call for yaml-cpp.

Issue Number: #787

**What is the new behavior?**

Added find_package call for yaml-cpp.
If yaml-cpp package is not installed, will receive the following errors at CMake time:
image

- [ ] Yes
- [x] No

**Other information**


FlorianReimold:
Please confirm that it is necessary to have the dash in `ECAL_THIRDPARTY_BUILD_YAML-CPP`. It looks stupid that way, because everything else is written with underscores 🙄. But I think I figured out why you did it (the name is used to define a target). If it is necessary, we merge it with the dash.

adriana-chirila:
It is necessary because this is both how the system package is named on Linux and the generated build name on Windows.

Issue #800, opened 2022-09-26
Samples: Create binary Pub/Sub sample (C++)


FlorianReimold:
# Current state

Currently, we have the minimal pub/sub samples in C++ that are not really that minimal, as they use std::string to send data and therefore use the msg/string class.
https://github.com/eclipse-ecal/ecal/tree/master/samples/cpp/minimal

This does not really reflect the "low level" C++ API that at the end just sends a memory blob by pointer and size

# Goal

The goal is to have a sample pair that sends a binary blob by **pointer** and **size**. The reason for this is to demonstrate that the user doesn't have to use a serialization format like protobuf to send anything else than an std::string.

chengguizi:
On this line, How about the binary support on Python API as well?

Currently, there are only `StringSubscriber` and `ProtoSubscriber` implemented. I can work around to read the raw bytes by removing the `decode()` function in the  `StringSubscriber` , like this
```
  def _on_receive(self, topic_name, msg, time):
    self.callback(topic_name, msg, time)    
```

Is it desired to have `ByteSubscriber` and perhaps `CapnPotoSubscriber` etc implemented in Python API as well? I could probably work on a PR if needed, as the change seems to be pure Python.

PR #810, opened 2022-09-28
add binary pub/sub c++ sample (#800)

rex-schilasky:
**Pull request type**

 

 

- [X] Feature

Binary publish / subscribe sample missing.

Issue Number: #800

**What is the new behavior?**
Binary publish / subscribe sample added.

### DIFF
--- a/samples/cpp/binary/binary_rec/src/binary_rec.cpp
+++ b/samples/cpp/binary/binary_rec/src/binary_rec.cpp
@@ -1,0 +1,51 @@
+/* ========================= eCAL LICENSE =================================
+ *
+ * Copyright (C) 2016 - 2019 Continental Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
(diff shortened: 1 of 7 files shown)


samples/cpp/binary/binary_rec/src/binary_rec.cpp
```diff
@@ -0,0 +1,51 @@
...
+  std::cout << "-------------------------------" << std::endl;
+
+  // initialize eCAL API
+  eCAL::Initialize(argc, argv, "binary_rec");
+
+  // subscriber for topic "blob"
+  eCAL::CSubscriber sub("blob");
+
+  // receive updates
+  std::string bin_buffer;
```
FlorianReimold, line 37: I don't like using an `std::string` here at all, as strings usually refer to a textual message and not a binary blob. We should use anything else that doesn't look like being text as an exampel. I think a `char[]` would be ok, or `std::vector<char>` / `std::array<char>`.

FlorianReimold (changes requested):

FlorianReimold (approved):
Publish & Receive with callback both use binary classes, but the ReceiveBuffer has to use the `std::string` due to its API. So it's fine for me. 👍

PR #817, opened 2022-10-07
Feature: use ecal config API

MihaiIMuraru:
**Pull request type**

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: N/A

**What is the new behavior?**

-
-
-

- [ ] Yes
- [x] No

**Other information**


FlorianReimold:
Looks good to me!

Can you also try if we now can remove `convert_utf.cpp / .h` from all Projects that don't link against Simpleini anymore? I think those files were only used for Simpleini. But they are deprecated and the License is awfull, so removing as much of them as possible would be great 😉

MihaiIMuraru:
On it! :)

PR #819, opened 2022-10-10
Rec: Fixed escaping of CURL URIs

### LINKED ISSUE #788: Rec 1-file-per-channel: Issue when uploading measurements
# The Bug

I detected a strange issue with eCAL Rec on different versions of Ubuntu (-> different libcurl versions?). When using the new "one file per channel" option, eCAL Rec fails to upload the measurement, if:

- One file contains a `#` (All OS)
- One file contians a space ` ` (Only Ubuntu 22.04)

I haven't tested on Windows, yet.

eCAL Rec client prints the following error (taken from a measurement with a space on Ubuntu 22.04)

```
Uploading measurement 34548846002
Scanning directory for upload: /home/florian/ecal_meas/2022-09-09_09-38-16.056_measurement
Found 7 files with total 310879 bytes.
Start uploading to ftp://34548846002:34548846002@florian-ubuntu20:34609//
Error uploading: URL using bad/illegal format or missing URL (/home/florian/ecal_meas/2022-09-09_09-38-16.056_measurement/florian-ubuntu22/florian-ubuntu22_hello_world_topic florian-ubuntu22.hdf5)
Finished uploading.
```

# Investigation so far

- As far as I know, the "URL using bad/illegal format or missing URL" comes from libcurl.
- Hashes fail to be uploaded from all operating systems, Spaces only from Ubuntu 22.04
- Having "real" UTF-8 characters in filenames like an Emoji `😑` works, only spaces fail

- libcurl versions:
    - Ubuntu 18.04: `7.58.0-2ubuntu3.20`
    - Ubuntu 20.04: `7.68.0-1ubuntu2.13`
    - Ubuntu 22.04: `7.81.0-1ubuntu1.4`
    - Windows Submodule: `curl-7_69_1`

_Maybe something changed in curl between 7.68 and 7.81?_

# To reproduce

Option 1: publish a channel containing a space and record it with eCAL Rec in a distributed recording using the one-file-per-channel option. Then upload it
Option 2: Create any distributed recording. Insert a space in any filename in the measurement directory on the remote machien. Then click upload

FlorianReimold:
This fixes a bug that prevented uploading files with spaces, hashes etc. from eCAL Rec.

Fixes #788


app/rec/rec_client_core/src/job/ftp_upload_thread.cpp
```diff
@@ -50,6 +52,273 @@ namespace eCAL
...
+        false,  //  7       (0x37)
+        false,  //  8       (0x38)
+        false,  //  9       (0x39)
+        false,  //  :       (0x3A)
+        false,  //  ;       (0x3B)
+        false,  //  <       (0x3C)
+        false,  //  =       (0x3D)
+        false,  //  >       (0x3E)
+        false,  //  ?       (0x3F)
+        false,  //  @       (0x40)
```
FlorianReimold, line 127: We have to escape that as well

KerstinKeller (approved):

PR #833, opened 2022-10-13
eCAL Play @ Linux: Increase FD Limit

### LINKED ISSUE #592: Play@Linux: Increase limit of file descriptors to not crash when hitting it
**Background:**
On Linux, the OS restricts the number of open file descriptors by process. eCAL Play needs a lot of file descriptors when replaying a huge measurement due to all of the shared memory files.

https://continental.github.io/ecal/configuration/player.html

Most new users don't know they have to configure something. Many experienced users also forget about it.
Currently Qt is so kind to catch the helpful exception and rethrow a generic one that is not helpful at all, so not even the command line displays an actual error message.

**Goal**
The goal is to better tell the user why their Player may be crashing.

**Ideas**
- Check max. amount of open FDs and warn the user on first start when the amount is below a magic-number
- Implement a heuristic that computes from the amount of channels how many file descriptors are probably needed when loading a measurement. Then compare it to the current maximum and warn the user when there may be an issue
- Properly handle the error and don't throw and exception. This would probably mean passing the whole error through the eCAL API and is most likely the hardest way to tackle this.

FlorianReimold:
eCAL Play GUI and eCAL Play CLI now both raise the open-fd-soft-limit up to the hard-limit.

Fixes #592

### DIFF
--- a/app/play/play_cli/src/ecal_play_cli.cpp
+++ b/app/play/play_cli/src/ecal_play_cli.cpp
@@ -48,6 +48,9 @@
 #pragma warning(pop)
 #endif //_MSC_VER
 
+#ifdef ECAL_OS_LINUX
+#include <sys/resource.h>
+#endif
 
 #ifdef ECAL_OS_WINDOWS
 #include <conio.h>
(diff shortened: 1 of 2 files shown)


KerstinKeller (approved):
I do not know if this check should also be added to the recorder.

PR #836, opened 2022-10-14
Feature/sub receive cv

### LINKED ISSUE #834: Synchronous receive function sometime returns true even if no data was received
**Describe the bug**
Every now and then, using `subscriber.Receive(proto_msg)` returns true and an empty (=default contructed) protobuf message object.
It must be that calling
https://github.com/eclipse-ecal/ecal/blob/f87f8242db312d62743c0b406cc90b856a7ad912/ecal/core/include/ecal/msg/subscriber.h#L156
returns true, even if no data was actually received.

**To Reproduce**
Run person_snd (at 50Hz) and person_rec (see code below).
at some point it breaks with and empty protobuf object.
image

**Expected behavior**
No default constructed objects should b

**Receiver code**
```
#include <ecal/ecal.h>
#include <ecal/msg/protobuf/subscriber.h>

#include <iostream>

#include "person.pb.h"


{
  // initialize eCAL API
  eCAL::Initialize(argc, argv, "person subscriber");

  // set process state
  eCAL::Process::SetState(proc_sev_healthy, proc_sev_level1, "I feel good !");

  // create a subscriber (topic name "person")
  eCAL::protobuf::CSubscriber<pb::People::Person> sub("person");

  pb::People::Person person;

  // receive updates
  std::string rcv_content;
  while (eCAL::Ok())
  {
    // receive content
    if (sub.Receive(person))
    {
      if (person.id() == 0)
      {
        std::cout << "o oh... error!!!";
      }
    }
    eCAL::Process::SleepMS(10);
  }

  // finalize eCAL API
  eCAL::Finalize();

  return(0);
}
```

KerstinKeller:
**Pull request type**

- [x] Bugfix

Sometimes, the synchronous receive function would return true with empty data, e.g. even though nothing was received.

Fixes #834

**What is the new behavior?**
It correctly returns false when no data was received.

- [ ] Yes
- [x] No

**Other information**

Please review this PR carefully! It was tested, but not extensively.
@FlorianReimold: can we safely use this or will we run in trouble regarding time jumps (back / forth) like we did before? (e.g. may there be the chance that it never times out?)

### DIFF
--- a/ecal/core/src/readwrite/ecal_reader.cpp
+++ b/ecal/core/src/readwrite/ecal_reader.cpp
@@ -356,17 +356,21 @@ namespace eCAL
 
   bool CDataReader::Receive(std::string& buf_, long long* time_ /* = nullptr */, int rcv_timeout_ms_ /* = 0 */)
   {
-    if(!m_created) return(false);
+    if (!m_created) return(false);
 
     std::unique_lock<std::mutex> read_buffer_lock(m_read_buf_mutex);
 
-    if (rcv_timeout_ms_ < 0)
+    // No need to wait (for whatever time) if something has been received)
+    if (!m_read_buf_received)
     {
-      m_read_buf_cv.wait(read_buffer_lock, [this]() { return this->m_read_buf_received; });
-    }
-    else if (rcv_timeout_ms_ > 0)
-    {
-      m_read_buf_cv.wait_for(read_buffer_lock, std::chrono::milliseconds(rcv_timeout_ms_), [this]() { return this->m_read_buf_received; });
+      if (rcv_timeout_ms_ < 0)
+      {
+        m_read_buf_cv.wait(read_buffer_lock, [this]() { return this->m_read_buf_received; });
+      }
+      else if (rcv_timeout_ms_ > 0)
+      {
+        m_read_buf_cv.wait_for(read_buffer_lock, std::chrono::milliseconds(rcv_timeout_ms_), [this]() { return this->m_read_buf_received; });
+      }
     }
 
     // did we receive new samples ?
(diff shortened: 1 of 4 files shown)


FlorianReimold (approved):
Looks good to me!
Please squash-merge with a consise commit message.

ecal/core/src/readwrite/ecal_reader.cpp
```diff
@@ -359,28 +354,40 @@ namespace eCAL
     return(true);
   }
 
-  bool CDataReader::Receive(std::string& buf_, long long* time_ /* = nullptr */, int rcv_timeout_ /* = 0 */)
+  bool CDataReader::Receive(std::string& buf_, long long* time_ /* = nullptr */, int rcv_timeout_ms_ /* = 0 */)
   {
     if(!m_created) return(false);
 
+    std::unique_lock<std::mutex> read_buffer_lock(m_read_buf_mutex);
+
```
KerstinKeller, line 362: @FlorianReimold, I wonder if we should only wait on the condition variable, if no data is present 
```
if (!m_read_buf_received)
```
How does the cv work? Does it "rember" signals? Or only receive signals while it is waiting?

ecal/core/src/readwrite/ecal_reader.cpp
```diff
@@ -356,17 +356,21 @@ namespace eCAL
...
 
     std::unique_lock<std::mutex> read_buffer_lock(m_read_buf_mutex);
 
-    if (rcv_timeout_ms_ < 0)
-    {
-      m_read_buf_cv.wait(read_buffer_lock, [this]() { return this->m_read_buf_received; });
-    }
-    else if (rcv_timeout_ms_ > 0)
+    // No need to wait (for whatever time) if something has been received)
+    if (!m_read_buf_received)
```
FlorianReimold, line 364: This is basiaclly dupliated code, as the predicate of the condition variable does exactly that

FlorianReimold: What you imlemented here is:

``` cpp

if (m_read_buf_received)
{
  while (m_read_buf_received)
  {
    cv.wait(lock);
  }
}

PR #849, opened 2022-10-26
REC: added browse button for selecting the root folder of the measurement(s)

### LINKED ISSUE #831: eCAL Rec GUI: Add graphical selection of Measurement root dir
# Feature request

Add a folder selection in the eCAL Rec GUI that let's the user select the measurement root dir on their local machine.

# Original issue
> We often want to change the location, where ecal measurements are stored, in ecal rec. Default place in our config is /home/continental/meas/ and I want e.g. store them in /media/continental/usbsticks01. Right now, when I click on this fiel in ecal rec, I have to change this in text mode! I then have to start a terminal in the wished folder, copy the path there and paste it in the text window. No check, a typo leads to bad things.
>
> We need here a select window, either standard linux style where I can select a location/path and then click in the top right border "select" (same as in linux file explorer when I select files and click "copy to" or "move to"). Or even better, scan all usb drives and make suggestions. We need default or usb ssd usually.
>
> Our workaround was to save the special path in a config but this seems to cause other problems as then sometimes some channels are not measured, seems to be a bug in ecal rec.

MihaiIMuraru:
**Pull request type**

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Fixes #831 

**What is the new behavior?**

-
-
-

- [ ] Yes
- [x] No

**Other information**


FlorianReimold (changes requested):
I have noticed the following 3 things:

1. Use the OS-native file dialog, unless there really is a reason not to do that.
2. The start directory for the file dialog should be the currently selected path that the user either loaded from a or typed in manually
3. The TAB order is messy (i.e. the order of input fields the cursor is jumping to when pressing the tab key)

PR #868, opened 2022-11-03
shm memory file mutex should not be unlinked on subscriber destruction

### LINKED ISSUE #854: subscriber should not unlink topic shm mutex file during it's own destroying
## Describe the bug
say:
1.  publisher is publishing topic t1, which will internally created xxx_shm and xxx_shm_mtx
2.  subscirber1 subscribed t1 and consumed messages normally, then during destroying, it [unlinked](https://github.com/eclipse-ecal/ecal/blob/55d597d99a2b5aaac0f3a6954f3241d6fc6d2b3b/ecal/core/src/io/ecal_memfile.cpp#L155) xxx_shm_mtx
3.  subscriber 2 subscribed, but being unable to located xxx_shm_mtx, it will create another xxx_shm_mtx by itself, which is different than the one holding by publisher now.
4.  subscirber2 can consume message as if everthing is normal, but actually the xxx_shm's read/write is no longer exclusively guarded. 

## To Reproduce
see above

## Desktop 
OS: Ubuntu1804
version: latest master

rex-schilasky:
shm memory file mutex should not be unlinked if memory file is not owned (created) by the caller (fixing issue #854)

**Pull request type**

- [X] Bugfix

See issue description.

Fixes #854

- [X] No

### DIFF
--- a/ecal/core/src/io/ecal_memfile.cpp
+++ b/ecal/core/src/io/ecal_memfile.cpp
@@ -148,11 +148,19 @@ namespace eCAL
     // destroy memory file
     ret_state &= memfile::db::RemoveFile(m_name, remove_);
 
-    // unlock mutex
-    ret_state &= DestroyMtx(&m_memfile_info.mutex);
+    if (remove_)
+    {
+      // destroy mutex
+      ret_state &= DestroyMtx(&m_memfile_info.mutex);
 
-    // cleanup mutex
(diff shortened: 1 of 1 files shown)


hannemn:
I would propose to fix this issue on my shm-monitoring branch, since i removed the current memfile mutex api and replaced them with a RAII style NamedMutex class.

PR #873, opened 2022-11-10
topic2mcast chopped group base as if using hard coded mask 0.255.255.255

tsingakbar:
**Pull request type**

- [x] Bugfix

UDP multicast data write will use a calculated multicast group address, but during deployment and testing, we found the calculation result is sometimes confusing.
```ini
[network]
multicast_group = 238.202.203.0
multicast_mask = 0.0.0.255
```
The [calculated](https://github.com/eclipse-ecal/ecal/blob/543ccdf1df9417906ce45172438aecff8ed5ab15/ecal/core/src/topic2mcast.h#L60) address is something like `238.0.0.x` while `238.202.203.x` is expected.
After looking into the calculation, I think the logic always assumes `0.255.255.255` is used as mask.

Issue Number: N/A

**What is the new behavior?**
The PR will apply the configured mask both to the base and calculated part of the result address.

- [ ] Yes
- [x] No

### DIFF
--- a/ecal/core/src/topic2mcast.h
+++ b/ecal/core/src/topic2mcast.h
@@ -48,6 +48,21 @@ namespace eCAL
     }
   };
 
+  inline uint32_t parse_ipv4(const std::string& ipv4_str_)
+  {
+    uint32_t ipv4 = 0;
+    std::stringstream ss(ipv4_str_);
+    std::string token;
+    int i = 0;
+    while(std::getline(ss, token, '.') && (i < 4)) //-V112
+    {
+      unsigned char mask = static_cast<unsigned char>(atoi(token.c_str()));
+      ipv4 = ipv4 << 8 | mask;
+      i++;
+    }
+    return ipv4;
+  }
+
   /**
    * @brief Get multicast address based on the topic_name.
    *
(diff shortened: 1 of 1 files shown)


rex-schilasky:
@tsingakbar first of all - WELCOME IN THE TEAM ! Great contribution. We were on a Hackathon this week and didn't find the time to check your PR. We know that configuration logic issue, but we need to solve it in compatible way.
We will hopefully get it checked the next days.

KerstinKeller:
As discussed, we will approach it in the following way:

- Keep both old (incorrect) and correct way of masking
- Per default, the correct way is activated
- Provide new default values, that after masking have the same value as default values before using the wrong calculation method

This way, compatibility with other ecal versions can be achieved.
The old way is deprecated and to be removed in an upcoming minor / major release.

FlorianReimold:
Closing this, as it was solved by #950. Special thanks to you, @tsingakbar, for this contribution. We built up on it, but made it compatible with the old crappy way of computing Multicast addresses; i.e. we just left the entire old code in place, added the new way at let the user select from the `ecal.ini`.

The enhancement will be released with eCAL 5.12, but the documentation is already in place: https://eclipse-ecal.github.io/ecal/configuration/options.html#cmdoption-arg-multicast_config_version

PR #876, opened 2022-11-15
shm handshake timeout CPublisher API (fixing issue #875)

### LINKED ISSUE #875: Shared memory handshake API
In some setups there is a need to define the shm handshake mechanism for specific publisher / subscriber connections only. Currently we can set this feature by defining a global shm handshake timeout in the ecal.ini configuration file only.

Provide a new `CPublisher` API function like that

```cpp
    /**
     * @brief Force connected subscribers to send acknowledge event after processing the message and 
     *        block publisher send call on this event with a timeout.
     *
     * Most applications perform very well with the default behavior. If subscribers are too slow 
     * to process incoming messages then the overall software architecture needs to be checked, software components 
     * need to be optimized or parallelized.
     * 
     * There may still be cases where it could make sense to synchronize the transfer of the payload from a publisher 
     * to a subscriber by using an additional handshake event. This event is signaled by a subscriber back to the 
     * sending publisher to confirm the complete payload transmission and the processed subscriber callback.
     * 
     * The publisher will wait up to the specified timeout for the acknowledge signals of all connected subscribers 
     * before sending new content. Finally that means the publishers CPublisher::Send API function call is now blocked 
     * and will not return until all subscriber have read and processed their content or the timeout has been reached.
     * 
     * @param acknowledge_timeout_ms_ timeout to wait for acknowledge signal from connected subscriber in ms (0 == no handshake).
     *
     * @return  True if it succeeds, false if it fails.
    **/
    bool ShmSetAcknowledgeTimeout(int acknowledge_timeout_ms_);
```

rex-schilasky:
**Pull request type**

Provide new shm handshake timeout API for the `CPublisher`.

- [X] Feature

Fixes #875

### DIFF
--- a/ecal/core/include/ecal/ecal_publisher.h
+++ b/ecal/core/include/ecal/ecal_publisher.h
@@ -29,6 +29,7 @@
 #include <ecal/ecal_qos.h>
 #include <ecal/ecal_tlayer.h>
 
+#include <chrono>
 #include <string>
 
 namespace eCAL
@@ -264,6 +265,23 @@ namespace eCAL
      * @return  True if it succeeds, false if it fails.
     **/
     bool ShmSetAcknowledgeTimeout(int acknowledge_timeout_ms_);
+
+    /**
+     * @brief Force connected subscribers to send acknowledge event after processing the message and
+     *        block publisher send call on this event with a timeout.
+     *
+     * See ShmSetAcknowledgeTimeout(int acknowledge_timeout_ms_)
+     * 
+     * @param acknowledge_timeout_ms_ timeout to wait for acknowledge signal from connected subscriber (0 == no handshake).
+     *
+     * @return  True if it succeeds, false if it fails.
+    **/
+    template <typename Rep, typename Period>
+    bool ShmSetAcknowledgeTimeout(std::chrono::duration<Rep, Period> acknowledge_timeout_)
+    {
+      auto acknowledge_timeout_ms = std::chrono::duration_cast<std::chrono::milliseconds>(acknowledge_timeout_).count();
+      return ShmSetAcknowledgeTimeout(acknowledge_timeout_ms);
+    }
 
     /**
      * @brief Set the the specific topic id.
--- a/ecal/core/include/ecal/ecal_publisher.h
+++ b/ecal/core/include/ecal/ecal_publisher.h
@@ -315,6 +315,8 @@ namespace eCAL
 
     /**
      * @brief Send a message to all subscribers synchronized with acknowledge timeout (see also ShmSetAcknowledgeTimeout).
+     * 
+     * This synchronized mode is currently implemented for local interprocess communication (shm-ecal layer) only.
      *
      * @param buf_                     Pointer to content buffer.
      * @param len_                     Length of buffer.
--- a/ecal/core/include/ecal/ecal_publisher.h
+++ b/ecal/core/include/ecal/ecal_publisher.h
@@ -215,9 +215,9 @@ namespace eCAL
     bool ShmSetBufferCount(long buffering_);
 
     /**
-     * @brief Enable zero copy shared memory trasnport mode.
+     * @brief Enable zero copy shared memory transport mode.
      *
-     * By default, the builtin shared memory layer is configured to make one memory copy
+     * By default, the built-in shared memory layer is configured to make one memory copy
      * on the receiver side. That means the payload is copied by the internal eCAL memory pool manager
      * out of the memory file and the file is closed immediately after this.
      * The intention of this implementation is to free the file as fast as possible after reading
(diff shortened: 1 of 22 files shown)


ecal/core/src/io/ecal_memfile_header.h
```diff
@@ -29,19 +29,21 @@ namespace eCAL
...
+    uint64_t   hash       = 0;
     // ----- > 5.8 -----
     struct optflags
     {
       unsigned char zero_copy : 1;    // allow reader to access memory without copying
       unsigned char unused    : 7;
     };
     optflags   options = { 0, 0 };
+    // ----- > 5.11 ----
+    uint64_t   ack_timout = 0;
```
FlorianReimold, line 46: please add the unit (probably milliseconds) to avoid confusion

ecal/core/src/readwrite/ecal_writer.cpp
```diff
@@ -526,14 +535,15 @@ namespace eCAL
...
-        wdata.zero_copy = m_zero_copy;
+        wdata.buf                 = buf_;
+        wdata.len                 = len_;
+        wdata.id                  = m_id;
+        wdata.clock               = m_clock;
+        wdata.hash                = snd_hash;
+        wdata.time                = time_;
+        wdata.buffering           = m_buffering_shm;
+        wdata.zero_copy           = m_zero_copy;
+        wdata.acknowledge_timeout = m_acknowledge_timeout;
```
FlorianReimold, line 546: please add the unit (probably milliseconds)  to the variable name to avoid confusion

ecal/core/src/readwrite/ecal_writer.h
```diff
@@ -132,6 +133,7 @@ namespace eCAL
 
     size_t             m_buffering_shm;
     bool               m_zero_copy;
+    long               m_acknowledge_timeout;
```
FlorianReimold, line 136: please add the unit (probably milliseconds)  to the variable name to avoid confusion. This may even be a properly typed chrono time object

ecal/core/src/readwrite/ecal_writer_data.h
```diff
@@ -29,15 +29,16 @@ namespace eCAL
...
+    size_t       len                 = 0;
+    long long    id                  = 0;
+    long long    clock               = 0;
+    size_t       hash                = 0;
+    long long    time                = 0;
+    size_t       buffering           = 1;
+    long         bandwidth           = 0;
+    bool         loopback            = false;
+    bool         zero_copy           = false;
+    long         acknowledge_timeout = 0;
```
FlorianReimold, line 42: please add the unit (probably milliseconds)  to the variable name to avoid confusion. This may even be a properly typed chrono time object, if the struct will never be used in C

ecal/core/include/ecal/ecal_publisher.h
```diff
@@ -243,6 +243,28 @@ namespace eCAL
...
+     * 
+     * The publisher will wait up to the specified timeout for the acknowledge signals of all connected subscribers 
+     * before sending new content. Finally that means the publishers CPublisher::Send API function call is now blocked 
+     * and will not return until all subscriber have read and processed their content or the timeout has been reached.
+     * 
+     * @param acknowledge_timeout_ms_ timeout to wait for acknowledge signal from connected subscriber in ms (0 == no handshake).
+     *
+     * @return  True if it succeeds, false if it fails.
+    **/
+    bool ShmSetAcknowledgeTimeout(int acknowledge_timeout_ms_);
```
FlorianReimold, line 266: Maybe it would be better to accept an std::chrono::milliseconds here, if possible

FlorianReimold (changes requested):

KerstinKeller:
@rex-schilasky the compatibility with old subscribers / new publishers works only in "handshake off" mode, correct?

rex-schilasky:
> @rex-schilasky the compatibility with old subscribers / new publishers works only in "handshake off" mode, correct?

1. New publisher / old subscriber
- if only ecal.ini is used: full compatibility
- if publisher uses API to switch handshake OFF it will be OFF
- if publisher uses API to switch handshake ON and ecal.ini defines OFF it will be OFF

2. Old publisher / new subscriber
- Feature is always OFF (new subscriber is no more using ecal.ini settings)

KerstinKeller:
I think we do not have to care 100% about the compatibility part. More important is that it will work 100% if all eCAL participants use eCAL 5.11.

ecal/core/include/ecal/ecal_publisher.h
```diff
@@ -237,14 +238,53 @@ namespace eCAL
...
+     *        block publisher send call on this event with a timeout.
+     *
+     * See ShmSetAcknowledgeTimeout(int acknowledge_timeout_ms_)
+     * 
+     * @param acknowledge_timeout_ms_ timeout to wait for acknowledge signal from connected subscriber (0 == no handshake).
+     *
+     * @return  True if it succeeds, false if it fails.
+    **/
+    template <typename Rep, typename Period>
+    bool ShmSetAcknowledgeTimeout(std::chrono::duration<Rep, Period> acknowledge_timeout_)
```
hannemn, line 280: Of course this is the proper way to define durations in C++, but for all other public API functions in eCAL we just use a simple 64 bit integer. For consistency reasons we should maybe keep it similiar to the other function signatures.

PR #877, opened 2022-11-15
CMake build on macos should use Python 3.9...

KerstinKeller:
Modify the github actions in such a way that they actually use the Python 3.9 installation.


KerstinKeller:
@FlorianReimold  I cancelled the branch builds because I only modified the actions. It seems that it is now finding the right Python version. Maybe there was a CMake update which changed the behavior and caused breaking.

FlorianReimold (approved):
LGTM 👍

PR #883, opened 2022-11-22
Made eCAL::app_pb, ::core_pb and ::time_pb shared libraries on Linux

### LINKED ISSUE #687: compile ecal::pb as .so/dll file on Linux/Windows

FlorianReimold:
On Linux, the targets are hardcoded compiled as shared objects. On Windows they are hardcoded as static libs. Even though this may not be optimal, it fixes #687.


KerstinKeller (approved):
i like the very extensive comment!

Issue #880, opened 2022-11-21
Add `CDataWriter::SetTopicType()` API


FlorianReimold:
**Is your feature request related to a problem? Please describe.**
The issue arose from the eCAL -> MQTT->eCAL Gateway setup. The gateway works like follows:

- The data is published from eCAL to an MQTT topic and back to eCAL again on the other side. The topic names have to be configured manually.
- The description and type from eCAL are both published on 2 different MQTT Topics. The information is published periodically. When now going back to eCAL, the Description can be set, even after the publisher has been created. The API however is not consistent and the Topic Type cannot be set anymore, it would have to been known on time of publisher creation.

**Describe the solution you'd like**
There should be an API to set the Topic Type after the Publisher has been created, just as it is the case for the Description.

https://github.com/eclipse-ecal/ecal/blob/528c0520f17a3cb6df604259794aa547c8e3e59c/ecal/core/src/readwrite/ecal_writer.h#L59

KerstinKeller:
Are we sure that we want this functionality, and we cannot solve this in any other way?

There are a few things we need to make sure:
- Descriptions / Topic Type / Topic name are only set once (per publisher)

We can add this, however I'd like to have a discussion about the bigger picture, especially in regard to #797.

PR #890, opened 2022-11-28
adding new publisher API function, SetType (fixing issue #880)

rex-schilasky:
**Pull request type**

Adding new SetType API function for eCAL Publisher. (Maybe we call it SetTopicType, maybe not. I called it SetType to be in synch with SetDescription).

- [X] Feature

Fixes: #880

### DIFF
--- a/ecal/core/include/ecal/ecal_clang.h
+++ b/ecal/core/include/ecal/ecal_clang.h
@@ -224,7 +224,7 @@ ECAL_API bool pub_destroy(ECAL_HANDLE handle_);
  *
  * @return  True if succeeded.
 **/
-ECAL_API bool pub_set_type(ECAL_HANDLE handle_, const char* topic_type_, const int topic_type_length_);
+ECAL_API bool pub_set_topic_type(ECAL_HANDLE handle_, const char* topic_type_, const int topic_type_length_);
 
 /**
  * @brief Setup topic type description.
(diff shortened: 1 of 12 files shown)


ecal/core/include/ecal/ecal_clang.h
```diff
@@ -224,7 +235,7 @@ ECAL_API bool pub_destroy(ECAL_HANDLE handle_);
  *
  * @return  True if succeeded.
 **/
-ECAL_API bool pub_set_description(ECAL_HANDLE handle_, const char* topic_desc_, const int topic_desc_length_);
+ECAL_API bool pub_set_topic_description(ECAL_HANDLE handle_, const char* topic_desc_, const int topic_desc_length_);
```
FlorianReimold, line 238: This breaks the public API. If we want to rename this function, we should offer the old one as deprecated function as well.

lang/python/core/ecal/core/core.py
```diff
@@ -175,16 +175,27 @@ def pub_destroy(topic_handle):
   return _ecal.pub_destroy(topic_handle)
 
 
-def pub_set_description(topic_handle, description):
-  """ set publisher description
+def pub_set_topic_type(topic_handle, topic_type):
```
FlorianReimold, line 178: This breaks the public API

FlorianReimold (changes requested):

PR #895, opened 2022-12-01
Core+Python: Added set-topic-type-name API function

### LINKED ISSUE #880: Add `CDataWriter::SetTopicType()` API

FlorianReimold:
Changes for C++:
- Added new API function: `CPublisher::SetTypeName()`

Changes for C:
- Added new API function `pub_set_type_name()`

Changes for Python:
- Added 2 new API functions:
    - `set_topic_description()`
    - `set_topic_type_name()`

This PR fixes #880

This PR closes #890 as it is a direct API-compatible version. Compared to the old PR, this PR:
1. is API compatible and doesn't refactor public API functions
2. Accepts setting the Topic Type Name in Python as String (the old PR for some reason required a binary bytes field as topic name)


lang/python/core/ecal/core/core.py
```diff
@@ -175,16 +175,27 @@ def pub_destroy(topic_handle):
...
+
+  :param topic_handle:    the topic handle
+  :type topic_handle:     string
+  :param topic_type_name: the topic type name
+  :type topic_type_name:  string
+
+  """
+  return _ecal.pub_set_topic_type_name(topic_handle, topic_type_name)
+
+def pub_set_description(topic_handle, topic_desc):
```
KerstinKeller, line 189: @FlorianReimold renaming arguments in Python is (unlike in C++) also an API change, as user might use named arguments. E.g.
```
pub_set_description(description="abc")
```
would now fail.

I am not 100% sure if this file is considered public API though, or only the eCAL publisher / subscriber objects.

lang/python/core/ecal/core/core.py
```diff
@@ -580,6 +591,26 @@ def destroy(self):
...
+  def set_topic_description(self, topic_desc):
+    """ set topic description
+
+    :param topic_desc: the topic type description
+    :type topic_desc: bytes
+
+    """
+
+    return pub_set_description(self.thandle, topic_desc)
+
```
KerstinKeller, line 613: why do we have two new functions here? the other just never existed? (set_topic_description)

lang/python/core/src/ecal_wrap.cxx
```diff
@@ -334,13 +334,13 @@ PyObject* pub_destroy(PyObject* /*self*/, PyObject* args)
...
 {
   ECAL_HANDLE topic_handle = nullptr;
-  char* topic_type = nullptr;
-  Py_ssize_t  topic_type_len = 0;
+  char* topic_type_name = nullptr;
+  Py_ssize_t  topic_type_name_len = 0;
 
-  if (!PyArg_ParseTuple(args, "ny#", &topic_handle, &topic_type, &topic_type_len))
+  if (!PyArg_ParseTuple(args, "ns#", &topic_handle, &topic_type_name, &topic_type_name_len))
     return nullptr;
```
KerstinKeller, line 341: I agree that a string makes sense here.
However I would think that the constructors have previously used that functions (at least) to pass on the info.
We need to make sure, that they now take the same types!

KerstinKeller (changes requested):

lang/python/core/ecal/core/core.py
FlorianReimold: You are right, I didn't realise that when changing everything back to a compatible form. I will revert this change as well.

lang/python/core/ecal/core/core.py
FlorianReimold: I honestly have no idea, but yea, I guess the other just never existed. Maybe @rex-schilasky can comment on that.

lang/python/core/src/ecal_wrap.cxx
FlorianReimold: From python side, the args look the same. `"s"` however converts to a null-terminated C-String in UTF-8 encoding, while `"s#"` converts to a C-String in UTF-8 encoding and a length, so the C-String can contain null-characters.

For Python both will be a string.

KerstinKeller: so "ss" means "s" for the `topic_name` and `s` for `topic_type`, which means that we would have to change this also to `ss#` or `s#s#`? Because i think, descriptors may contain `\0 ` chars....

FlorianReimold: > so "ss" means "s" for the `topic_name` and `s` for `topic_type`,

Exactly

> which means that we would have to change this also to `ss#` or `s#s#`?

You could totally do that, but what are you going to do with the string lengths? We use the Python->C conversion here to get the C parameters for calling the native eCAL C functions. And those expect a regular null-terminated C-String and cannot accept a size:

https://github.com/eclipse-ecal/ecal/blob/30097609467d858d647394f17f7b08693fbccf64/ecal/core/include/ecal/ecal_clang.h#L207

So even though you then could theoretically include a null-byte in your Python string (_why would you want to  do that?_), the C API would get rid of everything after the nullbyte after all.

> Because i think, descriptors may contain `\0 ` chars....

Sure, and a descriptor also isn't a UTF-8 string, but a binary blob. That's why `y#` is used for parsing the descriptor:
https://github.com/eclipse-ecal/ecal/blob/30097609467d858d647394f17f7b08693fbccf64/lang/python/core/src/ecal_wrap.cxx#L334-L344

lang/python/core/ecal/core/core.py
rex-schilasky: This function did not exist before. So I just added it for completeness.

lang/python/core/src/ecal_wrap.cxx
rex-schilasky: For topic name and topic type I used just `s` for encoding, assuming that we do not used strings containing `\0` inside.

FlorianReimold: Noted, but I vote for using the actual conversion that is needed for the C API. And in the case of `SetTypeName` that is a `char* name` and `int length`, so `s#` is the proper conversion to get exactly that. Why would we consider using any other conversion just to compute the necessary parameters afterwards?

**Current solution:**
1. Parse with `s#` -> gives `char*` and `size`
2. use `char*` and `size` for eCAL API

**Strange solution proposed here:**
1. Parse with `s` -> null-terminated `char*`
2. use `strlen` -> gives you the `size`
3. use `char*` and `size` for eCAL API

Do we really consider using the second solution? If that is the case please elaborate why.

Or are you suggesting to also use the eCAL C API to make the usage of `s` reasonable?

rex-schilasky (approved):

KerstinKeller (approved):
:+1:

PR #904, opened 2022-12-06
handling of long time consuming rpc calls

### LINKED ISSUE #903: Long-running RPC Calls block monitoring refresh
Dear eCAL community,

In our system, we use both pub/sub and service methods a.k.a RPCs.

We have one particular RPC that can take up to 10 seconds to complete (it's not actually doing work, it's just waiting for something to happen), but when it does take that long, the process it is running in loses its subscriber information, i.e. any topic it publishes to from that point onwards does not actually make it to the subscribers.

In the `registration_timeout` in `ecal.ini` is set to 10000 (10 seconds), if I increase that number to something higher, the problem goes away, so I have a solution to my problem that works.

My question is, though, whether it's the right solution. It appears as though a long running RPC blocks some maintenance/monitoring tasks needed to maintain the registration info, and if that happens for long enough, the `registration_timeout` is hit. Do you have any guidance on long-running RPCs? Is there a way give the maintenance/monitoring tasks an explicit chance to run while the RPC is waiting, or is the official advice to either steer clear of long-running RPCs or make sure the `registration_timeout` is long enough?

Regards,
Sam

_Originally posted by @samsta in https://github.com/eclipse-ecal/ecal/discussions/900_

rex-schilasky:
- [X] Bugfix

Dear eCAL community,

In our system, we use both pub/sub and service methods a.k.a RPCs.

We have one particular RPC that can take up to 10 seconds to complete (it's not actually doing work, it's just waiting for something to happen), but when it does take that long, the process it is running in loses its subscriber information, i.e. any topic it publishes to from that point onwards does not actually make it to the subscribers.

In the registration_timeout in ecal.ini is set to 10000 (10 seconds), if I increase that number to something higher, the problem goes away, so I have a solution to my problem that works.

My question is, though, whether it's the right solution. It appears as though a long running RPC blocks some maintenance/monitoring tasks needed to maintain the registration info, and if that happens for long enough, the registration_timeout is hit. Do you have any guidance on long-running RPCs? Is there a way give the maintenance/monitoring tasks an explicit chance to run while the RPC is waiting, or is the official advice to either steer clear of long-running RPCs or make sure the registration_timeout is long enough?

Regards,
Sam

Fixes: #903

### DIFF
--- a/ecal/core/src/service/ecal_service_server_impl.cpp
+++ b/ecal/core/src/service/ecal_service_server_impl.cpp
@@ -279,49 +279,71 @@ namespace eCAL
 
   int CServiceServerImpl::RequestCallback(const std::string& request_, std::string& response_)
   {
-    std::lock_guard<std::mutex> lock(m_method_map_sync);
+    // get method callback
+    {
+      std::lock_guard<std::mutex> lock(m_method_map_sync);
+      if (m_method_map.empty()) return 0;
+    }
 
-    if (m_method_map.empty()) return 0;
+    // try to parse request
+    eCAL::pb::Request  request_pb;
+    if (!request_pb.ParseFromString(request_))
+    {
+      Logging::Log(log_level_error, m_service_name + "::CServiceServerImpl::RequestCallback failed to parse request message");
+      // return value == -1 -> try to read more bytes from the socket
+      return -1;
+    }
 
-    int success(-1);
+    // return value == 0 -> we could deserialize, no need to try to read more bytes for the server socket
+    int success(0);
+
+    // prepare response
     eCAL::pb::Response response_pb;
     auto response_pb_mutable_header = response_pb.mutable_header();
     response_pb_mutable_header->set_hname(eCAL::Process::GetHostName());
     response_pb_mutable_header->set_sname(m_service_name);
     response_pb_mutable_header->set_sid(m_service_id);
 
-    eCAL::pb::Request request_pb;
-    if (request_pb.ParseFromString(request_))
+    // get method
+    SMethod method;
+    auto& request_pb_header = request_pb.header();
     {
-      // success == 0 means we could deserialize and
-      // no need to try to read more bytes for the server
-      // socket
-      success = 0;
-      auto request_pb_header = request_pb.header();
-      response_pb_mutable_header->set_mname(request_pb_header.mname());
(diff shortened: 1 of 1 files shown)


ecal/core/src/service/ecal_service_server_impl.cpp
```diff
@@ -279,49 +279,84 @@ namespace eCAL
...
+      if (m_method_map.empty()) return 0;
+    }
+
+    // try to parse request
+    eCAL::pb::Request  request_pb;
+    if (!request_pb.ParseFromString(request_))
+    {
+      Logging::Log(log_level_error, m_service_name + "::CServiceServerImpl::RequestCallback failed to parse request message");
+      // return value == -1 -> try to read more bytes from the socket
+      return -1;
```
FlorianReimold, line 294: The error handling (setting the values in response_pb_mutable_header) is missing here.

ecal/core/src/service/ecal_service_server_impl.cpp
```diff
@@ -279,49 +279,84 @@ namespace eCAL
...
+    }
 
-    if (m_method_map.empty()) return 0;
+    // return value == 0 -> we could deserialize, no need to try to read more bytes for the server socket
+    int success(0);
```
FlorianReimold, line 298: From this point on, everything will be handled as success anyways, so there isn't really a reason to have a variable here

ecal/core/src/service/ecal_service_server_impl.cpp
```diff
@@ -279,49 +279,84 @@ namespace eCAL
...
       else
       {
-        response_pb_mutable_header->set_state(eCAL::pb::ServiceHeader_eCallState_failed);
-        std::string emsg = "Service " + m_service_name + " has no method " + request_pb_header.mname();
-        response_pb_mutable_header->set_error(emsg);
+        // increase call count
+        auto call_count = iter->second.method_pb.call_count();
+        iter->second.method_pb.set_call_count(++call_count);
+        // store (copy) method element
+        method = m_method_map[request_pb_header.mname()];
```
FlorianReimold, line 331: You already have the `iter` variable by calling `find` (maybe rename to something more meaningful, like `method_map_iter`?). It would be more performant to just use it by de-referencing it instead of finding again with `operator[]`.

PR #908, opened 2022-12-09
Feature/csharp bindings extension

lukaswerner:
**Pull request type**

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

The C# bindings in Continental.eCAL.Core.dll are lacking some of the features of eCAL and some of the existing ones are not easy to use / break if used with .NET 5+

**What is the new behavior?**

- Added minimal ServiceClient implementation
- Added minimal ServiceServer implementation
- Added overload to Publisher.Send() that takes a byte[] to bypass string conversion
- Added GetTime() to Monitoring class

- *Changed ReceiveCallbackData->data of Subscriber from string to pointer + length
- *Changed return type of GetMonitoring and GetLogging() from string to byte[] 

- [x] Yes (Marked with *)
- [ ] No

The string conversion is cumbersome to use if eCAL is used to transmit anything (e.g. byte[]) other than strings. Also if used with .NET 5 or above the string conversion outright breaks due to breaking changes in the way encodings are handled. 
I therefore suggest changing these methods or adding alternative ways to get a raw byte array which can then be handled by the consumer.

**Other information**

### DIFF
--- a/lang/csharp/Continental/eCAL/Core/ecal_clr.h
+++ b/lang/csharp/Continental/eCAL/Core/ecal_clr.h
@@ -374,6 +374,63 @@ namespace Continental
         typedef void(__stdcall * stdcall_eCAL_ReceiveCallbackT)(const char*, const ::eCAL::SReceiveCallbackData*);
       };
 
+
+      /**
+       * @brief eCAL service client class.
+       *
+       * The CServiceClient class is used to call a matching eCAL server.
+       *
+      **/
+      public ref class ServiceClient
+      {
+      public:
+          /**
+           * @brief Constructor.
+          **/
+          ServiceClient();
+
+          /**
+           * @brief Constructor.
+           *
+           * @param service_name_   Unique service name.
+          **/
+          ServiceClient(System::String^ service_name_);
+
+          /**
+           * @brief Destructor.
+          **/
+          ~ServiceClient();
+
+          /**
+           * @brief structure which contains the data for callback functions
+          **/
+          ref struct ServiceClientCallbackData
+          {
+              String^  host_name;      //!< service host name
+              String^  service_name;   //!< name of the service
+              String^  service_id;     //!< id of the service
+              String^  method_name;    //!< name of the service method
+              String^  error_msg;      //!< human readable error message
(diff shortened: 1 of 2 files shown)


lang/csharp/Continental/eCAL/Core/ecal_clr.h
```diff
@@ -492,12 +642,17 @@ namespace Continental
         /**
          * @brief Get host, process and topic protobuf string.
         **/
-        static System::String^ GetMonitoring();
+        static array<Byte>^ GetMonitoring();
```
KerstinKeller, line 646: For eCAL releases, we're sticking to semantic versioning. That means, we don't "change" any existing function signatures for public functions in order not to break any user code.
I doubt that we do have a lot of C# users, however, I'd prefer to stick to this practice.
Which would mean, that we need to introduce a new function and deprecate the old one.
Like
 ```
 [Obsolete("please use GetMonitoringBytes instead")]
 static System::String^ GetMonitoring(); 
 static array<Byte>^ GetMonitoringBytes();
 ```
 The same, of course goes for the logging.

lang/csharp/Continental/eCAL/Core/ecal_clr.h
```diff
@@ -256,10 +266,11 @@ namespace Continental
         **/
         ref struct ReceiveCallbackData
```
KerstinKeller, line 267: I feel a bit uncomforatble changing the stucture, as it is also part of the public interface.
Maybe also here we define a new struct, and offer the callback for the new struct definition?

I would assume that also samples / protobuf receiver etc would have to be modified.

KerstinKeller (changes requested):
Thanks for your contribution!
I suspect that we have to iterate a bit on this, especially regarding API stability.
I think the samples are also missing.

I would really like to include this in the eCAL 5.11 release, otherwise we would have to wait quite a while to include this in eCAL 5.12, however we wanted to release soon, so we need to see what we can do!

KerstinKeller:
@lukaswerner  thanks for the contribution, everything is to be merged now on separate PRs.

PR #911, opened 2022-12-09
Submodules: Updated FTXUI to a commit that is compilable with gcc 12

FlorianReimold:
Unfortunatelly, FTXUI 3.0.0 cannot be compiled with gcc12, so I upgraded to a newer unreleased commit.
According to https://github.com/ArthurSonzogni/FTXUI/issues/520, currently master/HEAD cannot be compiled with gcc 12 either, so I used the last commit before the PR mentioned in that issue was merged.


ArthurSonzogni:
Sorry about this.
I think I fixed the issue with gcc12
https://github.com/ArthurSonzogni/FTXUI/commit/a52b959f660bff301cf8e03d9fcfa491fcc0f0dc

Issue #913, opened 2022-12-13
SHM Monitoring: Crashes & issues to find other processes


FlorianReimold:
This issue is about 2 (or more?) issues that seem to be related to the SHM Monitoring and maybe have the same cause. I am nevertheless describing them separately here.

The setup: 
- OS: Ubuntu 22.10 and 22.04, probably other Ubuntu Versions as well
- eCAL Version: master (probably equivalent to what will be released as 5.11.0 in a few days)
- eCAL.ini:
  - Network is turned off
  - shm monitoring is turned on
  - udp monitoring is disabled (i.e. the disabled setting is true)

# Crashes
- eCAL Play GUI replays a measurement
- Multiple eCAL Mon GUI instances are started, either from a console or from the app launcher.

Sometimes, now the eCAL Monitor crashes shortly after being started (after a second or so). The Call stack displayed by Ubuntu points to `eCAL::CMemoryFile::Create` being the issue.
![image](https://user-images.githubusercontent.com/11774314/207374882-0397d8fb-0478-4135-8c12-0c12381d934c.png)

When this issue happens, often for the other eCAL Monitors the next issue is triggered

# Cannot find the eCAL Play Process
- eCAL Play GUI replays a measurement
- Multiple eCAL Mon GUI instances are started, either from a console or from the app launcher.

After several launches (but sometimes even the first launch), either the launched eCAL Monitor or all already started eCAL Monitors fail to find the eCAL Play Process and the topics published by it. They seem to find each other just fine though.

As written, sometimes the newly started eCAL Mon causes the issue for all other already eCAL Mons as well and they recover, once that eCAL Monitor is closed.

hannemn:
Issue has been solved and does not longer occur on Linux systems with activated shm-monitoring. For more detail see PR #917 .

PR #917, opened 2022-12-15
Utilize registration receiver to enrich monitoring information

hannemn:
**Pull request type**

 

 

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes

At the moment there are two seperate receiver entities that handle the incoming traffic for registration information and monitoring. Since we currently do not differentiate internally between registration and monitoring traffic, it is more sensible to harmonize the redundant receiver code part from the software architecture perspective as well as to save os resources. Furthermore, the current architecture can lead to segfaults in linux with the new shm monitoring feature in certain circumstances because the respective mutexes / memory files are opened or locked multiple times within the same process.

Issue Number: #913 

**What is the new behavior?**
The registration receiver forwards the incoming udp registration traffic (as well as shm monitoring traffic) to the monitoring handler. This seems to fix also the crash on linux with activated shm monitoring that spontanously occured before.

- [ ] Yes
- [X] No

**Other information**


rex-schilasky (approved):
reviewed with author

Issue #735, opened 2022-08-01
Win Setup: Cryptographically sign binaries


FlorianReimold:
**Is your feature request related to a problem? Please describe.**
Currently, our binaries are just plain old unsigned binaries. Windows complains a lot when you download and execute the installer. 

**Describe the solution you'd like**
We should start signing our binaries cryptographically. From what I know, self-signing is enough for Windows to complain less. Nevertheless, it is always good practice to sign binaries for integrity.

PR #923, opened 2022-12-19
Feature/sign windows installer on jenkins

OrhanKupusoglu:
**Pull request type**

- [ ] Bugfix
- [ ] Feature
- [X] Build related changes

The Windows installer is not signed automatically by Eclipse Foundation's certificate.

Issue Number: [Win Setup: Cryptographically sign binaries #735](https://github.com/eclipse-ecal/ecal/issues/735)

**What is the new behavior?**

The action contains two newly-introduced environment variables to sign the Windows installer.
The first one sets whether signing on GitHub is allowed or not.
The second one sets whether signing on Jenkins of Eclipse Foundation is allowed or not.
Even IS_GITHUB_SIGNING_ALLOWED is true, but the required secrets of the certificate are missing then this step is skipped.
Similarly, even IS_JENKINS_SIGNING_ALLOWED is true, but the token of the Jenkins job is missing as a secret, then these steps are skipped. 

```
  IS_GITHUB_SIGNING_ALLOWED: true
  IS_JENKINS_SIGNING_ALLOWED: true
```
There was a warning for a deprecated feature used in the Windows-build action:

> **build-windows**
> The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

A minor refactoring removed this warning as well.

The script needs to find first the build number of the last build to wait for the current build.
After triggering a new run on Jenkins, the following job states are expected before before being able to download the binary file: 

- ... pending 
- building ...
- build success 

A log from a sample run illustrates the way this PR modifies the action:

```
-- get last build number
    ++ last build - HTTP status code: 200
    ++ last build#: 54, new build#: 55
-- trigger a new build
    ++ trigger - HTTP status code: 201
-- check build status
    ++ status check - HTTP status code: 200
    ++ ... pending, wait for 5s
    ++ status check - HTTP status code: 200
    ++ ... pending, wait for 5s
    ++ status check - HTTP status code: 200
    ++ ... pending, wait for 5s
    ++ status check - HTTP status code: 200
    ++ ... pending, wait for 5s
    ++ status check - HTTP status code: 200
    ++ ... pending, wait for 5s
    ++ status check - HTTP status code: 200
    ++ building ..., wait for 5s
    ++ status check - HTTP status code: 200
    ++ building ..., wait for 5s
    ++ status check - HTTP status code: 200
    ++ build success
-- download the newly-signed Windows installer
    ++ download - HTTP status code: 200
    ++ downloaded: 'ecal-5.11.0-win64.exe'
``` 

The downloaded binary is then uploaded as **windows-setup-eclipse**.

- [ ] Yes
- [X] No

**Other information**

A new Jenkins job, **gh_auto**, is created for this new behavior. 
To call this job a secret token is required: **secrets.JENKINS_TOKEN_GH_AUTO**
The value of this token can be found in the new job's configuration:
**gh_auto > configure > Build Triggers > Authentication Token**

The newly triggered job needs to be checked on Jenkins to confirm that it is triggered by the GitHub action.


FlorianReimold:
That sounds very promising. I have requested Admin rights from Eclipse, so we can add the required secrets to this repo.

From your description, however one thing is not clear to me, yet. Let's say, I make multiple commits that trigger some GH Action runs, let's call them in the order of starting them:
- `Run 1`
- `Run 2`
- `Run 3`
- `Run 4`
- `Run 5`

They will now run for some time and then finish in another order:

- `Run 5`
- `Run 1`
- `Run 2`
- `Run 3`
- `Run 4`

What will now happen with the artifacts? Will each run still receive its own signed artifact from Jenkins?

OrhanKupusoglu:
This scenario is untested.

But I think that the first line of defense will be GitHub Actions, the runs will be serialized in an order.
Hopefully there won't be any overlapping runs, but they will be called one after other, from "pending" to "running".
Jenkins also will behave the same way.
Another point to consider is that the step which is added by this PR runs only for 60 to 90 seconds in the total duration of 50-70 minutes, this lowers the chance of overlaps.

In this step first the last build number is asked, and right after the reponse a build is triggered with the expectation that the incremented value will be the next build number.
Therefore the few seconds in which these requests run are critical.

FlorianReimold:
I manually pushed the code [to a branch](https://github.com/eclipse-ecal/ecal/tree/feature/sign-windows-installer-on-jenkins), as I wanted to try out the secrets I have added and leave this PR open for discussion and questions at the moment.

There is one crucial thing that I need to understand:

**Is there a theoretical possibility that signed artifacts are pulled by the wrong GH Action?**

I can live with not having a signed binary at all (and thus having to manually start the signing process), but having the wrong binary attached to a build would be a no-go, as I consider finding that issue is be extremely hard.

OrhanKupusoglu:
I am not sure about this possibility (wrong binary, not the new one, is signed).
I think that this is theoretically possible.

It depends on whether two windows builds can be run simultaneously ("parallel") or not ("series").
If another GH action runs as this one is running, then it may be off by one binary upload.
This is the [concurrency ](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency) setting for the action.
Therefore the action must consider this possibility.
I was wrong to assume that parallel-builds are prevented.

FlorianReimold:
OK, then we have to find a way to prevent that from happening. The potential to have builds attached to the wrong GH Action run just isn't acceptable.

Maybe something like the following would work:
- `build-windows.yamo` workflow just compiles and uploads as always
- A `sign-windows.yaml` workflow is triggered. This workflow should now be able to access the artifacts uploaded earlier and get their artifact-id
  - It now triggers a Jenkins run with that ID
- Jenkins downloads, signs and uploads the artifact.
- The sign-windows.yaml workflow downloads the signed binary from the triggered workflow. Identification could e.g. happen by returning the jenkins workflow number when triggering (if possible) or by storing the artifact ID of the unsigned GH Action somewhere (e.g. as filename of the signed artifact on Jenkins)

--------------

The good news is that the secrets are properly set up and we now have our first automatically signed artifact! 👏
https://github.com/eclipse-ecal/ecal/actions/runs/3873438006

FlorianReimold:
Preventing concurrent Windows builds in general is not a good solution, as the runs tend to be very time consuming (>1h). Serializing them would lead to extremely time consuming PRs. For example releasing a bugfix causes at least 4 builds (Push commit + push tag, each on 2 branches). With a sequential build this would take at least 4h to build, even if nobody is working on the repo and doing additional commits and PRs.
But I could see having a succeeding "sign-windows" workflow as a non-concurrent workflow.

OrhanKupusoglu:
Separation of the deployment & signing steps into its own job which needs "build" may be a solution.

OrhanKupusoglu:
The scripts are improved against concurrency issues.
The Jenkins job now receives the run_id of the GitHub action, for example: **3893855766** 
Given the list of artifacts, the **archive_download_url** contained in the array element where **workflow_run.id** matches the given run_id, is then selected.

```
{
  "total_count": 2151,
  "artifacts": [
    {
      "id": 507138233,
      "node_id": "MDg6QXJ0aWZhY3Q1MDcxMzgyMzM=",
      "name": "windows-setup",
      "size_in_bytes": 57339910,
      "url": "https://api.github.com/repos/eclipse-ecal/ecal/actions/artifacts/507138233",
      "archive_download_url": "https://api.github.com/repos/eclipse-ecal/ecal/actions/artifacts/507138233/zip",
      "expired": false,
      "created_at": "2023-01-11T15:30:55Z",
      "updated_at": "2023-01-11T15:30:58Z",
      "expires_at": "2023-04-11T15:30:12Z",
      "workflow_run": {
        "id": 3893855766,
        "repository_id": 189982812,
        "head_repository_id": 472713913,
        "head_branch": "feature/sign-windows-installer-on-jenkins",
        "head_sha": "9894a7d31e254a5c05fa0a79e0b428af718b350c"
      }
    },
...
```
The  build number is now retrieved after the triggering request by checking the returned back **location** header.
Therefore it is an atomic operation.

FlorianReimold:
Currently, the Jenkins jobs complains, as it fails to find a matching artifact for a run_id. I think it can be improved with this:
https://docs.github.com/en/rest/actions/artifacts?apiVersion=2022-11-28#list-workflow-run-artifacts

OrhanKupusoglu:
The fork does not acces to GH SECRETs, therefore the PR builds do not attempt to sign the Windows installer.
The branch on my fork provides its own run_id, but the Jenkins job only lists the artifacts from the main repository.
Therefore the last test runs on Jenkins failed due to run_id match failures, but this was as expected.
If a build is started from a branch of the main repository, it will probably be OK.

OrhanKupusoglu:
It is seen that the artifact list does not contain the still ongoing run_id's artifact.
Even when it is one job or two jobs or even two workflows, the list of artifacts does not contain the run_id  given, until the run ends with failure due to this issue.
Unfortunately further work is required.

.github/workflows/build-windows.yml
```diff
@@ -269,40 +272,219 @@ jobs:
...
+          'GH_FILE_NAME'= $ASSET_NAME
+          'GH_EXECUTABLE'= Get-Item -Path $ASSET_NAME
+        }
+
+        $RESPONSE = Invoke-WebRequest -Uri $JENKINS_TRIGGER_URL -Method POST -Headers $TRIGGER_HEADERS -Form $FORM
+        $RES_CODE = $RESPONSE.StatusCode
+        Write-Output "    ++ trigger build - HTTP status code: $RES_CODE"
+
+        if ($RES_CODE -ge 200 -and $RES_CODE -lt 300) {
+          $QUEUE_URL = $RESPONSE.Headers.location
```
FlorianReimold, line 385: I am curious what that QUEUE_URL will look like. What information is embedded there?

Can in theory the "get the new build number" code retrieve the build number from another sign job started shortly before or after this one?

.github/workflows/build-windows.yml
```diff
@@ -269,40 +272,219 @@ jobs:
...
+      run: |
+        $LS_OUT = ls *.exe
+        $ASSET_NAME = "$($LS_OUT.Name)"
+        $ASSET_NAME_OLD = "$($ASSET_NAME).old"
+        $ASSET_NAME_NEW = "$ASSET_NAME"
+        $JENKINS_JOB = 'gh_auto_file'
+        $JENKINS_BASE_URL = "https://ci.eclipse.org/ecal"
+        $JENKINS_CRUMB_URL = "$JENKINS_BASE_URL/crumbIssuer/api/json"
+        $TOKEN_NAME_NEW = 'eCAL-Token'
+        $JENKINS_TOKEN_URL = "$JENKINS_BASE_URL/user/admin/descriptorByName/jenkins.security.ApiTokenProperty/generateNewToken?token=$Env:JENKINS_API_TOKEN&newTokenName=$TOKEN_NAME_NEW"
```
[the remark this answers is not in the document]
OrhanKupusoglu: You are right, Jenkins token generation was not required.

.github/workflows/build-windows.yml
OrhanKupusoglu: The triggering POST receives in the response a 'location' header with a queue URL value, for example:
`    "location": "https://ci.eclipse.org/ecal/queue/item/2853"`
In a later test run the queue number did change:
`    "lovation": !https://ci.eclipse.org/ecal/queue/item/2871"`

Then by appending '/api/json/' to this URL, the actual build number can be retrieved.
These URLs have all time expired, and now receive 404 responses.
To my best knowledge, the Jenkins API won't allow another use for this build number for this job, it is generated in an atomic transaction.

.github/workflows/build-windows.yml
```diff
@@ -306,3 +309,180 @@ jobs:
...
+        $JENKINS_TRIGGER_URL = "$JENKINS_JOB_URL/buildWithParameters?token=$Env:JENKINS_JOB_TOKEN"
+        $WAIT_FOR_JENKINS_SEC = 5
+        $COUNTER_LIMIT = 20
+        $RESPONSE_CODE = 0
+        $QUEUE_URL = ''
+        $CRUMB_ID = ''
+        $CRUMB_FIELD = ''
+        $CRUMB_COOKIE = ''
+        $COUNTER = 0
+        $QUEUE_URL = ''
```
FlorianReimold, line 335: `QUEUE_URL` is defined here and in line 330 (found by ChatGPT 😉)

FlorianReimold (changes requested):

.github/workflows/build-windows.yml
OrhanKupusoglu: Excellent, good catch by ChatGPT.

PR #941, opened 2023-01-12
Core Services Bugfix: Service response header now contains the methodname again

### LINKED ISSUE #937: Rec Server Service Protobuf Descriptor Issues
Somehow, we got a problem with the descriptor information of the eCAL Rec Server Service and probably other services as well. @adriana-chirila reported that controlling the Rec GUI via RPC in eCAL 5.11 does not work any more.

What I can say is that `ecal_sample_rec_server_service_gui` fails to print a proper status DebugString in 5.11. All the information seem to be there, but the proper de-serialization is somehow broken (?)

Left: eCAL 5.11
Right: eCAL 5.10

image

Screenshots come from Windows

FlorianReimold:
Fixes #937, a bug that was introduced with commit 419d7f33edca6cf4b02c50b8e879981f4043f873

### DIFF
--- a/ecal/core/src/service/ecal_service_server_impl.cpp
+++ b/ecal/core/src/service/ecal_service_server_impl.cpp
@@ -306,6 +306,7 @@ namespace eCAL
     // get method
     SMethod method;
     auto& request_pb_header = request_pb.header();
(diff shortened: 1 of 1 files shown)


KerstinKeller:
Seems I can't approve from within the phone app 🤔

PR #950, opened 2023-01-24
Core: Added proper topic_name -> UDP MC group computation.

KerstinKeller:
**Pull request type**
- [x] Feature

Unintuitive masking for multicast addresses

Issue Number: #873

**What is the new behavior?**
This PR introduces a new multicast configuration version mode, e.g. V2. The previous mode is called V1.
V1 stays the standard mode for now, to be compatible with older eCAL Versions. 
V2 is introduced, and allows better configuration.

Please review:
The behavior to treat collisions:
Per default, `multicast_group` is the MC address to send registration information on.
If a topic is to be sent on the same MC address as registration information, it is send on +1. If this causes "overflow" (e.g. multicast group `139.0.0.255` and mask of `255.255.255.0`, the topic will be sent on `139.0.0.0`.
It is to be discussed if this is a favorable behavior.

The behavior is now somewhat documented in the ini file, it should probably be documented extensively in the documentation.

- [x] No

It's not a breaking change as the default configuration stays as is.

### DIFF
--- a/doc/rst/configuration/options.rst
+++ b/doc/rst/configuration/options.rst
@@ -54,18 +54,38 @@ The network setting drive how and which ...
    
    ``false`` = local host only communication
    
+.. option:: multicast_config_version
+
+   ``v1`` / ``v2``, default: ``v1`` 
+
+   UDP configuration version (Since eCAL 5.12.)
+   
+   ``v1`` = default behavior
+
+   ``v2`` = new behavior, comes with a bit more intuitive handling regarding masking of the groups
+
 .. option:: multicast_group
 
    IPV4 Adress, default ``239.0.0.1``
 
-   UDP multicast group base. 
+   UDP multicast group base. All registration and logging information is sent on this address.
 
 .. option:: multicast_mask
 
+   ``v1`` **behavior:** 
+   
    ``0.0.0.1``-``0.0.0.255``
 
    Mask maximum number of dynamic multicast group
 
+   ``v2`` **behavior:** 
+   
+   ``255.0.0.0``-``255.255.255.255`` 
+   
+   Mask for the multicast group. Topic traffic may be set on any of the unmasked addresses.
+   
+   With ``multicast_group``: ``239.0.0.1`` and ``multicast_mask``: ``255.255.255.255``, topic traffic will be sent on addresses ``239.0.0.0``-``239.0.0.255``.
+   
 .. option:: multicast_port     
    
    ``14000 + x``
--- a/doc/rst/configuration/options.rst
+++ b/doc/rst/configuration/options.rst
@@ -84,7 +84,7 @@ The network setting drive how and which ...
    
    Mask for the multicast group. Topic traffic may be set on any of the unmasked addresses.
    
-   With ``multicast_group``: ``239.0.0.1`` and ``multicast_mask``: ``255.255.255.255``, topic traffic will be sent on addresses ``239.0.0.0``-``239.0.0.255``.
+   With ``multicast_group``: ``239.0.0.1`` and ``multicast_mask``: ``255.255.255.0``, topic traffic will be sent on addresses ``239.0.0.0``-``239.0.0.255``.
    
 .. option:: multicast_port     
    
--- a/ecal/core/include/ecal/ecal_config.h
+++ b/ecal/core/include/ecal/ecal_config.h
@@ -17,6 +17,8 @@
  * ========================= eCAL LICENSE =================================
 */
 
+#pragma once
+
 #include <ecal/ecal_os.h>
 #include <ecal/ecal_tlayer.h>
 #include <ecal/ecal_log_level.h>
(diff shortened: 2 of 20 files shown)


ecal/core/include/ecal/ecal_config.h
```diff
@@ -27,6 +29,12 @@ namespace eCAL
 {
   namespace Config
   {
+    enum class UDPConfigVersion
```
hannemn, line 32: Naming this enum to` UdpMulticastConfigVersion` would better match to the according getter function `GetUdpMulticastConfigVersion`.

ecal/core/src/io/udp_configurations.cpp
```diff
@@ -0,0 +1,36 @@
...
+    return localhost_udp_address;
+  }
+  else
+  {
+    // both in v1 and v2, the mulicast group is returned as the adress for the registration layer
+    return Config::GetUdpMulticastGroup();
+  }
+}
+
+std::string eCAL::UDP::GetLoggingMulticastAdress()
```
hannemn, line 23: Typo Ad**d**ress

ecal/core/src/io/udp_configurations.cpp
```diff
@@ -0,0 +1,36 @@
+#include "io/udp_configurations.h"
+
+#include <ecal/ecal_config.h>
+#include "topic2mcast.h"
+
+const std::string localhost_udp_address{ "127.255.255.255" };
+
+
+std::string eCAL::UDP::GetRegistrationMulticastAdress()
```
hannemn, line 9: Typo Ad**d**ress

ecal/core/src/io/udp_configurations.cpp
```diff
@@ -0,0 +1,36 @@
...
+}
+
+std::string eCAL::UDP::GetLoggingMulticastAdress()
+{
+  //TODO: At the moment, both logging and monitoring addresses seem to be the same
+  // Should it be kept or changed?
+  return GetRegistrationMulticastAdress();
+}
+
+std::string eCAL::UDP::GetTopicMulticastAdress(const std::string& topic_name)
```
hannemn, line 30: Typo Ad**d**ress

hannemn (changes requested):

ecal/core/src/io/udp_configurations.cpp
KerstinKeller: should I also rename the namespace `UDP` -> `Udp` ?

FlorianReimold: Would that change some public API?

If not, I usually prefer CamelCase where abbreviations are treated as words. So yes, I would have made it `Udp`. If incorporated in longer variable names, this makes it easier to distinguish between the components (e.g. `UDPSocket` vs `UdpSocket`, or `TCPIPConnection` vs `TcpIpConnection`). But in the end it doesn't really matter.

ecal/core/include/ecal/ecal_config.h
FlorianReimold: I agree

ecal/core/src/io/udp_configurations.cpp
KerstinKeller: No, no public API, I just introduced some namespacing.
Here we separate by `::` so it's not so much of an issue. but I mean `UDP` it's an abbreviation...
I'll just keep it. (we also have namespace `eCAL` not `Ecal`).

FlorianReimold (changes requested):

doc/rst/configuration/options.rst
```diff
@@ -54,18 +54,38 @@ The network setting drive how and which ...
...
 
+   ``v1`` **behavior:** 
+   
    ``0.0.0.1``-``0.0.0.255``
 
    Mask maximum number of dynamic multicast group
 
+   ``v2`` **behavior:** 
+   
+   ``255.0.0.0``-``255.255.255.255``
```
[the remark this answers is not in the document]
KerstinKeller: we have no verifaction of any inputs whatsoever in the ini files. neither for this entry nor for any of the others.
I would tackle this separately.

doc/rst/configuration/options.rst
```diff
@@ -54,18 +54,38 @@ The network setting drive how and which ...
...
 
    Mask maximum number of dynamic multicast group
 
+   ``v2`` **behavior:** 
+   
+   ``255.0.0.0``-``255.255.255.255`` 
+   
+   Mask for the multicast group. Topic traffic may be set on any of the unmasked addresses.
+   
+   With ``multicast_group``: ``239.0.0.1`` and ``multicast_mask``: ``255.255.255.255``, topic traffic will be sent on addresses ``239.0.0.0``-``239.0.0.255``.
```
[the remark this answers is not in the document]
KerstinKeller: you are correct

ecal/core/include/ecal/ecal_config.h
```diff
@@ -17,6 +17,8 @@
  * ========================= eCAL LICENSE =================================
 */
 
+#pragma once
```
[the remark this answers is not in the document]
KerstinKeller: yes, it was actually missing! An I got multiple definition erros due to defining a class enum in the header.
Apparently it's not a problem for function definitions.

FlorianReimold (approved):

PR #957, opened 2023-01-31
eCAL Rec: Added RPC Service "SetConfig"

### LINKED ISSUE #848: eCAL Rec GUI: Add configurable config of the RecServerService via ServiceClient
I want to change the config of the recorder gui from within a module so a [remote started recording like shown here](https://github.com/eclipse-ecal/ecal/blob/5345a5c0a111b23f3150a28834ccb57f55924490/samples/cpp/services/rec_server_service_gui/src/rec_server_service_gui.cpp#L113) has a specific pre-buffer that differs from the configured one in the config of the recorder itself. Best case would be that the recorder still runs with the old pre buffer but the recording only uses the for this started recording used time.

**Describe alternatives you've considered**
Alternatively, I could read the config at initialization and reset to that after each of my triggered recordings is done to the default.

FlorianReimold:
- Added SetConfig RPC Call to Rec GUI and Rec CLI
- Enabled --set-config from command line and setconfig from interactive mode of Rec CLI in Remote control mode
- Added Tests for Rec CLI Remote control mode (command-line version) testing the control of both eCAL Rec CLI and GUI 

Fixes #848

### DIFF
--- a/app/rec/rec_client_core/include/rec_client_core/rec_error.h
+++ b/app/rec/rec_client_core/include/rec_client_core/rec_error.h
@@ -75,16 +75,6 @@ namespace eCAL
       Error(ErrorCode error_code, const std::string& message) : error_code_(error_code), message_(message) {}
       Error(ErrorCode error_code) : error_code_(error_code) {}
 
-      // Copy
-      Error(const Error&)            = default;
-      Error& operator=(const Error&) = default;
-
-      // Move
-      Error& operator=(Error&&)      = default;
-      Error(Error&&)                 = default;
-
-      ~Error() {}
-
     //////////////////////////////////////////
     // Public API
     //////////////////////////////////////////
--- a/app/rec/rec_client_core/include/rec_client_core/rec_error.h
+++ b/app/rec/rec_client_core/include/rec_client_core/rec_error.h
@@ -74,7 +74,14 @@ namespace eCAL
     public:
       Error(ErrorCode error_code, const std::string& message) : error_code_(error_code), message_(message) {}
       Error(ErrorCode error_code) : error_code_(error_code) {}
-      Error(const Error& error) : error_code_(error.error_code_), message_(error.message_) {}
+
+      // Copy
+      Error(const Error&)            = default;
+      Error& operator=(const Error&) = default;
+
+      // Move
+      Error& operator=(Error&&)      = default;
+      Error(Error&&)                 = default;
 
       ~Error() {}
 
@@ -139,14 +146,9 @@ namespace eCAL
     //////////////////////////////////////////
       inline operator bool() const { return error_code_ != ErrorCode::OK; }
       inline bool operator== (const Error& other) const { return error_code_ == other.error_code_; }
+      inline bool operator== (const ErrorCode other) const { return error_code_ == other; }
       inline bool operator!= (const Error& other) const { return error_code_ != other.error_code_; }
-
-      Error& operator=(const Error& other)
-      {
-        error_code_ = other.error_code_;
-        message_    = other.message_;
-        return *this;
-      }
+      inline bool operator!= (const ErrorCode other) const { return error_code_ != other; }
 
       Error& operator=(ErrorCode error_code)
       {
--- a/app/rec/rec_gui/src/qecalrec.cpp
+++ b/app/rec/rec_gui/src/qecalrec.cpp
@@ -842,7 +842,7 @@ eCAL::rec_server::RecServerConfig QEcalRec::config() const
 
 eCAL::rec::Error QEcalRec::setConfig(const eCAL::rec_server::RecServerConfig& config, bool omit_dialogs)
 {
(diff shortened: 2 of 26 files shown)


app/rec/rec_client_core/include/rec_client_core/rec_error.h
```diff
@@ -74,7 +74,14 @@ namespace eCAL
     public:
       Error(ErrorCode error_code, const std::string& message) : error_code_(error_code), message_(message) {}
       Error(ErrorCode error_code) : error_code_(error_code) {}
-      Error(const Error& error) : error_code_(error.error_code_), message_(error.message_) {}
+
+      // Copy
```
hannemn, line 78: Since this class uses the default move and copy constructor as well as the default destructor anyways, there is no need to define them explicitly.

app/rec/rec_gui/src/qecalrec.cpp
```diff
@@ -840,6 +840,27 @@ eCAL::rec_server::RecServerConfig QEcalRec::config() const
   return rec_server_->GetConfig();
 }
 
+eCAL::rec::Error QEcalRec::setConfig(const eCAL::rec_server::RecServerConfig& config, bool omit_dialogs)
+{
+  auto error = rec_server_->SetConfig(config);
```
hannemn, line 845:
```suggestion
  const auto error = rec_server_->SetConfig(config);
```

app/rec/rec_client_core/include/rec_client_core/rec_error.h
FlorianReimold: Are your sure? I thought I had to define those, as I have a non-standard Constructor as well as non-standard operator= overloads

app/rec/rec_server_core/src/config/config.cpp
```diff
@@ -132,12 +109,12 @@ namespace eCAL
           if (config_elements.find(v) != config_elements.end())
           {
             // Load config
-            bool success = applyConfig(config_v2to4::readConfig(config_elements[v]), rec_server);
-            if (success && (version != nullptr))
+            auto error = rec_server.SetConfig(config_v2to4::readConfig(config_elements[v]));
```
hannemn, line 112: Countercheck if error variable can be used with const qualifier.

app/rec/rec_tests/rec_rpc_tests/src/external_ecal_rec.h
```diff
@@ -0,0 +1,73 @@
...
+#pragma warning(pop)
+#endif
+
+#include <string>
+#include <mutex>
+#include <rec_server_core/rec_server_config.h>
+#include <rec_client_core/rec_error.h>
+
+// Lock preventing multiple instances of this class
+static std::mutex ecal_rec_cli_instance_lock;
```
hannemn, line 37: Is the declarion of the mutex outside the `ExternalEcalRecInstance` namespace by purpose? There is also the possibility to define static member variables within the class.

app/rec/rec_client_core/include/rec_client_core/rec_error.h
hannemn: It's only in case of defining a custom destructor - then it's necessary to define the others as well. Rule of five from cpp reference: _"If a class requires a user-defined destructor, a user-defined copy constructor, or a user-defined copy assignment operator, it almost certainly requires all three. [...]"_

app/rec/rec_client_core/include/rec_client_core/rec_error.h
```diff
@@ -139,14 +146,9 @@ namespace eCAL
...
+      inline bool operator== (const ErrorCode other) const { return error_code_ == other; }
       inline bool operator!= (const Error& other) const { return error_code_ != other.error_code_; }
-
-      Error& operator=(const Error& other)
-      {
-        error_code_ = other.error_code_;
-        message_    = other.message_;
-        return *this;
-      }
+      inline bool operator!= (const ErrorCode other) const { return error_code_ != other; }
```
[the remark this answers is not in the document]
FlorianReimold: Yes, this one is intentional. The message is just some additional message going to a console. When actually comparing error codes I want to know e.g. if it is a GENERIC_ERROR, or OK, or CURRENTLY_RECORDING or something like that.

app/rec/rec_tests/rec_rpc_tests/src/external_ecal_rec.h
FlorianReimold: There is no C++ `namespace`, if I recall correctly. This is supposed to be global, so I put it there intentionally. I guess I could have put it inside the class as a static variable as well, but I didn't care, as this is just a helper class for my GTests.

FlorianReimold:
I fixed most of your comments and had clang-tidy run over all eCAL Rec Projects. There were like thousands of warnings (shame on me), but I only fixed those that were in scope of this PR.
I will probably come back later in another PR and work on the rest.

PR #995, opened 2023-03-03
std::map replaced by CExpMap in description gate

rex-schilasky:
**Pull request type**

- [X] Bugfix

Internal maps in description gate never expire.

**What is the new behavior?**
std::map replaced by CExpMap with expiration "monitoring timeout" (default = 5s).

### DIFF
--- a/ecal/core/src/ecal_descgate.cpp
+++ b/ecal/core/src/ecal_descgate.cpp
@@ -22,6 +22,8 @@
 **/
 
 #include <ecal/ecal_log.h>
+#include <ecal/ecal_config.h>
+
 #include "ecal_descgate.h"
 #include <assert.h>
 #include <algorithm>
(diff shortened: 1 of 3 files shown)


KerstinKeller (commented):
Looks good to me. 
My only question is why you are moving from a `mutable std::shared_timed_mutex` to a `std::mutex`.

And second question: will this be part of the new monitoring interface? Because then we will have to make the `ExpMap` public, e.g. move it to the public header files of eCAL.

If so, I am not 100% sure if an expired map is what users will expect from getting a monitoring info. They will expect a snapshot at a certain time.

rex-schilasky:
> Looks good to me. My only question is why you are moving from a `mutable std::shared_timed_mutex` to a `std::mutex`.
> 
> And second question: will this be part of the new monitoring interface? Because then we will have to make the `ExpMap` public, e.g. move it to the public header files of eCAL.
> 
> If so, I am not 100% sure if an expired map is what users will expect from getting a monitoring info. They will expect a snapshot at a certain time.

Thank you for the first review. The mutex can be changed back to a shared one .. this was replaced by a kind of copy and past from the monitoring CExpMap usage (sorry for that :-)).

The CExpMap will not be part of the public interface. The user API will get a snapshot of all entities in a std::map.

hannemn (approved):

PR #1002, opened 2023-03-08
Allow standalone build of CSharp bindings.

KerstinKeller:
With this commit, the C# project can be build as a standalone project, against an eCAL SDK (tested against internal conan build)
Also, .proto files are compiled to C# files, e.g. they are not committed with the source code.


lang/csharp/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+# C# build, can be executed standalone against an eCAL installation
+project(ecal_csharp)
+
+set(CMAKE_DOTNET_SDK "")
+set(ECAL_CSHARP_PROTOBUF_VERSION "3.11.4" CACHE STRING "Protobuf Version for C# build")
```
FlorianReimold, line 5: Is there a way to auto-detect this version? We have multiple hardcoded protobuf versions at this point and it will be a pain to change that.

I don't know whether this an be applied here, but google has a find-script for gRPC that compiles a small test application for determining the library version.
https://github.com/googleapis/google-cloud-cpp/blob/927cbee11233e106afc00394f1809ac805a3c646/cmake/FindgRPC.cmake#L275

samples/csharp/person/protobuf/protobuf_csharp_helper.cmake
```diff
@@ -0,0 +1,101 @@
...
+# 
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# ========================= eCAL LICENSE =================================
+
+function(PROTOBUF_GENERATE_CSHARP SRCS_RET PROTO_OUT_DIR_RET PROTO_ROOT)
```
FlorianReimold, line 19: Maybe we should replace this with the original [protobuf-generate.cmake](https://github.com/protocolbuffers/protobuf/blob/main/cmake/protobuf-generate.cmake)?

Just saying, i am fine with the custom C#-only solution, but the generic official protobuf_generate script also is amazing 😊

FlorianReimold (commented):
Looks good to me. Only wish that I would have is to auto-detect the protobuf version, if possible. 

If that's not feasible, we leave it as is and I will approve the PR.

lang/csharp/CMakeLists.txt
KerstinKeller: Yes I am also not 100% happy about it. Autodetect would be better.
And the good thing here, it's not hardcoded but an option. I compiled against two different protobuf versions in parallel by setting the option.
I am also not happy you need a "C++"  `find_package` call for a C# project to detect protoc.
So it would be more generic to have protoc somehow also as a nuget package / dependency and then take that somehow.
But this was a first compromise.
If possible I'd like to merge this as-is, and then maybe we can come back to improve it?

samples/csharp/person/protobuf/protobuf_csharp_helper.cmake
KerstinKeller: Yeah I know, I thought about it too after you showed that sample, but now it's already working 😄 maybe for a next PR?

FlorianReimold (approved):

Issue #988, opened 2023-02-28
Shared memory not being created on macOS


JerryHyun:
### Problem Description

I am considering using eCAL for real-time synchronization of large-sized 3d mesh data. When I tested it on windows os using minimal_rec and minimal_snd samples, there was no problem. However, the test did not work on macOS, so I tried to find the cause of the problem.

1. macOS is very limited in the filename of shared memory. The maximum length for a filename currently is 31 characters per section (see PSHMNAMELENGTH in /usr/include/sys/posix_shm.h)
For this reason, the minimal_rec and minimal_snd samples are also problematic because the longest shared memory name created when running the minimal_rec sample is 38 characters.(/Hello_1058310900281_shm_38316_ack_evt)
And the default shutdown process memory name is also a problem because it is over 31 characters long.(/ecal_shutdown_process_38316_evt)
The solution is to use a random 4-character UNIQUE ID instead of the 13-character EPOCH TIME used to create the shared memory name to avoid duplicating the shared memory (if the reason for the EPOCH TIME is to avoid duplicates...).
**This change allows you to use topic names of up to 7 characters.**
Also, overriding EVENT_SHUTDOWN_PROC to "ecal_shutdown_proc" solves the shutdown process memory name issue.

2.  When setting g_process_name, if I give a length like std::string(buf, length); I get a problem on clang 13.0.0 x86_64-apple-darwin20.6.0.
Please see the code snippet below.
```
//strTest.cpp
#include <iostream>
#include <string.h>

int main()
{
	uint32_t length = 25;
	char buf[length];
	char origin[] = "/usr/temp";

	strncpy_s(buf, sizeof(buf), origin, sizeof(origin));
	std::string bufStr = std::string(buf, length);

	std::cout << bufStr << std::endl;

	return 0;
}
```
After building clang 13.0.0 x86_64-apple-darwin20.6.0, the results of the run are shown below.
```
jerry@Jerry-MacBookAir ~ %  ./strTest
/usr/temp�?l`QG�
jerry@Jerry-MacBookAir ~ % 
```
As you can see above, a strange value is added to the string.
The problem can be easily fixed as shown below.
`std::string(buf);`

3. There is an issue with not finding the ecaltime-localtime dynamic lib.
This can be easily fixed with a small modification to ecal_timegate.cpp.

Let's create a pull request to fix the above issue I found.

### How to reproduce

.

### How did you get eCAL?

Custom Build / Built from source

### Environment

[...]

PR #1004, opened 2023-03-09
Bugfix/shared memory not being created on mac

JerryHyun:
**Pull request type**

 

 

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

Issue Number: #988 

**What is the new behavior?**

To resolve the issue of limiting the number of characters in shared memory names in macOS, I changed it to use the hash name when creating shared memory.
At first, I tried to apply it only to macOS, but I changed it to use hashed names in all OSs for consistency.
After this modification, it is expected that it will not be compatible with the previous version regarding shared memory.

- [ ] Yes
- [ ] No

**Other information**

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -55,3 +55,6 @@
 [submodule "thirdparty/udpcap"]
 	path = thirdparty/udpcap
 	url = https://github.com/eclipse-ecal/udpcap.git
+[submodule "thirdparty/xxhash"]
+	path = thirdparty/xxhash
+	url = https://github.com/stbrumme/xxhash.git
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -94,6 +94,11 @@ This project has dependencies - list with dependency name and license follows.
   - URL: http://www.legroom.net/software
   - License: LGPL 3
 
+2.3.7 xxHASH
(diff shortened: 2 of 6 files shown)


KerstinKeller:
Thanks for splitting the PR! It looks really clean and good!

The good thing is, that this fix will not really break compatibility, as the publisher can freely choose memory file names and communicate those to the subscribers via the monitoring layer.
So basically we are free to choose filenames to our liking.

As for the actual names:
I would like to discuss this with the team, but we haven't gotten around to it.
I would suggest:
- memory file names should still be prefixed with `ecal_` (We've had some cleanup issues. This allows the user clearly to identify all files created by ecal). But if everyone else is fine with giving them proper random names, I'm good, too.
-- I would probably not do the hasing based on the communicated event_name, but create a function creates the event names (already hashed) based on topics name etc. I think the way it is implemented right now, compatibility is actually broken.

I will take this PR into our meeting and also make a proposal how to implement it.
Thanks so much for your contribution!

KerstinKeller:
I am closing this PR. Please see #1009 and it would be great if you can provide feedback whether this works for you on MacOS.

PR #1009, opened 2023-03-10
Feature/size limited memfile names

KerstinKeller:
**Pull request type**

With this PR, ecal memory files / events will be limited to max 29 characters

`ecal_<topic&time_hash>` as a base name, and then possibly `_ack` as well as process specific events for synchronization.

```
ecal_        5 char
<hash>       8 char
_            1 char
<PID>    max 7 char
_ack         4 char
_evt         4 char
```

This totally works and communication works fine, there is no kind of incompatibility.

However, we should reiterate on what really makes sense here.
After all we want to avoid collisions. To avoid collisions, all we really need for the filename would be the creation time, it's not even necessary to include the topic name.

Also if we think about possibly taking out the shared memory as a separate library, we'd probably want to keep being able to set a file name, and not have the library compute some random hashes.

### DIFF
--- a/ecal/core/src/io/ecal_memfile_naming.cpp
+++ b/ecal/core/src/io/ecal_memfile_naming.cpp
@@ -9,19 +9,14 @@ namespace eCAL
   {
     std::string BuildMemFileName(const std::string& base_name_, const std::chrono::time_point<std::chrono::steady_clock>& time_point)
     {
-      std::string mfile_name(base_name_);
       std::stringstream out;
-      out << mfile_name << "_" << std::chrono::duration_cast<std::chrono::microseconds>(time_point.time_since_epoch()).count();
-      mfile_name = out.str();
 
-      // replace all '\\' and '/' to '_'
-      std::replace(mfile_name.begin(), mfile_name.end(), '\\', '_');
-      std::replace(mfile_name.begin(), mfile_name.end(), '/', '_');
+      size_t hash = std::hash<std::string>()(base_name_ + std::to_string(time_point.time_since_epoch().count()));
+      uint32_t compressed_hash = static_cast<uint32_t>(hash);
 
-      // append "_shm" for debugging purposes
-      mfile_name += "_shm";
+      out << "ecal_" << std::hex << compressed_hash;
 
-      return mfile_name;
+      return out.str();
     }
   }
 }
--- a/ecal/core/src/io/ecal_memfile_naming.cpp
+++ b/ecal/core/src/io/ecal_memfile_naming.cpp
@@ -1,20 +1,41 @@
+/* ========================= eCAL LICENSE =================================
+ *
+ * Copyright (C) 2016 - 2019 Continental Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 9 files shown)


KerstinKeller:
@JerryHyun  FYI

FlorianReimold:
@KerstinKeller: Very good overview of the names along with the used-up characters, but you forgot the `_evt`, which takes up 4 more characters and gets us to 29 out of 31 in total :)

ecal/core/src/io/ecal_memfile_naming.cpp
```diff
@@ -0,0 +1,22 @@
+#include <io/ecal_memfile_naming.h>
```
KerstinKeller, line 1: License header missing!!!!!

KerstinKeller (commented):
`CSyncMemoryFile` should keep base_name, which can be `ecal_` and passed from outside, which be used in `BuildRandomMemFileName`.

ecal/core/src/io/ecal_memfile_naming.cpp
```diff
@@ -0,0 +1,22 @@
...
+
+namespace eCAL
+{
+  namespace memfile
+  {
+
+    std::string BuildRandomMemFileName()
+    {
+      static std::random_device rd;
+      static std::uniform_int_distribution<uint32_t> dist(0, UINT32_MAX);
```
FlorianReimold, line 14:
```suggestion
      static std::uniform_int_distribution<uint32_t> dist(0, std::numeric_limits<uint32_t>::max());
```

ecal/core/src/io/ecal_memfile_naming.cpp
```diff
@@ -0,0 +1,22 @@
+#include <io/ecal_memfile_naming.h>
+
+#include <sstream>
+#include <random>
```
FlorianReimold, line 4:
```suggestion
#include <random>
#include <limits>
```

FlorianReimold (changes requested):
Looks good, besides the license header and the numeric_limits change 😉

ecal/core/src/io/ecal_memfile_naming.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include <sstream>
+
+namespace eCAL
+{
+  namespace memfile
+  {
+
+    std::string BuildRandomMemFileName(const std::string& base_name)
+    {
+      static std::random_device rd;
```
hannemn, line 34: Doesn't the random device requires a seed value?

KerstinKeller: No, that's the point of the random device, that it is truly random. It's somehow connected to the hardware.

hannemn: I see, but it seems to be at least implementation specific. Cpp reference remarks: "[...]std::random_device may be implemented in terms of an implementation-defined pseudo-random number engine if a non-deterministic source (e.g. a hardware device) is not available to the implementation. In this case each std::random_device object may generate the same number sequence. [...]"

PR #1013, opened 2023-03-14
C#: Overload current API functions to take/return byte[] instead of string

KerstinKeller:
C# strings are not pure byte containers as they are in C++ and thus not very suitable for the API in many places.
This commit provides some overloads / additional functions to use `byte[]` instead of `string`.


FlorianReimold (approved):
I am approving this without actually having any C# competence 😶

Issue #1021, opened 2023-03-19
 When big msg is transmitted, it will lead to high memory usage of the legacy STL.


KanonWY:
### How Found
When use the example (datarate_rec/datarate_snd) , I found some question(maybe). 
For datarate_snd, I set the msg' length to 200Mib and  remove the sleep function in the loop. 
For datarate_rec, I alse remove the sleep function in the loop. Then I run them. Of course, the above operation will cause extremely high load. After a few seconds or so, we'll see memory spike on the receiving side, and then stop the sending side(otherwise the receiver would OOM).
### observed situation
After performing the above operations, the receiver program will consume a lot of memory, and it will always occupy it if it is not stopped.
While running the program, I noticed some logs appearing on both ends.
```bash
Could not lock memory file mutex: topic_107489272315_shm.
```
After read source code,  this is due to the slow response, re-creating a new shared memory file. 
I'm wondering about the high memory usage on the receiving end, because I understand from the [official documentation](https://eclipse-ecal.github.io/ecal/advanced/message_drops.html) that dropping messages on the receiving end is allowed. But why does the receiving end have such a high memory consumption (even being OOM).[Even if my example code is not normal this should not happen]
### What I Try
I tried linking to the executable using `tcmalloc` and then using the pprof tool to see memory allocations.I found out that the internal high memory consumption is in a STL object(in the code block below) which accounts for almost 99%.
```c++
class CMemFileObserver
{
      std::vector<char> m_ecal_buffer;
};
```
Since I transferred a large amount of data, the capacity of this data structure is 200M(but But the receiving end left over G memory). So I continued to dig into the source code, I noticed that each shared memory corresponds to a `CMemFileObserver` which is managed by `CMemFileThreadPool`. This means that the creation of a new shared memory file will be tracked by a new CMemFileObserver object.  Since we recreated many of the shared memory, Although each shared memory file is released, the expired `CMemFileObserver` object has not been recycled. I noticed that `CMemFileThreadPool::CleanupPool` is never called (There was a call that was removed two years ago.)
```cpp
  bool CMemFileThreadPool::ObserveFile(const std::string& memfile_name_, const std::string& memfile_event_, const std::string& topic_name_, const std::string& topic_id_, int timeout_observation_ms, const MemFileDataCallbackT& callback_)
  {
    if(!m_created)            return(false);
    if(memfile_name_.empty()) return(false);

    // remove outdated observers
    //CleanupPool();
    ......
}
```
[...]

PR #1024, opened 2023-03-20
memfile observer pool and receive buffer cleanup

rex-schilasky:
**Pull request type**

- [X] Bugfix

* For every observed memory file a separate thread is checking for updates (`CMemFileObserver::Observer`). When a memory file is outdated the observing thread ends without deallocating the `CMemFileObserver` internal receive buffer buffer variable `m_ecal_buffer` (renamed to `receive_buffer`).
* All `CMemFileObserver` are organized in a central map inside `CMemFileThreadPool`. Outdated/stopped observer entities are not removed from the global `CMemFileThreadPool::m_observer_pool` map.

Issue Number: #1021

**What is the new behavior?**
* The memory file content receive buffer is now defined inside the scope of the `CMemFileObserver::Observe` function and will be freed when observation is stopped/outdated.
* A new, additional eCAL global thread is cleaning up the `CMemFileThreadPool::m_observer_pool` map periodically.

- [ ] Yes
- [X] No

### DIFF
--- a/ecal/core/src/ecal_def.h
+++ b/ecal/core/src/ecal_def.h
@@ -126,7 +126,7 @@
 
 /* timeout for create / open a memory file using mutex lock in ms */
 #define PUB_MEMFILE_CREATE_TO                        200
-#define PUB_MEMFILE_OPEN_TO                           50
+#define PUB_MEMFILE_OPEN_TO                          200
 
 /* timeout for memory read acknowledge signal from data reader in ms */
 #define PUB_MEMFILE_ACK_TO                            0   /* ms */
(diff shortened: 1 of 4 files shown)


KerstinKeller (commented):
It sounds reasonable to make the `receive_buffer` a local variable to the Observe function. 👍 

However, it sounds a tiny bit overkill to have a whole thread dedicated to cleaning up the pool. Especially it seems to me that trying to clean up the pool every 100ms seems a bit obsessive, wouldn't every couple of seconds be enough?

ecal/core/src/ecal_def.h
```diff
@@ -126,7 +126,7 @@
 
 /* timeout for create / open a memory file using mutex lock in ms */
 #define PUB_MEMFILE_CREATE_TO                        200
-#define PUB_MEMFILE_OPEN_TO                           50
+#define PUB_MEMFILE_OPEN_TO                          200
```
hannemn, line 129: I guess we need some new mechanism here to prevent such kind of magic numbers in the future ;-)

hannemn (approved):

Issue #1025, opened 2023-03-21
publisher send synchronized method not implemented for CMsgPublisher


rex-schilasky:
### Problem Description

The new `CPublisher::SendSynchronized` send method is not available for the message publisher API (`CMsgPublisher`).

### How to reproduce

Just try to call the method :-).

### How did you get eCAL?

Custom Build / Built from source

### Environment

eCAL 5.x all versions and platforms

PR #1026, opened 2023-03-21
CMsgPublisher::SendSynchronized method added

rex-schilasky:
**Pull request type**

 

 

- [X] Feature

Issue Number: #1025

### DIFF
--- a/ecal/core/include/ecal/msg/publisher.h
+++ b/ecal/core/include/ecal/msg/publisher.h
@@ -143,7 +143,7 @@ namespace eCAL
       else
       {
         // send a zero payload length message to trigger the subscriber side
-        return(CPublisher::Send(nullptr, 0));
+        return(CPublisher::Send(nullptr, 0, time_));
       }
       return(0);
     }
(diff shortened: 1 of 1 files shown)


ecal/core/include/ecal/msg/publisher.h
```diff
@@ -148,6 +148,43 @@ namespace eCAL
...
+        }
+      }
+      else
+      {
+        // see CMsgPublisher::Send
+        return(CPublisher::Send(nullptr, 0));
+      }
+      return(0);
+    }
+
```
KerstinKeller, line 187: Hmhm, this implementation looks a bit strange to me, but I guess the regular implementation is similar.
However, if the `size = 0`, it is not sending synchronized, but shouldn't it?  Even if no data was transferred, the receiver should still acknowledge that it has been signaled?

KerstinKeller (commented):
See comment.

KerstinKeller (approved):
Looks good now.

PR #1029, opened 2023-03-27
Hotfix/proper socket initialization

rex-schilasky:
**Pull request type**

- [X] Bugfix
- [X] Enhancement
- [X] Other (please describe): 
  -  improved error logging

A service initialization may fail silent. Besides this a service may be registered with an uninitialized port number.

**What is the new behavior?**
Extra error logging added. Service checks port number before registration starts.

- [X] No

### DIFF
--- a/ecal/core/src/service/asio_server.h
+++ b/ecal/core/src/service/asio_server.h
@@ -183,6 +183,7 @@ public:
     acceptor_(io_service, asio::ip::tcp::endpoint(asio::ip::tcp::v4(), port)),
     connect_cnt_(0)
   {
+    acceptor_.listen();
     start_accept();
   }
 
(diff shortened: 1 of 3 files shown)


app/rec/rec_tests/rec_rpc_tests/src/external_ecal_rec.cpp
```diff
@@ -64,7 +64,12 @@ ExternalEcalRecInstance::ExternalEcalRecInstance(bool gui)
     return;
   }
 
-  std::this_thread::sleep_for(std::chrono::milliseconds(3000));
+  // wait up to 5 seconds for server startup and connection
```
FlorianReimold, line 67: This code could probably also fix the clientserver_test!

rex-schilasky: That was my idea. I included all client_server tests again and want to get them to run on GH action.

PR #1035, opened 2023-03-29
restart test of new tcp client implementation

rex-schilasky:
**Pull request type**

- [X] Refactoring (no functional changes, no api changes)

**What is the new behavior?**
Improved logging of the tcp service creation.

### DIFF
--- a/ecal/core/src/service/ecal_tcpclient.cpp
+++ b/ecal/core/src/service/ecal_tcpclient.cpp
@@ -57,43 +57,86 @@ namespace eCAL
     m_socket     = std::make_shared<asio::ip::tcp::socket>(*m_io_service);
     asio::ip::tcp::resolver resolver(*m_io_service);
 
-    try
-    {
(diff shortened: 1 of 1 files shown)


ecal/core/src/service/ecal_tcpclient.cpp
```diff
@@ -59,39 +59,41 @@ namespace eCAL
...
       {
         m_event_callback(client_event_connected, "CTcpClient connected");
       }
     }
-    catch (std::exception& /*e*/)
+    catch (std::exception& e)
     {
-      //std::cerr << "CTcpClient::Connect conect exception: " << e.what() << "\n";
+      // log error message
+      std::cerr << "CTcpClient::Connect exception: " << e.what() << "\n";
```
KerstinKeller, line 91: why don't you use ecal logging here instead of cerr?

rex-schilasky:
GH actions did not work properly

PR #1044, opened 2023-04-04
feature/true-zero-copy

rex-schilasky:
**Pull request type**

- [X] Feature

**What is the new behavior?**
The new approach applies true zero copy if a publisher is connected to local subscribers only. The new approach enables the modification of the memory file content without any additional copy action (zero copy).

This pictures shows two applications `ecal_sample_performance_snd` and `ecal_sample_performance_rec_cb` running on a Windows host and exchanging/modifying 8MB payloads in less then 20 µs !

image

- [X] Yes

**Sample Usage**

```cpp
// binary payload (std::vector<char>)
class CVectorPayload : public eCAL::payload
{
public:
  CVectorPayload(size_t size_)
  {
    payload.resize(size_);
    std::fill(payload.begin(), payload.end(), (char)(42));
    std::cout << "Message size  =  " << int(payload.size()) << " Byte = " << int(payload.size() / 1024) << " kByte = " << int(payload.size() / 1024 / 1024) << " MByte" << std::endl;
  }

  void write_complete(void* buf_, size_t len_)
  {
    assert(len_ == size());

    // write complete content to shared memory
    memcpy(buf_, payload.data(), payload.size());
  };

  void write_partial(void* buf_, size_t len_)
  {
    size_t write_idx((clock % 1024) % len_);
    char   write_chr(clock % 10 + 48);

    // write partial content to shared memory
    static_cast<char*>(buf_)[write_idx] = write_chr;

    clock++;
  };

  const void* data() { return payload.data(); };
  size_t      size() { return payload.size(); };

private:
  std::vector<char> payload;
  int               clock = 0;
};

// main entry
int main(int argc, char **argv)
{
  // initialize eCAL API
  eCAL::Initialize(argc, argv, "performance_snd");

  // create payload
  CVectorPayload binary_payload(payload_size);

  // create publisher
  eCAL::CPublisher pub("zero-copy-pub");
 
  // enable zero copy mode
  pub.ShmEnableZeroCopy(true);

  // activate handshake mode with 50 ms
  pub.ShmSetAcknowledgeTimeout(50 /*ms*/);

  // send updates
  while(eCAL::Ok())
  {
    // send payload
    pub.Send(binary_payload);
  }

  // destroy publisher
  pub.Destroy();

  // finalize eCAL API
  eCAL::Finalize();

  return(0);
}
```


KerstinKeller (commented):
One suggesting: it looks like you're replicating a lot of code, depending if you have to apply now, or send the traditional way.
Do you think it might make sense to wrap the traditional sending also into a payload object? E.g. that it is just a wrapper around the tranditional `void*` and `len` arguments.

```class default_payload : public payload
  {
  public:
    default_payload (void* buf, size_t len) : m_buf{buf}, m_len(len)
    void write_complete(void* buf_, size_t len_) override {std::memcopy( ....) ;};
    void write_partial (void* buf_, size_t len_) override {write_complete(buf_, len_);};

    const void* data() override {return m_buf};
    size_t      size()  override {return m_len};

    private:
      void* m_buf;
      size_t m_len;
  };
  ```

I don't think this wrapper should cause any overhead and you can create it in the Send() function, and use `Apply` everywhere further down the line.

rex-schilasky:
created a new clean branch "feature/true-zwero-copy-2"

PR #1057, opened 2023-04-08
feature/true-zero-copy-3

rex-schilasky:
**Pull request type**

- [X] Feature

**What is the new behavior?**

- [X] Yes

**Sample Usage**

```cpp
#include <ecal/ecal.h>

// binary payload class to allow zero memory operation
// 
// the `WriteComplete`and `WritePartial` calls perform 
// write/copy operations on the target memory directly
class CBinaryPayload : public eCAL::CPayload
{
public:
  CBinaryPayload(size_t size_) : size(size_) {}

  bool WriteComplete(void* buf_, size_t len_) override
  {
    memset(buf_, 42, size);
    return true;
  };

  bool WritePartial(void* buf_, size_t len_) override
  {
    size_t write_idx((clock % 1024) % len_);
    char   write_chr(clock % 10 + 48);
    static_cast<char*>(buf_)[write_idx] = write_chr;
    clock++;
    return true;
  };

  size_t GetSize() override { return size; };

private:
  size_t size = 0;
  int    clock = 0;
};

// main entry
{
  // initialize eCAL API

  // create payload
  CBinaryPayload binary_payload(8 * 1024 * 1024);

  // create publisher

  // enable zero copy mode
  pub.ShmEnableZeroCopy(true);

  // enable handshake mode

  // send updates
  while (eCAL::Ok())
  {
    // send content
    pub.Send(binary_payload);
  }

  // destroy publisher
  pub.Destroy();

  // finalize eCAL API
  eCAL::Finalize();

  return(0);
}
```


rex-schilasky:
progressing now in feature/true-zero-copy-4

Issue #907, opened 2022-12-09
Unreliable RPC calls with large messages


floroeske:
**Describe the bug**
We are sending some long strings in protobuf RPC calls around and it seems like the calls get lost more often with increasing message size. This just seems to be happening when communicating over an ethernet network. I couldn't reproduce it on localhost communication.

I think it is related to https://github.com/eclipse-ecal/ecal/issues/905

This might also be related to https://github.com/eclipse-ecal/ecal/issues/581 but probably doesn't happen that often in a docker setup.

**To Reproduce**
1. Patch the example ping rpc application:

I have set the size of the string data to something similar we are trying to use in our application. It's really just a small base64 encoded image. The size is likely to be a bit larger in the future.

```
diff --git a/samples/cpp/services/ping_client/src/ping_client.cpp b/samples/cpp/services/ping_client/src/ping_client.cpp
index 75ab654e..06d111ed 100644
--- a/samples/cpp/services/ping_client/src/ping_client.cpp
+++ b/samples/cpp/services/ping_client/src/ping_client.cpp
@@ -51,10 +51,11 @@ int main(int argc, char **argv)
       //////////////////////////////////////
       PingRequest               ping_request;
       eCAL::ServiceResponseVecT service_response_vec;
-      ping_request.set_message("PING");
+      ping_request.set_message(std::string(25900, 'a'));
+      std::cout << "ping_request.message().size(): " << ping_request.message().size() << std::endl;
       if (ping_client.Call("Ping", ping_request, -1, &service_response_vec))
       {
-        std::cout << std::endl << "PingService::Ping method called with message : " << ping_request.message() << std::endl;
+        std::cout << std::endl << "PingService::Ping method called with message : " << ping_request.message().size() << std::endl;
 
         for (auto service_response : service_response_vec)
         {
@@ -84,7 +85,7 @@ int main(int argc, char **argv)
     }
 
     // sleep a second
-    std::this_thread::sleep_for(std::chrono::milliseconds(1000));
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
   }
 
   // finalize eCAL API
diff --git a/samples/cpp/services/ping_server/src/ping_server.cpp b/samples/cpp/services/ping_server/src/ping_server.cpp
index 1973d57d..c87349f9 100644
--- a/samples/cpp/services/ping_server/src/ping_server.cpp
+++ b/samples/cpp/services/ping_server/src/ping_server.cpp
[...]

PR #1067, opened 2023-04-13
feature/server-client-protocol-version

rex-schilasky:
**Pull request type**

- [X] Feature

We can not handle incompatible communication protocol changes because there is no protocol version information in the registration messages for publisher, subscriber, server and clients.

Issue Number: #907

**What is the new behavior?**
A protocol version number is introduced that can be used to later on run different protocols for server / client communication between different eCAL version.

- [X] No

### DIFF
--- a/ecal/core/src/readwrite/ecal_reader.h
+++ b/ecal/core/src/readwrite/ecal_reader.h
@@ -109,6 +109,7 @@ namespace eCAL
     void Disconnect();
     bool CheckMessageClock(const std::string& tid_, long long current_clock_);
 
+    static constexpr int                      m_version = 1;
     std::string                               m_host_name;
     int                                       m_host_id;
     int                                       m_pid;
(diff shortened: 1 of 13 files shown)


KerstinKeller (commented):
Although this is already merged, I have some remarks.

Also, I really appreciate all the (clang-tidy) cleanup, but it makes it really hard to review the PR.
Maybe in the future we can have two separate PRs?

ecal/core/src/readwrite/ecal_reader.h
```diff
@@ -109,6 +109,7 @@ namespace eCAL
     void Disconnect();
     bool CheckMessageClock(const std::string& tid_, long long current_clock_);
 
+    static constexpr int                      m_version = 1;
```
KerstinKeller, line 112: Let's set the version to 0, while we haven't made any changes to the tcp client/server protocol.

ecal/core/src/readwrite/ecal_writer.h
```diff
@@ -116,6 +116,7 @@ namespace eCAL
 
     void LogSendMode(TLayer::eSendMode smode_, const std::string & base_msg_);
 
+    static constexpr int               m_version = 1;
```
KerstinKeller, line 119: Let's set the version to 0, while we haven't made any changes to the tcp client/server protocol.

KerstinKeller (commented):
I think it makes sense to have the version set to 0, while the current protocol is still in place.

PR #1071, opened 2023-04-14
Add CMake to be able to compile flatbuffer samples and let it work wi…

KerstinKeller:
…th the newest flatbuffers version (23.3.3).

- [x] Feature

- Add CMake files to flatbuffers sample and
- Introduce `HAS_FLATBUFFERS` option
- Adjust example to structs api.


rex-schilasky (approved):
Looks okay, It's all just prepared and switched OFF, so let's see how it works if we can enable the `HAS_FLATBUFFER` option.

PR #1072, opened 2023-04-18
feature/server-client-protocol-version-2

### LINKED ISSUE #905: Q: How is reading from a Service TCP Socket handled in eCAL?
This question is about how the data-read from the TCP Socket is handled in eCAL, when a service has been called. So I am talking about the **Service Server** side.

First of all: here is a screenshot of a call stack:
image

This question explicitly goes to @rex-schilasky. Please confirm or correct me, if I am wrong.

# My understanding

From what I understand, the logic is as follows:

1. A random number of bytes is read from the Socket
    https://github.com/eclipse-ecal/ecal/blob/d78e95b1c27dd99872474d6ba9b4c382c9013c78/ecal/core/src/service/asio_server.h#L54-L60

2. If there is a callback, the data is added to a member variable. I think there always is a request callback.
    https://github.com/eclipse-ecal/ecal/blob/d78e95b1c27dd99872474d6ba9b4c382c9013c78/ecal/core/src/service/asio_server.h#L78-L82
    https://github.com/eclipse-ecal/ecal/blob/d78e95b1c27dd99872474d6ba9b4c382c9013c78/ecal/core/src/service/asio_server.h#L170

3. It is checked if there is any more data available on the socket. If there is, again a random number of bytes is read. The data is added to the `request_` variable forever or until all currently available bytes are read
    https://github.com/eclipse-ecal/ecal/blob/d78e95b1c27dd99872474d6ba9b4c382c9013c78/ecal/core/src/service/asio_server.h#L84-L91

4. If there are no more bytes available, the callback is executed with whatever has been accumulated since the last call. Then the accumulated `request_` is cleared.
   https://github.com/eclipse-ecal/ecal/blob/d78e95b1c27dd99872474d6ba9b4c382c9013c78/ecal/core/src/service/asio_server.h#L98-L99

# The issue

If it is true what I wrote here, the Service API is inherently unstable and unsafe
- There seems to be no protocol involved telling how many bytes a message has
- Multiple messages can stack up in the TCP buffer and are interpreted as one contiguous message. There is no way of knowing.
- The pure absense of data is immediatelly interpreted as "end-of-message". The code always requests a random number of bytes from the socket, so when the TCP Stream is empty e.g. because an ethernet frame has been delayed or had to be re-transmitted, the message is cut into pieces and the next chunks are interpreted as their own messages or are used to prefix the next message, that may then be available.

### LINKED ISSUE #907: Unreliable RPC calls with large messages

rex-schilasky:
**Pull request type**

- [X] Bugfix

See issue #907 and pull request #918.

Fixes #905
Fixes #907

**What is the new behavior?**
Different client/server protocol versions introduced based on the work of @floroeske. 

1. Client and server exchange their protocol version via registration layer.
2. Server will not start before any client registration is received.
3. Server starts with clients protocol version (currently 0 or 1).
4. A server can handle multiple clients with different protocol version.
5. A client can be connected to multiple server with different protocol versions.
 
- [X] Yes

Public API changed /include/ecal/ecal_callback.h.


ecal/core/src/service/asio_server.h
```diff
@@ -125,6 +140,84 @@ class CAsioSession
...
+          memcpy(&tcp_header, header_.data(), sizeof(eCAL::STcpHeader));
+          header_request_size_ = static_cast<size_t>(ntohl(tcp_header.psize_n));
+        }
+
+        // collect request_
+        if (header_request_size_ != 0 && request_.size() < header_request_size_)
+        {
+          request_ += data.substr(bytes_used, bytes_transferred - bytes_used);
+        }
+
```
KanonWY, line 177: Consider a relatively rare case, suppose pack_1 and pack_2 are sent in rapid succession on the tcpclient side. If pack_1 is internally split for some reason, and then resend part of the internally split packets (tcp automatic retransmission), then the new packet actually contains the second half of pack_1 and the first half of pack_2, the current code implementation does not do anything to the first half of pack_2.
Of course, due to the nodelay socket option enabled on the sender side, this makes it difficult to reproduce the situation, but there is still a low probability of this problem.

ecal/core/src/service/ecal_tcpclient.cpp
```diff
@@ -241,7 +316,7 @@ namespace eCAL
       // read stream header (sync)
       else
       {
-        size_t bytes_read = m_socket->read_some(asio::buffer(&tcp_header, sizeof(tcp_header)));
+        size_t const bytes_read = m_socket->read_some(asio::buffer(&tcp_header, sizeof(tcp_header)));
```
floroeske, line 319: Maybe same thing here. Validate header and check if psize_n makes sense.

ecal/core/src/service/ecal_tcpclient.cpp
```diff
@@ -196,6 +224,53 @@ namespace eCAL
...
+    {
+      std::cerr << "CTcpClient::SendRequest: Failed to send request: " << e.what() << "\n";
+      m_connected = false;
+      return false;
+    }
+
+    return true;
+  }
+
+  std::vector<char> CTcpClient::PackRequest(const std::string& request)
```
floroeske, line 260: Wondering if PackRequest and PackResponse should live in core/src/service/ecal_tcpheader.h. Maybe one PackHeader() function would be enough.

floroeske (commented):
Cool, I'll try it out in the next few days.

ecal/core/src/service/asio_server.h
rex-schilasky: First of all, thank you for your review - GREAT. I will check all in detail later.

floroeske:
I've been running this branch for the last week on some of our devices and I haven't found any issues so far. I haven't tested the protocol backwards compatibility.

FlorianReimold:
I am finally handing this over to anybody who would like to review the changes 😉

hannemn (approved):

PR #1079, opened 2023-04-24
feature/true-zero-copy-6

rex-schilasky:
**Pull request type**

- [X] Feature

**What is the new behavior?**

- [X] Yes

**Sample Usage**

```cpp
#include <ecal/ecal.h>

// binary payload writer class to allow zero copy memory operations
//
// the `Write`and `Update` calls may operate on the target memory file directly (zero copy mode)
class CBinaryPayloadWriter : public eCAL::CPayloadWriter
{
public:
  CBinaryPayloadWriter(size_t size_) : m_size(size_) {}

  bool Write(void* buffer_, size_t size_) override
  {
    memset(buffer_, 42, size_);
    return true;
  };

  bool Update(void* buffer_, size_t size_) override
  {
    size_t const write_idx((m_clock % 1024) % size_);
    char   const write_chr(m_clock % 10 + 48);
    static_cast<char*>(buffer_)[write_idx] = write_chr;
    m_clock++;
    return true;
  };

private:
  size_t m_size = 0;
  int    m_clock = 0;
};

// main entry
{
  // initialize eCAL API

  // create payload
  CBinaryPayloadWriter binary_payload_writer(8 * 1024 * 1024);

  // create publisher

  // enable zero copy mode
  pub.ShmEnableZeroCopy(true);

  // enable handshake mode

  // send updates
  while (eCAL::Ok())
  {
    // send content
    pub.Send(binary_payload_writer);
  }

  // destroy publisher
  pub.Destroy();

  // finalize eCAL API
  eCAL::Finalize();

  return(0);
}
```

### DIFF
--- a/app/mon/mon_plugins/raw_data_reflection/src/plugin_widget.cpp
+++ b/app/mon/mon_plugins/raw_data_reflection/src/plugin_widget.cpp
@@ -62,6 +62,11 @@ PluginWidget::PluginWidget(const QString& topic_name, const QString&, QWidget* p
   subscriber_.AddReceiveCallback(std::bind(&PluginWidget::ecalMessageReceivedCallback, this, std::placeholders::_2));
 }
 
+PluginWidget::~PluginWidget()
+{
+  subscriber_.RemReceiveCallback();
+}
+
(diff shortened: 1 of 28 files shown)


rex-schilasky:
Mixed up with a bug in ecal monitor :-/

PR #1082, opened 2023-04-24
feature/true-zero-copy-7

rex-schilasky:
**Pull request type**

- [X] Feature

**What is the new behavior?**

- [X] Yes

**Sample Usage**

```cpp
#include <ecal/ecal.h>

// binary payload writer class to allow zero copy memory operations
//
// the `Write`and `Update` calls may operate on the target memory file directly (zero copy mode)
class CBinaryPayloadWriter : public eCAL::CPayloadWriter
{
public:
  CBinaryPayloadWriter(size_t size_) : m_size(size_) {}

  bool Write(void* buffer_, size_t size_) override
  {
    memset(buffer_, 42, size_);
    return true;
  };

  bool Update(void* buffer_, size_t size_) override
  {
    size_t const write_idx((m_clock % 1024) % size_);
    char   const write_chr(m_clock % 10 + 48);
    static_cast<char*>(buffer_)[write_idx] = write_chr;
    m_clock++;
    return true;
  };

private:
  size_t m_size = 0;
  int    m_clock = 0;
};

// main entry
{
  // initialize eCAL API

  // create payload
  CBinaryPayloadWriter binary_payload_writer(8 * 1024 * 1024);

  // create publisher

  // enable zero copy mode
  pub.ShmEnableZeroCopy(true);

  // enable handshake mode

  // send updates
  while (eCAL::Ok())
  {
    // send content
    pub.Send(binary_payload_writer);
  }

  // destroy publisher
  pub.Destroy();

  // finalize eCAL API
  eCAL::Finalize();

  return(0);
}
```

### DIFF
--- a/ecal/core/include/ecal/ecal_payload_writer.h
+++ b/ecal/core/include/ecal/ecal_payload_writer.h
@@ -1,0 +1,83 @@
+/* ========================= eCAL LICENSE =================================
+ *
+ * Copyright (C) 2016 - 2019 Continental Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * ========================= eCAL LICENSE =================================
+*/
+
+/**
+ * @file   ecal_payload_writer.h
+ * @brief  eCAL payload writer base class
+**/
+
+#pragma once
+
+#include <cstddef>
+#include <cstring>
+
+namespace eCAL
+{
+  // base payload writer class to allow zero copy memory operations
+  // 
+  // the `Write`and `Update` calls may operate on the target memory file directly (zero copy mode)
+  class CPayloadWriter
+  {
+  public:
+    CPayloadWriter() = default;
+    virtual ~CPayloadWriter() = default;
+
+    CPayloadWriter(const CPayloadWriter &) = default;
+    CPayloadWriter(CPayloadWriter &&) = default;
+
+    CPayloadWriter& operator=(const CPayloadWriter &) = default;
+    CPayloadWriter& operator=(CPayloadWriter &&) = default;
+
+    // the provisioned memory is uninitialized ->
+    // perform a full write operation
+    virtual bool Write(void* buffer_, size_t size_) = 0;
+
+    // the provisioned memory is initialized and contains the data from the last write operation ->
+    // perform a partial write operation or just modify a few bytes here
+    //
+    // by default this operation will just call `Write`
+    virtual bool Update(void* buffer_, size_t size_) { return Write(buffer_, size_); };
+
+    // provide the size of the required memory (eCAL needs to allocate for you).
+    virtual size_t GetSize() = 0;
+  };
+
+  // payload writer class that wraps classic (void*, size_t) interface
+  class CBufferPayloadWriter : public CPayloadWriter
+  {
+  public:
+    CBufferPayloadWriter(const void* const buffer_, size_t size_) : m_buffer(buffer_), m_size(size_) {};
+
+    // make a dump memory copy
+    bool Write (void* buffer_, size_t size_) override {
+      if (size_ < m_size)      return false;
+      if (m_buffer == nullptr) return false;
+      if (m_size == 0)         return false;
+      memcpy(buffer_, m_buffer, m_size);
+      return true;
+    }
+
+    // size of the memory that needs to be copied
+    size_t GetSize() override { return m_size; };
+  
+  private:
+    const void* m_buffer = nullptr;
+    size_t      m_size   = 0;
+  };
+}
--- a/ecal/core/include/ecal/ecal_payload_writer.h
+++ b/ecal/core/include/ecal/ecal_payload_writer.h
@@ -66,6 +66,7 @@ namespace eCAL
 
     // make a dump memory copy
     bool Write (void* buffer_, size_t size_) override {
+      if (buffer_ == nullptr)  return false;
       if (size_ < m_size)      return false;
       if (m_buffer == nullptr) return false;
       if (m_size == 0)         return false;
--- a/ecal/core/include/ecal/msg/capnproto/publisher.h
+++ b/ecal/core/include/ecal/msg/capnproto/publisher.h
@@ -37,124 +37,12 @@
 #pragma warning(pop)
 #endif /*_MSC_VER*/
 
+#include <memory>
 
 namespace eCAL
 {
   namespace capnproto
   {
-    /**
-     * @brief eCAL capnp publisher class.
-     *
-     * Publisher template  class for capnp messages. For details see documentation of CPublisher class.
-     *
-    **/
-    class CBuilderPublisher : public CMsgPublisher<capnp::MallocMessageBuilder>
-    {
-    public:
-      /**
-       * @brief  Constructor.
-      **/
-      CBuilderPublisher() : CMsgPublisher<capnp::MallocMessageBuilder>()
-      {
-      }
-
-      /**
-       * @brief  Constructor.
-       *
-       * @param topic_name_  Unique topic name.
-      **/
-      CBuilderPublisher(const std::string& topic_name_, const std::string& topic_type_, const std::string& topic_desc_) : CMsgPublisher<capnp::MallocMessageBuilder>(topic_name_, topic_type_, topic_desc_)
-      {
-      }
-
-      /**
-      * @brief  Copy Constructor is not available.
-      **/
-      CBuilderPublisher(const CBuilderPublisher&) = delete;
-
-      /**
-      * @brief  Copy Constructor is not available.
-      **/
-      CBuilderPublisher& operator=(const CBuilderPublisher&) = delete;
-
-      /**
-      * @brief  Move Constructor
-      **/
-      CBuilderPublisher(CBuilderPublisher&&) = default;
-
-      /**
-      * @brief  Move assignment
-      **/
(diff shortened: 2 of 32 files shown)


ecal/core/include/ecal/msg/protobuf/publisher.h
```diff
@@ -52,13 +52,46 @@ namespace eCAL
...
+
+        bool Write(void* buf_, size_t len_) override
+        {
+          return message.SerializeToArray(buf_, (int)len_);
+        }
+
+        size_t GetSize() override {
+          size_t size(0);
+#if GOOGLE_PROTOBUF_VERSION >= 3001000
+          size = (size_t)message.ByteSizeLong();
```
hannemn, line 79: Use c++ style static_cast instead

ecal/core/include/ecal/ecal_payload_writer.h
```diff
@@ -0,0 +1,83 @@
...
+
+namespace eCAL
+{
+  // base payload writer class to allow zero copy memory operations
+  // 
+  // the `Write`and `Update` calls may operate on the target memory file directly (zero copy mode)
+  class CPayloadWriter
+  {
+  public:
+    CPayloadWriter() = default;
```
hannemn, line 38: I'm not sure if make sense to explicitly define default contructors and assignment operator in an abstract base class as you cannot call them directly.

ecal/core/include/ecal/msg/protobuf/publisher.h
```diff
@@ -52,13 +52,46 @@ namespace eCAL
...
+
+        CPayload(const CPayload&) = default;
+        CPayload(CPayload&&) noexcept = default;
+
+        CPayload& operator=(const CPayload&) = delete;
+        CPayload& operator=(CPayload&&) noexcept = delete;
+
+        bool Write(void* buf_, size_t len_) override
+        {
+          return message.SerializeToArray(buf_, (int)len_);
```
hannemn, line 73: Use c++ style cast for (int)len_.

ecal/core/include/ecal/ecal_payload_writer.h
```diff
@@ -0,0 +1,83 @@
...
+  };
+
+  // payload writer class that wraps classic (void*, size_t) interface
+  class CBufferPayloadWriter : public CPayloadWriter
+  {
+  public:
+    CBufferPayloadWriter(const void* const buffer_, size_t size_) : m_buffer(buffer_), m_size(size_) {};
+
+    // make a dump memory copy
+    bool Write (void* buffer_, size_t size_) override {
```
hannemn, line 68: Maybe should check the plausibility of buffer_ != nullptr as well before invoking memcpy.

ecal/core/src/pubsub/ecal_publisher.cpp
```diff
@@ -322,80 +321,93 @@ namespace eCAL
...
-    // publisher can send an empty package
-    // or we do not have any subscription at all
-    // then the data writer will only do some statistics
-    // for the monitoring layer and return
-    if (!IsSubscribed())
-    {
-      m_datawriter->RefreshSendCounter();
-      return(len_);
-    }
+  // @Rex, do we even need this function? I guess so for convenience...
```
hannemn, line 328: Unintented comment

ecal/core/src/io/ecal_memfile.cpp
```diff
@@ -72,8 +73,9 @@ namespace eCAL
       Destroy(create_);
 
       // reset states
-      m_created      = false;
-      m_access_state = access_state::closed;
+      m_created             = false;
+      m_payload_initialized = false;
```
hannemn, line 77: You should reset m_payload_inititialized also in CMemoryFile::Destroy()

ecal/core/include/ecal/msg/capnproto/publisher.h
```diff
@@ -226,7 +160,8 @@ namespace eCAL
 
       void Send()
       {
-        publisher.Send(builder);
+        CPayload payload{ *builder };
+        eCAL::CPublisher::Send(payload);
```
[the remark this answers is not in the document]
rex-schilasky: false positive

ecal/core/include/ecal/ecal_payload_writer.h
KerstinKeller: I think you still need them. Especially the assignment / move ones, since you NEED the virtual desctructor. If you don't have the others, the objects will not be movable!

ecal/core/src/readwrite/ecal_writer_inproc.h
```diff
@@ -44,16 +44,16 @@ namespace eCAL
...
     SWriterInfo GetInfo() override;
 
     bool Create(const std::string& host_name_, const std::string& topic_name_, const std::string & topic_id_) override;
     // this virtual function is called during construction/destruction,
     // so, mark it as final to ensure that no derived classes override it.
-    bool Destroy() final override;
+    bool Destroy() final;
 
-    bool Write(const SWriterData& data_) override;
+    bool Write(const void* const buf_, const SWriterAttr& attr_) override;
```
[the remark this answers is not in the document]
rex-schilasky: fixed

ecal/core/src/readwrite/ecal_writer_tcp.h
```diff
@@ -47,27 +47,27 @@ namespace eCAL
...
     SWriterInfo GetInfo() override;
 
     bool Create(const std::string& host_name_, const std::string& topic_name_, const std::string & topic_id_) override;
     // this virtual function is called during construction/destruction,
     // so, mark it as final to ensure that no derived classes override it.
-    bool Destroy() final override;
+    bool Destroy() final;
 
-    bool Write(const SWriterData& data_) override;
+    bool Write(const void* const buf_, const SWriterAttr& attr_) override;
```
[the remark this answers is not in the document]
rex-schilasky: fixed

ecal/core/include/ecal/ecal_payload_writer.h
rex-schilasky: fixed

ecal/core/src/pubsub/ecal_publisher.cpp
rex-schilasky: fixed

ecal/core/src/pubsub/ecal_publisher.cpp
```diff
@@ -322,80 +321,93 @@ namespace eCAL
...
+     bool const sent = m_datawriter->Write(payload_, write_time, m_id);
+
+     if (acknowledge_timeout_ms_ != DEFAULT_ACKNOWLEDGE_ARGUMENT)
+     {
+       // reset acknowledge timeout
+       m_datawriter->ShmSetAcknowledgeTimeout(current_acknowledge_timeout_ms);
+     }
 
-    return sent ? len_ : 0;
+     //@Rex we have to call GetSize() again. Can't we return this from the Write call?
```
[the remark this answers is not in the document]
rex-schilasky: fixed, now just number of pre-calculated bytes (`payload_buffer_size`) will be returned or zero if writing failed

ecal/core/src/io/ecal_memfile.cpp
rex-schilasky: fixed

ecal/core/src/readwrite/ecal_writer_tcp.h
```diff
@@ -47,27 +47,27 @@
...
 
     std::string GetConnectionParameter() override;
 
   private:
-    static std::mutex                       g_tcp_writer_executor_mtx;
+    static std::mutex                            g_tcp_writer_executor_mtx;
```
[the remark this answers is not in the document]
rex-schilasky: won't be fixed for now

ecal/core/src/readwrite/ecal_writer_udp_mc.h
```diff
@@ -43,22 +43,22 @@ namespace eCAL
...
     SWriterInfo GetInfo() override;
 
     bool Create(const std::string& host_name_, const std::string& topic_name_, const std::string & topic_id_) override;
     // this virtual function is called during construction/destruction,
     // so, mark it as final to ensure that no derived classes override it.
-    bool Destroy() final override;
+    bool Destroy() final;
 
-    bool Write(const SWriterData& data_) override;
+    bool Write(const void* const buf_, const SWriterAttr& attr_) override;
```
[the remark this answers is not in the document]
rex-schilasky: fixed

ecal/core/include/ecal/msg/protobuf/publisher.h
```diff
@@ -52,13 +52,46 @@ namespace eCAL
      *
     **/
     template <typename T>
-    class CPublisher : public CMsgPublisher<T>
+    class CPublisher : public eCAL::CPublisher
```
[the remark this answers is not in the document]
rex-schilasky: It's from my understanding not an API break. What do you mean @KerstinKeller ?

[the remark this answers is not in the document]
rex-schilasky: Fixed then ..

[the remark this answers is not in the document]
rex-schilasky: Fixed

ecal/core/include/ecal/ecal_payload_writer.h
hannemn: I was talking about the contructors, not the destructor. The object is moveable and copyable by default even if you mark the destructor as virtual method.

hannemn (approved):

chengguizi:
Very interesting development! Two questions:
1. Will this be included in the next version release?
2. Can it be already used with serialization layer, such as protobuf message or capnproto message. The example here shows only binary data

rex-schilasky:
@chengguizi the zero-copy message transport feature will be part of the next 5.12 release.

The message protocols google protobuf and capnproto will be supported as well, flatbuffers currently not.

For these serialization formats the zero-copy feature would change the behavior that way that the message content is serialized in the opened memory file directly without creating an intermediate buffer. This new behavior may also have some overall performance drawbacks if your serialization call takes a long time and is so blocking the file for connected subscribers. But the good news is you can switch on/off this feature for every single publisher.

chengguizi:
Pretty much look forward, can't wait for 5.12 release!

PR #1083, opened 2023-04-24
hotfix/monplugins-subscriber-callback

rex-schilasky:
**Pull request type**

- [X] Bugfix

Two monitor plugin widgets (raw and string) do not remove the callback of the attached subscriber before destruction. This can lead to data races when closing the plugin window.

**What is the new behavior?**
Data receive callback is removed in the plugin destructor.


KerstinKeller (approved):
Looks good. But what about the protobuf plugin and the others?

rex-schilasky:
> Looks good. But what about the protobuf plugin and the others?

All other plugins implemented that fix.

PR #1086, opened 2023-04-27
feature/entity-unregistration

rex-schilasky:
**Pull request type**

- [X] Feature

Destroyed entities (publisher, subscriber, client, server) do not active unregister from the monitoring layer. These entities are removed after a monitoring timeout that can be defined in the global configuration ecal.ini file (`[monitoring]/timeout`).

**What is the new behavior?**
New additional unregistration messages are fired by all ecal entities on destruction. Entities disappear immediately from the monitoring layer.

### DIFF
--- a/ecal/core/src/ecal_expmap.h
+++ b/ecal/core/src/ecal_expmap.h
@@ -212,6 +212,19 @@ namespace eCAL
         }
       };
 
+      // Remove specific element from the cache
+      // @Kerstin, pretty sure that this is not well implemented, please take a look
+      bool remove(const Key& k)
+      {
+        auto it = _key_to_value.find(k);
+        if (it != _key_to_value.end())
+        {
+          _key_to_value.erase(k);
+          return true;
+        }
+        return false;
+      };
+
       // Remove all elements from the cache 
       void clear()
       {
--- a/ecal/core/src/mon/ecal_monitoring_impl.h
+++ b/ecal/core/src/mon/ecal_monitoring_impl.h
@@ -43,7 +43,7 @@ namespace eCAL
   {
   public:
     CMonitoringImpl();
-    ~CMonitoringImpl();
+    ~CMonitoringImpl() override = default;
 
     void Create();
     void Destroy();
@@ -59,22 +59,30 @@ namespace eCAL
     int PubMonitoring(bool state_, std::string& name_);
     int PubLogging(bool state_, std::string& name_);
 
+    bool HasSample(const std::string& /* sample_name_ */) override { return(true); };
+    size_t ApplySample(const eCAL::pb::Sample& ecal_sample_, eCAL::pb::eTLayerType /*layer_*/) override;
+
+  protected:
+    bool RegisterProcess(const eCAL::pb::Sample& sample_);
+    bool UnregisterProcess(const eCAL::pb::Sample& sample_);
+
+    bool RegisterServer(const eCAL::pb::Sample& sample_);
+    bool UnregisterServer(const eCAL::pb::Sample& sample_);
+
+    bool RegisterClient(const eCAL::pb::Sample& sample_);
+    bool UnregisterClient(const eCAL::pb::Sample& sample_);
+
     enum ePubSub
     {
       publisher = 1,
       subscriber = 2,
     };
 
-    bool HasSample(const std::string& /* sample_name_ */) { return(true); };
-    size_t ApplySample(const eCAL::pb::Sample& ecal_sample_, eCAL::pb::eTLayerType /*layer_*/);
-
-    bool RegisterProcess(const eCAL::pb::Sample& sample_);
-    bool RegisterServer(const eCAL::pb::Sample& sample_);
-    bool RegisterClient(const eCAL::pb::Sample& sample_);
     bool RegisterTopic(const eCAL::pb::Sample& sample_, enum ePubSub pubsub_type_);
+    bool UnregisterTopic(const eCAL::pb::Sample& sample_, enum ePubSub pubsub_type_);
+
     void RegisterLogMessage(const eCAL::pb::LogMessage& log_msg_);
 
-  protected:
     using TopicMonMapT = eCAL::Util::CExpMap<std::string, eCAL::Monitoring::STopicMon>;
     struct STopicMonMap
     {
@@ -131,7 +139,7 @@ namespace eCAL
 #endif
       }
     };
-    typedef std::set<std::string, InsensitiveCompare> StrICaseSetT;
+    using StrICaseSetT = std::set<std::string, InsensitiveCompare>;
 
     STopicMonMap* GetMap(enum ePubSub pubsub_type_);
 
@@ -162,7 +170,7 @@ namespace eCAL
     SClientMonMap                                m_clients_map;
 
     // logging
-    typedef std::list<eCAL::pb::LogMessage> LogMessageListT;
+    using LogMessageListT = std::list<eCAL::pb::LogMessage>;
     std::mutex                                   m_log_msglist_sync;
     LogMessageListT                              m_log_msglist;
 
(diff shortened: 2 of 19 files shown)


ecal/core/src/ecal_expmap.h
```diff
@@ -212,6 +212,19 @@ namespace eCAL
         }
       };
 
+      // Remove specific element from the cache
+      // @Kerstin, pretty sure that this is not well implemented, please take a look
+      bool remove(const Key& k)
```
KerstinKeller, line 217: actually, since it's a wrapper around the container, i would use the same function names, e.g. erase.
Also you have to remove it from the two containers: `_key_to_value` and `_key_tracker` variables (see the `clear()` function).'

KerstinKeller (changes requested):
See my comment regarding the expmap.
The rest in general looks good, but it's a bit much to really tell if everythingis ok.

ecal/core/src/ecal_expmap.h
rex-schilasky: Renamed and fixed to remove item from both internal containers (thanks @hannemn)

ecal/core/src/ecal.cpp
```diff
@@ -119,12 +121,24 @@ namespace eCAL
       }
     }
 
+    // be nice and gently to our customers :-)
+    static bool ecal_welcome_msg(false);
+    if (!ecal_welcome_msg)
+    {
+      std::cout << std::endl;
+      std::cout << "----------------------------------------" << std::endl;
+      std::cout << "        THANK YOU FOR USING eCAL        " << std::endl;
```
FlorianReimold, line 130: Well yeah, it's funny, but is it necessary? Can it be disabled? This will break all tools linking against eCAL that operate on STDIN / STDOUT communication.

ecal/core/src/ecal.cpp
```diff
@@ -231,9 +245,22 @@ namespace eCAL
...
     delete g_globals_ctx;
     g_globals_ctx = nullptr;
+
+    // be nice and gently to our customers :-)
+    static bool ecal_goodbye_msg(false);
+    if (!ecal_goodbye_msg)
+    {
+      std::cout << std::endl;
+      std::cout << "----------------------------------------" << std::endl;
+      std::cout << " THANK YOU FOR SHUTDOWN eCAL GRACEFULLY " << std::endl;
```
FlorianReimold, line 258: Same thing here, I think this would count as a debug message that should not be printed by default. It is only important for developers.

ecal/core/src/ecal.cpp
rex-schilasky: That makes unfortunately sense ..

ecal/core/src/ecal.cpp
rex-schilasky: eCAL's logging interface is no more available here .. maybe we just comment it out and can uncomment/use that for debugging ?

ecal/core/src/readwrite/ecal_reader.cpp
```diff
@@ -299,14 +294,39 @@ namespace eCAL
...
-    if(g_registration_provider()) g_registration_provider()->RegisterTopic(m_topic_name, m_topic_id, ecal_reg_sample, force_);
+    if(g_registration_provider() != nullptr) g_registration_provider()->RegisterTopic(m_topic_name, m_topic_id, ecal_reg_sample, force_);
 #ifndef NDEBUG
     // log it
     Logging::Log(log_level_debug4, m_topic_name + "::CDataReader::DoRegister");
 #endif
     return(true);
   }
 
+  bool CDataReader::Unregister(const bool force_)
```
hannemn, line 305: I am wondering if it makes sense to have an explicit `force` parameter? By invoking the `Unregister()` call the respective ecal unregister sample has to be sent anyways instantly. In comparsion to `Register()` there are no recurrent unregistration updates.

ecal/core/src/readwrite/ecal_reader.cpp
```diff
@@ -645,14 +692,8 @@ namespace eCAL
...
-    switch (type_)
-    {
-    case eCAL::pb::tl_ecal_tcp:
-      CTCPReaderLayer::Get()->SetConnectionParameter(par);
-      break;
-    default:
-      break;
-    }
+    // process only for tcp layer
+    CTCPReaderLayer::Get()->SetConnectionParameter(par);
```
hannemn, line 696: In this case `SetConnectionParameter()` will be always called even if `type_` is for instance` eCAL::pb::tl_ecal_shm`. Is that the desired behaviour?

hannemn (changes requested):

ecal/core/src/ecal.cpp
```diff
@@ -26,6 +26,8 @@
 #include <tclap/CmdLine.h>
 #include <custom_tclap/advanced_tclap_output.h>
 
+#include <iostream>
```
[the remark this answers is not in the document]
rex-schilasky: fixed with next commit

ecal/core/src/ecal.cpp
```diff
@@ -165,7 +167,7 @@ namespace eCAL
     g_globals_ctx_ref_cnt++;
 
     // (post)initialize single components
-    int success = g_globals()->Initialize(components_, &config_keys);
+    int const success = g_globals()->Initialize(components_, &config_keys);
```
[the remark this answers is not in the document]
rex-schilasky: fixed with next commit

ecal/core/src/mon/ecal_monitoring_impl.h
```diff
@@ -59,22 +59,30 @@ namespace eCAL
     int PubMonitoring(bool state_, std::string& name_);
     int PubLogging(bool state_, std::string& name_);
 
+    bool HasSample(const std::string& /* sample_name_ */) override { return(true); };
+    bool ApplySample(const eCAL::pb::Sample& ecal_sample_, eCAL::pb::eTLayerType /*layer_*/) override;
+
+  protected:
```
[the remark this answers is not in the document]
rex-schilasky: yes

ecal/core/src/readwrite/ecal_reader.cpp
rex-schilasky: fully correct I removed the `force_` parameter for the unregistration functions for datawriter and client/server as well, they were always called with `true`

ecal/core/src/ecal_expmap.h
```diff
@@ -212,6 +212,20 @@ namespace eCAL
         }
       };
 
+      // Remove specific element from the cache
+      // @Kerstin, pretty sure that this is not well implemented, please take a look
```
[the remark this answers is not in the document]
rex-schilasky: I checked that again debugging the unit test
```cpp
TEST(ExpMap, ExpMapRemove)
{
  eCAL::Util::CExpMap<std::string, int> expmap(std::chrono::milliseconds(200));
  expmap["A"] = 1;
  EXPECT_EQ(1, expmap.size());
  EXPECT_TRUE(expmap.erase("A"));
  EXPECT_EQ(0, expmap.size());
  EXPECT_FALSE(expmap.erase("B"));
}
```
and both internal container `_key_tracker `and `_key_to_value` are cleaned up correctly.

hannemn (approved):

ecal/core/src/readwrite/ecal_writer.cpp
```diff
@@ -654,10 +657,14 @@ namespace eCAL
   void CDataWriter::ApplyLocSubscription(const std::string& process_id_, const std::string& tid_, const std::string& ttype_, const std::string& tdesc_, const std::string& reader_par_)
   {
     Connect(tid_, ttype_, tdesc_);
+
+    // add key to local subscriber map
+    const std::string topic_key = process_id_ + tid_;
```
KerstinKeller, line 662: This introduces unwanted side-effects!

PR #1088, opened 2023-05-03
Mon GUI: Added Raw Data Widget

FlorianReimold:
Added a Raw-Data-Widget to the eCAL Mon GUI

The Raw Data Widget displays the eCAL Monitoring Information in textual form. The Protobuf `DebugString()` is used for generating the Text. Therefore, the Raw Data Widget will never need to be adapted, even if we add / remove elements from the monitoring layer.

Features:
- Display raw text (monospace)
- Syntax highlighting (different ones for dark- and light mode)
- Save-to-file
- Full text search with result highlighting and keyboard shortcuts

image


app/mon/mon_gui/src/widgets/ecalmon_tree_widget/topic_widget.cpp
rex-schilasky, line 1: this is just a search and replace from `format` -> `format_` I assume :-)

FlorianReimold:
I just saw that this PR accidentally refactored some stuff that wasn't supposed to be refactored

rex-schilasky (approved):

PR #1089, opened 2023-05-04
Add CapnProto Publisher and Subscriber to Python API

CruxDevStuff:
**Pull request type**

 

 

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes

**What is the new behavior?**

- Subscribe to CapnProto Messages from Python
- Publish CapnProto Messages from Python

- [ ] Yes
- [X] No


CruxDevStuff:
> Thank you @CruxDevStuff for contributing ! We will check this next week. If you have python samples for sending and receiving capnproto messages like those [ones](https://github.com/eclipse-ecal/ecal/tree/master/samples/cpp/pubsub/capnp) in cpp, it would be great too.

I've added an [Example](https://github.com/CruxDevStuff/ecal/tree/master/samples/python/capnp)

rex-schilasky:
> > Thank you @CruxDevStuff for contributing ! We will check this next week. If you have python samples for sending and receiving capnproto messages like those [ones](https://github.com/eclipse-ecal/ecal/tree/master/samples/cpp/pubsub/capnp) in cpp, it would be great too.
> 
> 
> 
> I've added an [Example](https://github.com/CruxDevStuff/ecal/tree/master/samples/python/capnp)

Great. Thank you.

rex-schilasky:
@CruxDevStuff can you please create an [eclipse account](https://accounts.eclipse.org/user/login?destination=legal/eca/validation/153175), then we can merger your PR after the review has  been done.

samples/python/capnp/messages.capnp
```diff
@@ -0,0 +1,8 @@
+@0xb6249c5df1f7e512;
```
KerstinKeller, line 1: can you use the same message like is used for the c++ samples? this way we can communicate the samples with python send / c++ receive and vice versa.

KerstinKeller (changes requested):
Hi, sorry for the long wait.
Thanks for making this contribution, however I do have a few remarks:

- Dependencies: we now have a dependency to `pycapnp`, it needs to be added to requirements.txt for ecal.
- Before running the samples, the user will have to run the capnproto compiler on the schema file themselves. I'd prefer if we'd check in the generated code here as we do in the protobuf case. Or, alternatively, have a way to generate this on python import? At least it should be in the README.md

Peguen:
In the course of changing the Python API and bindings during 6.x release, this functionality will be implemented.

PR #1090, opened 2023-05-05
Core CMake: Restructured source_groups of files for IDEs

FlorianReimold:
Files are now not artifically scattered into different folders anymore (e.g. in Visual Studio). Rather, the filesystem is used, where (private) header files are usually located next to the corresponding cpp file. So now, the `src` dir contains everything private and the `include` dir contains everything public, just as in the folder structure.

grafik


rex-schilasky (approved):
Small change, great effect :-)

PR #1091, opened 2023-05-07
Fix building python bindings on unix systems with both python2 and python3 

CruxDevStuff:
**Pull request type**

 

 

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes

Building ecal with python bindings on a unix system with both python2 and python3 fails. CMake selects python2 instead of python3 when both versions are installed on the same system. 

```
-- Could NOT find Python (missing: Python_INCLUDE_DIRS Development Development.Module Development.Embed) (found version "2.7.18")
-- Python interpreter found: /usr/bin/python2.7                                                                                 
-- Installing Python extensions                                                                                                 
-- Could NOT find Python (missing: Python_INCLUDE_DIRS Development Development.Module Development.Embed) (found version "2.7.18")
CMake Error at cmake/helper_functions/ecal_python_functions.cmake:47 (Python_add_library):                                      
  Unknown CMake command "Python_add_library".                                                                                   
Call Stack (most recent call first):                                                                                            
  lang/python/core/CMakeLists.txt:39 (ecal_add_python_module)           

```

Issue Number: N/A

**What is the new behavior?**

- CMake checks if its a unix system and sets `Python_EXECUTABLE` to  ```/usr/bin/python3```
```
if (UNIX)
     set(Python_EXECUTABLE /usr/bin/python3)
endif (UNIX)
```

- [ ] Yes
- [x] No

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,10 @@
 
 cmake_minimum_required(VERSION 3.13)
 
+if (UNIX)
+    set(Python_EXECUTABLE /usr/bin/python3)
+endif (UNIX)	   
+
 include(CMakeDependentOption)
 include("${CMAKE_CURRENT_LIST_DIR}/thirdparty/cmake_functions/qt/qt_msvc_path.cmake")
 


FlorianReimold:
Hi @CruxDevStuff,

Thanks for contributing!
However, hardcoding a path in the CMakeLists.txt is not the proper way to solve the issue and will lead to many problems. Basically whenever you want to use anything else then what is hardcoded, e.g. different python versions that may be natively installed as `/usr/bin/python3.10`, `/usr/bin/python3.11` or when you use a python venv. We make heavy use of python venvs in our GitHub Action build scripts to generate all those different python wheels.
What you should rather do is provide the path as command line argument to CMake, or set the according environment variable. The CMake command line expects the python path as `-DPython_EXECUTABLE=/usr/bin/python3`. I think you can also provide the `Python3_ROOT_DIR` variable for FindPython3: https://cmake.org/cmake/help/latest/module/FindPython3.html
As mentioned earlier, we use python venvs in our GitHub Action CI, which also solves the issue of choosing between python2 / python3, as a venv only holds 1 python version that you specifically chose when creating the venv. You can check how we create those venvs here in our build scripts:
https://github.com/eclipse-ecal/ecal/blob/6b2843ad29597e7de06a6f80e99572006eeb097a/.github/workflows/build-ubuntu-22.yml#L45-L54

Also note the command line parameter that tells cmake to only find a currently activated venv:
https://github.com/eclipse-ecal/ecal/blob/6b2843ad29597e7de06a6f80e99572006eeb097a/.github/workflows/build-ubuntu-22.yml#L96

PR #1097, opened 2023-05-09
Upgrade github actions ecal build to v142, and remove C++17 hack.

KerstinKeller:
eCAL installer will be build with v142 toolset.


KerstinKeller:
Yes, we discussed this, but VS2017 is not enough, ftxui does not build with it. So I discussed with Florian to do 2019 instead.
Maybe later we will also distribute an SDK only for VS2015.

FlorianReimold (approved):

PR #1100, opened 2023-05-11
clang-tidy/can-be-declared-const

rex-schilasky:
**Pull request type**

- [X] Other (please describe): 
Clang-Tidy

Clang-tidy warnings.

**What is the new behavior?**
Fixed only clang-tidy warnings for local variables which could be declared as const.

- [X] No


rex-schilasky:
ftuxi added :-(

Issue #1103, opened 2023-05-11
[Python] message with default values only is None


ENAC-Robotique:
### Problem Description

Using the python bindings, I get this error when receiving a message with only default values (e.g. only 0):
```
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/ecal/core/subscriber.py", line 131, in _on_receive
    proto_message.ParseFromString(msg)
TypeError: a bytes-like object is required, not 'NoneType'
```
I temporarily "solved" the issue by inserting these 2 lines at line 131 (https://github.com/eclipse-ecal/ecal/blob/master/lang/python/core/ecal/core/subscriber.py#L131):
```
if msg is None:
    msg = b''
```

### How to reproduce

1. Define the following protobuf message:
```
message Position {
    float x = 1;
    float y = 2;
    float theta = 3;
}
```
2. In a python program, make a ProtoSubscriber over this message, and set a callback.
3. Send a Position message from an other python program with all fields at the value 0.

### How did you get eCAL?

Ubuntu PPA (apt-get)

### Environment

- eCAL: v5.11.4 (2023-04-06 09:25:10 +0200)
- OS: linux Ubuntu 22.04 ARM (raspberryPi 4)
- Python: 3.10.6

### eCAL System Information

```text
------------------------- SYSTEM ---------------------------------
Version                  : v5.11.4 (2023-04-06 09:25:10 +0200)
Platform                 : linux

------------------------- CONFIGURATION --------------------------
Default INI              : /etc/ecal/ecal.ini

------------------------- NETWORK --------------------------------
Host name                : cooking-mama
Network mode             : cloud
Network ttl              : 2
Network sndbuf           : 5 MByte
Network rcvbuf           : 5 MByte
Multicast group          : 239.0.0.1
Multicast mask           : 0.0.0.15
Multicast ports          : 14000 - 14010
Multicast join all IFs   : on
Bandwidth limit (udp)    : not limited

------------------------- TIME -----------------------------------
Synchronization realtime : "ecaltime-localtime"
Synchronization replay   : 
State                    :  synchronized 
Master / Slave           :  Master 
Status (Code)            : "everything is fine." (0)

------------------------- PUBLISHER LAYER DEFAULTS ---------------
Layer Mode INPROC        : off
Layer Mode SHM           : auto
Layer Mode TCP           : off
Layer Mode UDP MC        : auto

------------------------- SUBSCRIPTION LAYER DEFAULTS ------------
[...]

PR #1104, opened 2023-05-12
Python: Intermediate callback in subscribers, to map `None` msg in ca…

KerstinKeller:
…llbacks to ByteArray of length 0 `b""`, see #1103

I'd rather solve this from cpp instead of Python here, but maybe someone has an idea.
@rex-schilasky @FlorianReimold @hannemn

### DIFF
--- a/lang/python/core/ecal/core/core.py
+++ b/lang/python/core/ecal/core/core.py
@@ -752,7 +752,8 @@ class subscriber(object):
     :param callback: python callback function (f(topic_name, msg, time))
 
     """
-    return sub_set_callback(self.thandle, callback)
+    self.callback = callback
+    return sub_set_callback(self.thandle, self._on_receive)
 
   def rem_callback(self, callback):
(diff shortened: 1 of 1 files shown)


FlorianReimold:
That actually looks stupid, but I think it is the proper solution. What else did you have in mind?

FlorianReimold (approved):

FlorianReimold:
Hm, probably we could. But giving some random pointer to non-owned memory feels extremely unsafe, even if we pass a length of 0.

Fabien-B:
Hello, I am the one that post #1103 (with an other github account).
This seems to do the trick to create an empty bytearray:
```
PyByteArray_FromStringAndSize(NULL, 0);
```

KerstinKeller:
Ah thanks for the heads up, I think we can try this one. We've used `PyBuildValue` everywhere, but I don't know why we wouldn't be able to use `PyByteArray_FromStringAndSize` instead...

PR #1112, opened 2023-05-25
feature/qnx-performance-511

rex-schilasky:
**Pull request type**

 

 

- [X] Feature

System memory API functions like malloc, memset, std::vector clean/resize perform bad on specific qnx embedded platforms.

**What is the new behavior?**
* clear/resize vector -> just resize vector if needed
* udp snd payload buffer static

- [X] No


ecal/core/src/readwrite/ecal_writer_udp_mc.cpp
```diff
@@ -73,17 +72,19 @@ namespace eCAL
...
     if (Config::IsNetworkEnabled()) ttl = Config::GetUdpMulticastTtl();
     attr.loopback   = true;
     attr.ttl        = ttl;
-    m_sample_snd_loopback.Create(attr);
+    m_udp_sender_loopback    = std::make_shared<CUDPSender>(attr);
+    m_sample_sender_loopback = std::make_shared<CSampleSender>(m_udp_sender_loopback, m_udp_ipaddr);
+
+    // create udp/sample sender without activated loop-back
+    attr.loopback   = false;
+    attr.ttl        = Config::GetUdpMulticastTtl();
```
KerstinKeller, line 85: is the ttl set twice now? (see lines above)
Maybe you make 2 different attribute objects to make tings clearer? (`attr_loopback`, `attr_no_loopback` ?)

rex-schilasky: Kristof and me noticed this as well :-)

PR #1113, opened 2023-05-25
Python: Zero size messages should create zero size bytearray instead …

KerstinKeller:
…of None. #1104

- [x] Bugfix

Sending empty messages creates Python `None` types instead of a zero size Byte array.


rex-schilasky (approved):
This was hard to review, but looks okay to me ;-)

Issue #1074, opened 2023-04-18
If memfile_buffer_count != 0, first messages are lost


KerstinKeller:
### Problem Description

With a higher `memfile_buffer_count`, not all messages reach the subscriber, see the testcase in this commit:
https://github.com/eclipse-ecal/ecal/compare/master...hotfix/shm-buffer-count-testcase

### How to reproduce

Run the provided testcase.

### How did you get eCAL?

I don't know

### Environment

Current master

chengguizi:
Any update on this so far? Not sure how this would be affected by the major zero-copy change in #1082 

Having a circular buffer of memfiles + zero copy would be a really great feature to have for low-latency, high-frequency, low-drop message system.

rex-schilasky:
@chengguizi this fix is in progress and will be part of 5.12.

chengguizi:
Sounds good! Very look forward to an improved zero-copy implementation that also supports circular buffer (hence avoid missing messages) :)

PR #1114, opened 2023-05-26
hotfix/data-loss-memfile-buffering

rex-schilasky:
**Pull request type**

- [X] Bugfix

If we set  the number of memory files that should be used by a Publisher, the change is applied in the Send/Write action. Because registration of these shm paramters need some time (registration refresh cylce) the first data packages get lost.

Issue Number: #1074

**What is the new behavior?**
The number of handled memory files will be adapted and communicated by ShmSetBufferCount.

- [X] No

### DIFF
--- a/ecal/core/src/readwrite/ecal_writer.cpp
+++ b/ecal/core/src/readwrite/ecal_writer.cpp
@@ -352,7 +352,11 @@ namespace eCAL
 
   bool CDataWriter::ShmSetBufferCount(size_t buffering_)
   {
-    if (buffering_ < 1) return false;
+    if (buffering_ < 1)
+    {
+      buffering_ = 1;
+      std::cerr << "CDataWriter::ShmSetBufferCount minimal number of memory files is 1 !" << std::endl;
+    }
     m_buffering_shm = static_cast<size_t>(buffering_);
 
     // adapt number of used memory files
--- a/ecal/core/src/readwrite/ecal_writer.cpp
+++ b/ecal/core/src/readwrite/ecal_writer.cpp
@@ -354,8 +354,8 @@ namespace eCAL
   {
     if (buffering_ < 1)
     {
-      buffering_ = 1;
-      std::cerr << "CDataWriter::ShmSetBufferCount minimal number of memory files is 1 !" << std::endl;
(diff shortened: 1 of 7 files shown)


ecal/core/src/readwrite/ecal_writer.cpp
```diff
@@ -347,10 +350,21 @@ namespace eCAL
...
 
-  bool CDataWriter::ShmSetBufferCount(long buffering_)
+  bool CDataWriter::ShmSetBufferCount(size_t buffering_)
   {
-    if (buffering_ < 1) return false;
+    if (buffering_ < 1)
+    {
+      buffering_ = 1;
+      std::cerr << "CDataWriter::ShmSetBufferCount minimal number of memory files is 1 !" << std::endl;
+    }
```
rex-schilasky, line 359: return false here
eCAL::LogMessage ? log_error

ecal/core/src/readwrite/ecal_writer_shm.cpp
```diff
@@ -105,51 +101,74 @@ namespace eCAL
     return true;
   }
 
+  bool CDataWriterSHM::SetBufferCount(size_t buffer_count_)
+  {
+    if (buffer_count_ < 1)
+    {
+      buffer_count_ = 1;
+      std::cerr << "CDataWriterSHM::SetBufferCount minimal number of memory files is 1 !" << std::endl;
```
rex-schilasky, line 109: same as in datawriter return false and log message

ecal/core/src/readwrite/ecal_writer.cpp
```diff
@@ -347,10 +350,21 @@ namespace eCAL
...
-    if (buffering_ < 1) return false;
+    if (buffering_ < 1)
+    {
+      Logging::Log(log_level_error, m_topic_name + "::CDataWriter::ShmSetBufferCount minimal number of memory files is 1 !");
+      return false;
+    }
     m_buffering_shm = static_cast<size_t>(buffering_);
+
+    // adapt number of used memory files
+    if (m_created)
```
hannemn, line 363: The function should also fail with return value false if m_created is NOT true, since the buffer count will not be set as from the user desired.

hannemn (changes requested):
See comment! Rest looks good to me as already discussed.

hannemn (approved):
Now everthing is clear and your changes can go into the master ;-)

PR #1124, opened 2023-06-10
feat: add package.xml

wep21:
**Pull request type**

 

 

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe): add package.xml to be recognized as a ros package suggested [here](https://github.com/eclipse-ecal/rmw_ecal/pull/61#issuecomment-1550505782)

Issue Number: N/A

**What is the new behavior?**

-
-
-

- [ ] Yes
- [ ] No

**Other information**

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -13,10 +13,7 @@
 
     <buildtool_depend>cmake</buildtool_depend>
 
-    <depend>hdf5</depend>
-    <depend>libqwt-qt5-dev</depend>
     <depend>protobuf-dev</depend>
-    <depend>qtbase5-dev</depend>
 
     <export>
         <build_type>cmake</build_type>
--- a/package.xml
+++ b/package.xml
@@ -2,7 +2,7 @@
 <?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="3">
     <name>ecal</name>
-    <version>5.11.4</version>
+    <version>5.12.0-nightly2</version>
     <description>eCAL - enhanced Communication Abstraction Layer. A fast publish-subscribe cross-plattform middleware using Shared Memory and UDP.</description>
     <maintainer email="ecal-dev@eclipse.org">Eclipse Foundation, Inc.</maintainer>
     <license>Apache 2.0</license>
--- a/package.xml
+++ b/package.xml
@@ -2,7 +2,7 @@
 <?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="3">
     <name>ecal</name>
-    <version>5.12.0-nightly2</version>
+    <version>5.12.0</version>
     <description>eCAL - enhanced Communication Abstraction Layer. A fast publish-subscribe cross-plattform middleware using Shared Memory and UDP.</description>
     <maintainer email="ecal-dev@eclipse.org">Eclipse Foundation, Inc.</maintainer>
     <license>Apache 2.0</license>
--- a/package.xml
+++ b/package.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+    <name>ecal</name>
+    <version>5.11.4</version>
+    <description>eCAL - enhanced Communication Abstraction Layer. A fast publish-subscribe cross-plattform middleware using Shared Memory and UDP.</description>
+    <maintainer email="ecal-dev@eclipse.org">Eclipse Foundation, Inc.</maintainer>
+    <license>Apache 2.0</license>
+
+    <url type="website">https://eclipse-ecal.github.io/ecal</url>
+    <url type="bugtracker">https://github.com/eclipse-ecal/ecal/issues</url>
+    <url type="repository">https://github.com/eclipse-ecal/ecal</url>
+
+    <buildtool_depend>cmake</buildtool_depend>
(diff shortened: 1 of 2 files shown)


FlorianReimold:
Please tell us anything about this PR, e.g. what that file is good for.

package.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+    <name>ecal</name>
+    <version>5.11.4</version>
```
FlorianReimold, line 5: Is there any way to make this variable? For instance making it check git tags, creating this file with cmake etc.
This PR is to master, so this defnitively is NOT 5.11.4. It's more like a 5.12 alpha. Also, maintaining version numbers that are scattered across files is always a pain

package.xml
```diff
@@ -0,0 +1,24 @@
...
+    <url type="website">https://eclipse-ecal.github.io/ecal</url>
+    <url type="bugtracker">https://github.com/eclipse-ecal/ecal/issues</url>
+    <url type="repository">https://github.com/eclipse-ecal/ecal</url>
+
+    <buildtool_depend>cmake</buildtool_depend>
+
+    <depend>hdf5</depend>
+    <depend>libqwt-qt5-dev</depend>
+    <depend>protobuf-dev</depend>
+    <depend>qtbase5-dev</depend>
```
FlorianReimold, line 19: Where did this list come from?
- libecal_core.so / ecal_core.dll does not need hdf5. That is only needed for the recording and replay tools
- libqwt-qt5-dev is only needed for building 1 specific eCAL Mon Plugin. Is it the goal to build all the tools, too?
- If all the tools are meant to be built as well, the dependency list will likely be much longer

package.xml
```diff
@@ -0,0 +1,24 @@
...
+
+    <export>
+        <build_type>cmake</build_type>
```
FlorianReimold, line 22: I don't know much about the package.xml, but does it contain information on how to actually build the lib? Like CMake options / a command line?

package.xml
wep21: In ros build farm, a build job is executed without any options, so I added the libraries which is necessary to build successfully without any options.

package.xml
wep21: No, we need to specify options via `colcon` or `cmake` or write them in CMakeLists.txt directly. `package.xml`  doesn't contain the information.

package.xml
wep21: > Is there any way to make this variable? For instance making it check git tags, creating this file with cmake etc.

I'm sorry, but I don't have any idea to do so. I usually update version in `package.xml` and add git tag at the same time via command line tool named [catkin_preprare_release](http://wiki.ros.org/bloom/Tutorials/ReleaseCatkinPackage).

package.xml
FlorianReimold: Hm, not being able to specify cmake options doesn't sound optimal at all. Here is a list of dependencies that I pulled from the source package of our PPA (Jammy). All of them should be required for default build on Ubuntu 22.04.

```qtbase5-dev (>= 5.9), libqwt-qt5-dev (>= 6.1), libhdf5-dev, libprotobuf-dev (>= 3.0), libprotoc-dev (>= 3.0), protobuf-compiler (>= 3.0), libcurl4-openssl-dev, libyaml-cpp-dev```

KerstinKeller:
It seems that this is what a `colcon.pkg` file is for?
This is an examplary file https://github.com/eProsima/Micro-CDR/blob/1c712a08cf5571af0435bf1d881075e813c09513/colcon.pkg#L5,  also see this discussion https://github.com/colcon/colcon.readthedocs.org/issues/44

As Florian said, building all of eCAL is probably not what is needed in the ROS context. eCAL has a lot of options (and what we consider a "default" package, might entirely not be what is useful in the ROS context).

wep21:
@KerstinKeller cc @FlorianReimold
Thank you for your comment. I added `colcon.pkg` to build minimal libraries and remove some dependencies from `package.xml`.

package.xml
wep21: The system dependencies in package.xml is mainly defined [here](https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml).
These dependecies are for some linux distros, but I guess we can also define another platform dependencies like [this](https://github.com/ros/rosdistro/blob/master/rosdep/osx-homebrew.yaml).

FlorianReimold (changes requested):
Could you still correct the version number? The current version of the master branch is `5.12.0-nightly2`

wep21:
@FlorianReimold I changed the version to `5.12.0-nightly2`, but the version in package.xml should be in the pattern `(0|[1-9][0-9]*)(.(0|[1-9][0-9]*)){2}`, so I would like you to update the version in package.xml to satisfy the pattern when you create a new stable version tag.

FlorianReimold:
Hm, maybe then just have `5.12.0`? I am just trying to not make people think that the master branch would represent any 5.11 release. For releases, we always create support branches like `support/v5.11`, where we maintain an ABI compatible release. After we have created a new support branch, we then increase the version on our master branch to a new nightly version, so if we create a new `support/v5.12` branch, we will then tag the master `v5.13.0-nightly`, so you directly see that you are in fact **not** using a 5.12 compatible release, if you are working on master. On the 5.12 branche we will then create a release candidate and later the final release, which is then just tagged `v5.12.0`.

Sad to see, that the package xml does not support semantic versioning pre-release identifiers, but maybe it is then better to directly increase the version number to the next minor version (5.12.0 in this case)?

FlorianReimold (approved):

PR #1127, opened 2023-06-15
hotfix/io-memfile-test

rex-schilasky:
**Pull request type**

 

 

- [X] Bugfix

**What is the new behavior?**

Added missing memory file destruction in memfile gtest.

- [X] No


KerstinKeller (approved):
Interesting! I would still vote to adress this in the destructor?

rex-schilasky:
> Interesting! I would still vote to adress this in the destructor?
We need to check this additionally to avoid side effects.

Issue #1132, opened 2023-06-27
eCAL Sys: Start and monitor tasks running in Docker Containers


FlorianReimold:
### Context

current situation:

Container runs with ipc enabled which relies that the container name is equal to the host name. That implies that we can't run an ecalsys remote client in the container which starts some tasks and forward the state of these to the hosts ecalsys. 

Additionally we can't stop algos in the container and therefore need to kill relaunch the container to stop / restart the task within the container.

We would like to have a solution, how to start tasks inside a docker container with remote control from the host, ipc (shared memory) and ecal state visible in the hosts ecalsys.

### Proposal

Ideas: 
1. One eCAL Sys Client running on the host
  - the Client must be able to start tasks inside docker containers now, which is way more complicated than just loading and executing an executable. It probably involves executing a script.
  - Establish a feedback mechanism from Docker containers to Ecalsys, enabling effective communication and information exchange between the two systems. 

2. Multiple eCAL Sys Clients (one per container)
  - The preferred mechanism
  - eCAL Sys must be able to address these clients individually, which is not possible at the moment, as all containers and the host share the same hostname
    - Either docker containers must have different hostnames and still be able to communicate via SHM
    - Or the eCAL service API must be able to address different Clients on the same host indivudually
      - _I think this should work nevertheless, as there is no technical reason why this shouldn't work. It's just artifically disabled by eCAL at the moment_
      - The new Service implementation #1072 is starting to solve this issue already
      - The feature must still be exposed to the eCAL API
      - It must also be configurable in eCAL Sys, where currently the hostname is the only "address" which is expected to be unique (but isn't any more, when all docker containers share the same hostname and run their own eCAL Sys Clients).

KerstinKeller:
As discussed, I suggest rather to finetune the logic behind matching "local" shm subscriptions.
We can define a so called "SHM namespace". Two processes that share the "SHM namespaces" are considered to be able to communicate based on "intrahost" communication.

Right now, the "SHM namespace" is the Hostname. However, it could be anything else, and entities could configure it via ecal.ini / API.

This would allow docker images to have different hostnames, thus allowing eCAL Sys clients to run on them and to monitor the applications running in these docker containers.

I prefer this solution over the service API approach, because it feels cleaner to me.

PR #1143, opened 2023-07-17
Introducing the attribute "host group name" that allows interprocess communication across host borders

hannemn:
**Pull request type**

 

 

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

Issue Number: #1132 

**What is the new behavior?**

The PR introduces a new setting called `host group name` to enable interprocess communication between docker containers with different host names. The `host group name` setting in ecal.ini can be configured to an abitrary string by the user. Hosts or rather containers that shares the same `host group name` are allowed to exchange data via ipc mechanisms even if there are not on the same (virtual) host. By default the local `host name` will be set as `host group name` if the respective configuration section was left empty in ecal.ini.

- [ ] Yes
- [X] No

**Other information**


KerstinKeller (commented):
Do you really want to put it in the experimental ini file group?
Everything else (at least the API) ist stable / fix anyways...
I think I'd just put it in the ini in a regular section.

hannemn:
Yes, good point! I was a bit too pessimistic putting it into the productive config section. ;-) I will move it.

FlorianReimold:
I am wondering whether the Host Group Name should actually be in the `ecal.ini`. If using Docker, you will probably want to set the host group name from the outside, e.g. via an environment variable. The same goes for the (actual) hostname, which is also controlled from the outside and not by an `ecal.ini` setting. Having the HostGroupName statically configured in the `ecal.ini` will prevent you from pre-configuring a docker container with libecal_core.so + custom software + ecal.ini, as you would have to know at docker compose time where the docker container is going to run.

KerstinKeller (approved):
Good work, Kristof  👍

PR #1152, opened 2023-07-20
5.12: Reverted mcap changes

FlorianReimold:
The new functionality is not usable, yet and the new API may still change. So we are reverting the changes for 5.12.


KerstinKeller (approved):
😢

PR #1166, opened 2023-07-27
Shorten default shm_monitoring_domain to avoid path length problems o…

KerstinKeller:
…n MacOS.

- [x] Bugfix

Cherry-Pick to:
- 5.12

### DIFF
--- a/ecal/core/cfg/ecal.ini
+++ b/ecal/core/cfg/ecal.ini
@@ -182,7 +182,7 @@ filter_excl               = ^eCALSysClient$|^eCALSysGUI$|^eCALSys$
 ; --------------------------------------------------
 [experimental]
 shm_monitoring_enabled      = false
-shm_monitoring_domain       = ecal_monitoring
+shm_monitoring_domain       = ecal_mon
 shm_monitoring_queue_size   = 1024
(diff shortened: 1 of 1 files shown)


FlorianReimold:
Is that enough? If it is, then this will obviously get my approval

hannemn:
Only as a remark: This option only influences the name prefix of the shared memory files that are used for shared memory **monitoring** and has no effect on the file names where the acutual payloads become tranfered.

FlorianReimold:
Yes, but as I understood, that also was the only issue, right? we already reduced the path length for the payload and event SHM files some time ago

KerstinKeller:
@hannemn are you sure? From the code it seems that every opened memory file for the monitoring will hold the prefix.
Eg it tried to open ecal_monitoring_1394252728 (some ultraling hash) resulting in 33 or 36 chars. Shortening to ecal_mon then gives the a length <31 chars

FlorianReimold (approved):

hannemn:
Okay, I tought the shared memory file names of payload and event are even longer. But if this was fixed already some time ago, then your code adaptions should solve the name length issue good enough.

hannemn (approved):
LGTM

PR #1168, opened 2023-07-27
disable zero copy partial write combined with multibuffering (#1163)

### LINKED ISSUE #1163: ZeroCopy + Multibuffering leads to inconsitent data
### Problem Description

Let's say we have 3 buffers for the same topic:
- 0000
- 0000
- 0000

We now write a 1 at pos 1. We expect a [1000] being published
- 1000 < gets published
- 0000
- 0000

We now write a 1 at pos2. We exepct a [1100] being published
- 1000
- 0100 < gets published
- 0000

So at the moment, multibuffering + zerocopy leads to inconsistent data.

### How to reproduce

This was a theoretical discussion.

### How did you get eCAL?

Other

### Environment

- eCAL 5.12.0-rc.1 / master

### eCAL System Information

_No response_

rex-schilasky:
**Pull request type**

- [X] Bugfix

See issue #1163

Fixes #1163

**What is the new behavior?**
If the publisher multibuffering feature is activated (`CPublisher::ShmSetBufferCount`), multiple memory files are used as ring buffers and cyclically written.

The new zero copy (`CPublisher::ShmEnableZeroCopy`) mode on the other hand allows to write partial content (modify the the content) of a memory file, this leads to much higher data throughput if just a few byte of a payload needs to be changed on every publication.

This partial / differential write makes no sense if multiple memory files are handled by a single publisher. In this case, each of these files must be written completely to ensure consistent payload data. So, the partial write variant is now blocked in the described multibuffering mode.

- [ ] Yes
- [X] No

**Cherry-pick to:**
- 5.12

### DIFF
--- a/ecal/core/src/io/ecal_memfile.cpp
+++ b/ecal/core/src/io/ecal_memfile.cpp
@@ -314,7 +314,7 @@ namespace eCAL
     }
   }
 
-  size_t CMemoryFile::WritePayload(CPayloadWriter& payload_, const size_t len_, const size_t offset_)
+  size_t CMemoryFile::WritePayload(CPayloadWriter& payload_, const size_t len_, const size_t offset_, bool force_full_write_ /*= false*/)
   {
     if (!m_created) return(0);
 
@@ -322,7 +322,7 @@ namespace eCAL
     if (GetWriteAddress(wbuf, len_ + offset_) != 0u)
     {
       // (re)write complete buffer
-      if (!m_payload_initialized)
+      if (!m_payload_initialized || force_full_write_)
       {
         bool const success = payload_.Write(static_cast<char*>(wbuf) + offset_, len_);
         if (!success)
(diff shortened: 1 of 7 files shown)


FlorianReimold (approved):

chengguizi:
https://eclipse-ecal.github.io/ecal/stable/advanced/layers/shm_zerocopy.html#combining-zero-copy-and-multibuffering

@rex-schilasky May I know if the statement in the doc is still true? The documentation seems to suggest even WriteFull is disabled when using multibuffering, however, in actual code, seems in eCAL 6, and mentioned in this issue, only WriteModified is disabled in multi buffering mode.

```
  bool CDataWriterSHM::Write(CPayloadWriter& payload_, const SWriterAttr& attr_)
  {
    // write content
    const bool force_full_write(m_memory_file_vec.size() > 1);
    const bool sent = m_memory_file_vec[m_write_idx]->Write(payload_, attr_, force_full_write);

    // and increment file index
    m_write_idx++;
    m_write_idx %= m_memory_file_vec.size();

    return sent;
  }
```

I intend to use full zero copy mode with multi buffering, without any paritial write

Issue #1145, opened 2023-07-17
Document eCAL 5.12 Features


FlorianReimold:
We need to document some newly introduced features:
- Zero Copy
  - [X] Document Zero Copy Feature @FlorianReimold @rex-schilasky 
  - [X] Add Doxygen Documentation for `CPayloadWriter` @rex-schilasky 
  - [X] CPlayload for the Publisher for 5.12. We should write a section on it in the documentation, especially for end users who have their own publishers. @rex-schilasky @KerstinKeller 
  - [X]  zero copy functionality available if there is only local communication. @rex-schilasky @KerstinKeller 
- [x] Host group feature (maybe modify [docker tutorial](https://eclipse-ecal.github.io/ecal/advanced/ecal_in_docker.html)) @hannemn @KaanConti 
- Release Documentation (Branch started here: [feature/doc_5_12](https://github.com/eclipse-ecal/ecal/tree/feature/doc_5_12/))
  - [x] Changelog
  - [x] Highlights
  - [ ] Screenshot / Graphic
  - [x] Compatibility Table

FlorianReimold:
Some doc is here and can probably be copied:
https://github.com/eclipse-ecal/ecal/pull/1082

g-tejas:
Is there a summary of API breaking changes somewhere, like a changelog? Thanks!

FlorianReimold:
We are still working on it and will be publishing it, soon. But there won't be any API breaks. The API got extended with new functions, but nothing old was removed, so it is downwards compatible.

PR #1169, opened 2023-07-29
Core / ZeroCopy: Added Doxygen API documentation and 2 new c++ samples demonstrating usage of low level zero copy API

rex-schilasky:
**Pull request type**

- [X] Feature
- [X] Documentation content changes

We missed samples to demonstrate low level zero copy API and the CPayloadWriter class was not documented (Doxygen).

Issue Number: #1145

**What is the new behavior?**
Two new binary send / rec zero copy samples added, CPayloadWriter class documented.

- [ ] Yes
- [X] No

**Cherry-Pick to**
- 5.12

### DIFF
--- a/ecal/core/include/ecal/ecal_payload_writer.h
+++ b/ecal/core/include/ecal/ecal_payload_writer.h
@@ -139,7 +139,8 @@ namespace eCAL
      *
      * @return True if the copy operation is successful, false otherwise.
      */
-    bool Write(void* buffer_, size_t size_) override {
+    bool Write(void* buffer_, size_t size_) override
+    {
       if (buffer_ == nullptr) return false;
       if (size_ < m_size) return false;
       if (m_buffer == nullptr) return false;
--- a/ecal/core/include/ecal/ecal_payload_writer.h
+++ b/ecal/core/include/ecal/ecal_payload_writer.h
@@ -29,56 +29,139 @@
 
 namespace eCAL
 {
-  // base payload writer class to allow zero copy memory operations
-  // 
-  // the `Write`and `Update` calls may operate on the target memory file directly (zero copy mode)
+  /**
+   * @brief Base payload writer class to allow zero copy memory operations.
+   *
+   * This class serves as the base class for payload writers, allowing zero-copy memory
+   * operations. The `Write` and `Update` calls may operate on the target memory file
+   * directly in zero-copy mode.
+   */
   class CPayloadWriter
   {
   public:
+    /**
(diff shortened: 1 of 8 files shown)


FlorianReimold:
If I see correctly, this PR basically contains 2 changes:

1. It adds the samples
2. More importantly: It adds doxygen documentation to the payload writer (public API)

The latter change requires this commit to be cherry-picked to 5.12.

@rex-schilasky can you confirm this?

rex-schilasky:
> If I see correctly, this PR basically contains 2 changes:
> 
> 1. It adds the samples
> 2. More importantly: It adds doxygen documentation to the payload writer (public API)
> 
> The latter change requires this commit to be cherry-picked to 5.12.
> 
> @rex-schilasky can you confirm this?

This is all correct. The doxygen documented `CPayloadWriter` class and both samples demonstrating the usage of this class should go to 5.12.

FlorianReimold (changes requested):

ecal/core/include/ecal/ecal_payload_writer.h
```diff
@@ -29,56 +29,140 @@
...
+     * @return The size of the memory that needs to be copied.
+     */
     size_t GetSize() override { return m_size; };
-  
+
   private:
-    const void* m_buffer = nullptr;
-    size_t      m_size   = 0;
+    const void* m_buffer = nullptr;  ///< Pointer to the buffer containing the data to be written.
+    size_t      m_size = 0;          ///< Size of the data to be written.
```
FlorianReimold, line 165: Is this really the "data to be written"? If so, why isn't this just a temporary locally scoped variable? Or is this capacity of the memory file, as the `GetSize()` function says?

Either way, I guess either the documentation of the variable or the `GetSize()` function is wrong

FlorianReimold (changes requested):
After reviewing the Doxygen I don't really get what "data to be written" means here. I assume that this term refers to the size of the underlying memory file?

ecal/core/include/ecal/ecal_payload_writer.h
rex-schilasky: The `CBufferPayloadWriter` is a kind of primitive wrapper of the classic `CBufferPayloadWriter(const void* const buffer_, size_t size_)` Send API. Basically it stores the arguments `buffer_` and `size_` into the internals of `CBufferPayloadWriter` and does the memcpy action in the `Write` call.
`GetSize` reports the needed size correctly, Write is just doing the copy.

FlorianReimold (approved):

PR #1175, opened 2023-07-31
GitHub: Radically shortened the PR template

FlorianReimold:
**Description**

The old issue template was very long. It was hard to fill out correctly and thus often contained wrong information. Thousands of checkboxes distracted from what is actually important.

The most important thing for a PR should always be: What does it change.

I also added a section for related issues and cherry-pick. The latter is meant for giving myself a better overview over what still needs to be cherry-picked to which branch and what is master-only. Usually the developer knows that when submitting the PR.

**Related issues**
- _none_

**Cherry-pick to**:
- _none_

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,38 +1,8 @@
-**Pull request type**
+**Description**
+<!-- What does your PR change? -->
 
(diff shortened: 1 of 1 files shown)


rex-schilasky (approved):
We can start with this new minimalistic template and adapt later if we miss something.

PR #1197, opened 2023-09-12
Fixes bug regarding unregistration introduced with #1086.

### LINKED ISSUE #1196: #1086 Breaks unregistration via monitoring expiration
### Problem Description

Prior to #1086, the `m_loc_sub_map` variable in ecal_writer.cpp stored the process id. With this PR it now stores a `topic_key` which is a combination of the `process_id` and `topic_id`.

https://github.com/eclipse-ecal/ecal/blob/95480add0f802fa9d8d89bb39eddb7cd184f7f7b/ecal/core/src/readwrite/ecal_writer.cpp#L663

For regular Unregistration, (`RemoveLocSubscription`) this is not a problem, however, if Removed via `RefreshRegistration`, the content of the removed items of the `m_loc_sub_map` are still interpreted as "process id only", and passed to `RemLocConnection` function (which expects the process id, not the combination).

https://github.com/eclipse-ecal/ecal/blob/95480add0f802fa9d8d89bb39eddb7cd184f7f7b/ecal/core/src/readwrite/ecal_writer.cpp#L771-L784

### How to reproduce

Run a publisher and a subscriber. Kill the Subscriber task. See that the Subscriber will never be unregistered properly.

### How did you get eCAL?

I don't know

### Environment

Current Master (5.12+)

### eCAL System Information

_No response_

KerstinKeller:
Introduces two new structs (SLocalSubscriptionInfo and SExternSubscriptionInfo) and uses them for applying / unapplying registrations. They are also stored in the ExpiredMap when topic are no longer visible on monitoring layer.

### Related issues
Fixes #1196

### Cherry-pick to
- 5.12

### DIFF
--- a/ecal/core/src/pubsub/ecal_pubgate.cpp
+++ b/ecal/core/src/pubsub/ecal_pubgate.cpp
@@ -114,9 +114,10 @@ namespace eCAL
 
     const auto& ecal_sample = ecal_sample_.topic();
     const std::string& topic_name = ecal_sample.tname();
-    const std::string& topic_id   = ecal_sample.tid();
+    CDataWriter::SLocalSubscriptionInfo subscription_info;
+    subscription_info.topic_id = ecal_sample.tid();
+    subscription_info.process_id = std::to_string(ecal_sample.pid());
     SDataTypeInformation topic_information{ eCALSampleToTopicInformation(ecal_sample_) };
-    const std::string  process_id = std::to_string(ecal_sample.pid());
 
     std::string reader_par;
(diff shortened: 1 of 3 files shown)


FlorianReimold:
Why only cherry-pick this to 5.12? Was it introduced with 5.12 and didn't exist in 5.11?

KerstinKeller:
Yes as mentioned in the issue, the problem was introduced with the mentioned PR, which was merged for the 5.12 release.

FlorianReimold (approved):

PR #1198, opened 2023-09-13
Core: Fixed a bug that may have caused drops in the SHM Layer

### LINKED ISSUE #1194: SHM Acks degrade over time
### Problem Description

eCAL seems to not handle SHM Acknowledgements properly. Even in a system that has enough resources, everything seems to degrade over time and the publisher will run into SHM ack timouts.

- A first analysis shows, that if the Subscriber is not able to get memfile access in 5ms (hardcoded!) It will omit sending the shm ack and let the publisher wait for the timeout.

    https://github.com/eclipse-ecal/ecal/blob/95480add0f802fa9d8d89bb39eddb7cd184f7f7b/ecal/core/src/io/ecal_memfile_pool.cpp#L166

- Also, the list of events hold by the publisher is never cleaned up, so when subscribers go away, the publisher will still try to receive their events.

    - **Hypothesis**: The following TODO has something to do with that:
        https://github.com/eclipse-ecal/ecal/blob/95480add0f802fa9d8d89bb39eddb7cd184f7f7b/ecal/core/src/readwrite/ecal_writer_shm.cpp#L193-L205

### How to reproduce

1. Set Ack timeout to 500 in `ecal.ini`
2. Compile the [ecal-perftool](https://github.com/FlorianReimold/ecal-perftool)
3. Start  1 publisher
    ```
    ecal_perftool pub topic 25 1024000
    ```
4. Start 4 susbscribers
    ```
    ecal_perftool sub topic 10
    ```
5. Start 5 subscribers that slow down the system a little bit, but not too much
    ```
    ecal_perftool sub topic 100
    ```

The publishing freq should be 10 Hz  (due to the 100ms sub, that slows everything down), but after some time it will go down to 2Hz, which is the ack timeout:
image

### How did you get eCAL?

Ubuntu PPA (apt-get)

### Environment

eCAL 5.11.5
Ubuntu 20.04

### eCAL System Information

```text
------------------------- SYSTEM ---------------------------------
Version                  : v5.11.5 (2023-07-24 12:19:00 +0200)
Platform                 : linux

------------------------- CONFIGURATION --------------------------
Default INI              : /etc/ecal/ecal.ini

------------------------- NETWORK --------------------------------
Host name                : florian-ubuntu20
Network mode             : local
Network ttl              : 2
Network sndbuf           : 5 MByte
Network rcvbuf           : 5 MByte
Multicast group          : 239.0.0.1
Multicast mask           : 0.0.0.15
Multicast ports          : 14000 - 14010
Multicast join all IFs   : off
Bandwidth limit (udp)    : not limited

------------------------- TIME -----------------------------------
Synchronization realtime : "ecaltime-localtime"
Synchronization replay   :
State                    :  synchronized
Master / Slave           :  Master
Status (Code)            : "everything is fine." (0)

------------------------- PUBLISHER LAYER DEFAULTS ---------------
Layer Mode INPROC        : off
Layer Mode SHM           : auto
Layer Mode TCP           : off
Layer Mode UDP MC        : auto

------------------------- SUBSCRIPTION LAYER DEFAULTS ------------
Layer Mode INPROC        : on
Layer Mode SHM           : on
Layer Mode TCP           : on
Layer Mode UDP MC        : on
Npcap UDP Reciever       : off
```

FlorianReimold:
### Description
Prior to this commit, SHM data was dropped, if the subscriber wasn't able to access it in 5ms. Now, the subscriber re-tries, until it got access to it, or new data is available.

### Related issues
- Found by debugging #1194, but this fix did not fix the issue.

### Cherry-pick to
- 5.11 (old stable)
- 5.12 (current stable)

### DIFF
--- a/ecal/core/src/io/ecal_memfile_pool.cpp
+++ b/ecal/core/src/io/ecal_memfile_pool.cpp
@@ -36,7 +36,7 @@ namespace eCAL
     m_created(false),
     m_do_stop(false),
     m_is_observing(false),
-    m_timeout_read(0)
+    m_time_of_last_life_signal(std::chrono::steady_clock::now())
   {
   }
 
@@ -137,7 +137,7 @@ namespace eCAL
   {
     if (!m_is_observing) return false;
 
-    m_timeout_read = 0;
+    m_time_of_last_life_signal = std::chrono::steady_clock::now();
     
     return true;
   }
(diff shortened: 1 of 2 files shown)


KerstinKeller (approved):
Let's merge it

PR #1211, opened 2023-10-09
HDF5: Make sure datasets are always closed upon reading...

### LINKED ISSUE #1210: Using ecalhdf5 API causes problems with zero-sized channels
### Problem Description

We have measurements that have zero-sized channels. When trying to access the data from that channel the method just returns false leaving the dataset unreleased thus the HDF5 object is inaccessible.
https://github.com/eclipse-ecal/ecal/blob/87d0adeec5e981e1ff0ec622db7545ac27442b85/contrib/ecalhdf5/src/eh5_meas_file_v2.cpp#L297
Also, this method returns true no matter if the actual closing happened successfully.  https://github.com/eclipse-ecal/ecal/blob/87d0adeec5e981e1ff0ec622db7545ac27442b85/contrib/ecalhdf5/src/eh5_meas_dir.cpp#L96
Probably it would be a good idea to check everywhere if the closing of all the resources is successful.
Also, if it makes sense to add a test using a zero-sized channel.

### How to reproduce

Read a measurement with zero-sized channels using eCAL API, close it and for example try to rename/copy/delete it.

### How did you get eCAL?

Download from Release Page

### Environment

- eCAL 5.11.1
- Windows

### eCAL System Information

_No response_

KerstinKeller:
, even when size 0 data is read.

HDF5MeasDir should return if all files were closed properly.

### Description
Properly Close Opened hdf5 datasets when reading

### Related issues
Fixes #1210

### Cherry-pick to
- 5.11 (old stable)
- 5.12 (current stable)

### DIFF
--- a/contrib/ecalhdf5/src/eh5_meas_dir.cpp
+++ b/contrib/ecalhdf5/src/eh5_meas_dir.cpp
@@ -95,18 +95,20 @@ bool eCAL::eh5::HDF5MeasDir::Open(const std::string& path, eAccessType access /*
 
 bool eCAL::eh5::HDF5MeasDir::Close()
 {
+  bool successfully_closed{ true };
+
   if (access_ == eAccessType::CREATE)
   {
     // Close all existing file writers
     for (auto& file_writer : file_writers_)
     {
-      file_writer.second->Close();
+      successfully_closed &= file_writer.second->Close();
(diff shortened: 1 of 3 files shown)


hannemn (approved):
lgtm!

PR #1214, opened 2023-10-13
bugfix/unregistration memfile event handling

rex-schilasky:
### Description
The new introduced "active register/unregister logic" calls depending on the connection type specific functions like (AddLocConnection/RemLocConnection,AddExtConnection/RemExtConnection) for the SHM and UDP data writer (TCP was just not handled for now).

The RemLocConnection was not called in that context before the new unregister/register logic and is closing the 2 connection events for updating and acknowledging memory file data transfer/receive. This leads to a dead shm connection when the subscriber on the other side is destroyed (triggering RemLocConnection in the publishing process) and created again.

Solution: Events are not closed on that level but are cleaned up together with the memory file resource (CSyncMemFile) properly.

### Related issues
No issue now, will be created soon ..

### Cherry-pick to
- None

### DIFF
--- a/ecal/core/src/readwrite/ecal_writer.cpp
+++ b/ecal/core/src/readwrite/ecal_writer.cpp
@@ -668,8 +668,10 @@ namespace eCAL
     m_loc_subscribed = true;
 
     // add a new local subscription
-    m_writer.udp_mc.AddLocConnection (local_info_.process_id, reader_par_);
-    m_writer.shm.AddLocConnection    (local_info_.process_id, reader_par_);
+    // currently for the shm writer implemented only
+    m_writer.udp_mc.AddLocConnection (local_info_.process_id, local_info_.topic_id, reader_par_);
+    m_writer.shm.AddLocConnection    (local_info_.process_id, local_info_.topic_id, reader_par_);
+    m_writer.tcp.AddLocConnection    (local_info_.process_id, local_info_.topic_id, reader_par_);
 
 #ifndef NDEBUG
     // log it
@@ -686,8 +688,10 @@ namespace eCAL
     }
 
     // remove a local subscription
-    m_writer.udp_mc.RemLocConnection (local_info_.process_id);
-    m_writer.shm.RemLocConnection    (local_info_.process_id);
+    // currently all these functions are not implemented (there is no need to react on)
+    m_writer.udp_mc.RemLocConnection (local_info_.process_id, local_info_.topic_id);
+    m_writer.shm.RemLocConnection    (local_info_.process_id, local_info_.topic_id);
+    m_writer.tcp.RemLocConnection    (local_info_.process_id, local_info_.topic_id);
 
 #ifndef NDEBUG
     // log it
@@ -708,8 +712,10 @@ namespace eCAL
     m_ext_subscribed = true;
 
     // add a new external subscription
-    m_writer.udp_mc.AddExtConnection (external_info_.host_name, external_info_.process_id, reader_par_);
-    m_writer.shm.AddExtConnection    (external_info_.host_name, external_info_.process_id, reader_par_);
+    // currently all these functions are not implemented (there is no need to react on)
+    m_writer.udp_mc.AddExtConnection (external_info_.host_name, external_info_.process_id, external_info_.topic_id, reader_par_);
+    m_writer.shm.AddExtConnection    (external_info_.host_name, external_info_.process_id, external_info_.topic_id, reader_par_);
+    m_writer.tcp.AddExtConnection    (external_info_.host_name, external_info_.process_id, external_info_.topic_id, reader_par_);
 
 #ifndef NDEBUG
     // log it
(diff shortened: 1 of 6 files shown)


ecal/core/src/readwrite/ecal_writer_base.h
```diff
@@ -48,11 +48,11 @@ namespace eCAL
...
 
-    virtual bool AddLocConnection(const std::string& /*process_id_*/, const std::string& /*conn_par_*/) { return false; };
-    virtual bool RemLocConnection(const std::string& /*process_id_*/) { return false; };
+    virtual bool AddLocConnection(const std::string& /*process_id_*/, const std::string& /*topic_id_*/, const std::string& /*conn_par_*/) { return false; };
+    virtual bool RemLocConnection(const std::string& /*process_id_*/, const std::string& /*topic_id_*/) { return false; };
 
-    virtual bool AddExtConnection(const std::string& /*host_name_*/, const std::string& /*process_id_*/, const std::string& /*conn_par_*/) { return false; };
-    virtual bool RemExtConnection(const std::string& /*host_name_*/, const std::string& /*process_id_*/) { return false; };
+    virtual bool AddExtConnection(const std::string& /*host_name_*/, const std::string& /*process_id_*/, const std::string& /*topic_id_*/, const std::string& /*conn_par_*/) { return false; };
+    virtual bool RemExtConnection(const std::string& /*host_name_*/, const std::string& /*process_id_*/, const std::string& /*topic_id_*/) { return false; };
```
KerstinKeller, line 56: Why did you change the function signatures? Are the arguments actually being used somewhere?

ecal/core/src/readwrite/ecal_writer.cpp
```diff
@@ -764,22 +772,15 @@ namespace eCAL
...
 
       m_loc_subscribed = !m_loc_sub_map.empty();
       m_ext_subscribed = !m_ext_sub_map.empty();
     }
 
-    for(const auto& loc_sub : *loc_timeouts)
-    {
-      m_writer.shm.RemLocConnection(loc_sub.process_id);
-    }
-
```
KerstinKeller, line 782: This is not the "goodbye" message but happens on registration timeout?

ecal/core/src/readwrite/ecal_writer_shm.cpp
```diff
@@ -190,28 +190,6 @@ namespace eCAL
...
-      // TODO: Fix this in 'CSyncMemoryFile::Disconnect' to handle event resources properly.
-      if (std::to_string(eCAL::Process::GetProcessID()) != process_id_)
-      {
-        ret_state &= memory_file->Disconnect(process_id_);
-      }
-    }
-
-    return ret_state;
-  }
-
```
KerstinKeller, line 214: It feels a bit wrong that connections cannot be removed again. So the resources stay open until the publisher is closed? Is that what happens? Also they continue being signalled, etc?

And btw, I guess you have addd line 206 to prevent the testcase to fail. see the huge comment. Maybe it was just not working at all on Posix...

ecal/core/src/readwrite/ecal_writer.cpp
```diff
@@ -686,8 +688,10 @@ namespace eCAL
     }
 
     // remove a local subscription
-    m_writer.udp_mc.RemLocConnection (local_info_.process_id);
-    m_writer.shm.RemLocConnection    (local_info_.process_id);
+    // currently all these functions are not implemented (there is no need to react on)
+    m_writer.udp_mc.RemLocConnection (local_info_.process_id, local_info_.topic_id);
+    m_writer.shm.RemLocConnection    (local_info_.process_id, local_info_.topic_id);
+    m_writer.tcp.RemLocConnection    (local_info_.process_id, local_info_.topic_id);
```
KerstinKeller, line 695: Adding the tcp `RemLocConnection` / `AddExtConnection` should preferably be done in a different PR as it is not related to the issue that is being fixed here.

KerstinKeller (commented):
For me, a lot of questions (concerning) the original design, still remain.

I find it a bit questionable, to keep all signalling resources open until the memory file itself is destroyed.
Couldn't it lead to the same problems on subscriber side, e.g. if the subscriber side destroys the events (e..g after 60 second timeout) and then creates a new subscriber?
(E.g. now we made sure events are not closed on sender side, because in the testcase they stayed open on subscriber side. What if they are closed/reopened on subscriber side... in that case communication can't be established, either?)

It feels to me, like we might have just shifted the problem.
Maybe it's worth to take a look at the underlying NamedEvents.

KerstinKeller (changes requested):
Now it doesn't work in the case that the subscriber is calling `finalize()` and `initialize()`

ecal/core/src/readwrite/ecal_writer_base.h
rex-schilasky: They are not used yet, but IF any layer wants to intercept that information it makes sense to provide host_name, process_id, topic_id to identify exactly the connection the is added or removed. The old signature supported just SHM with the needed information process_id.

ecal/core/src/readwrite/ecal_writer.cpp
rex-schilasky: Yes it's not the goodbye message and no more needed at all. It closed the connected events for the memory file that was connected to this shm writer. This is now done since many version in the destruction of the CSyncMemFile.

ecal/core/src/readwrite/ecal_writer_shm.cpp
rex-schilasky: No it's no more signalling, it is just holding the resources for MemFile + Events. The publisher (writer) knows the number of connections local/extern all the time.

ecal/core/src/readwrite/ecal_writer.cpp
rex-schilasky: I will mark this PR as a cleanup and we do the fix in another one.

rex-schilasky:
> Now it doesn't work in the case that the subscriber is calling `finalize()` and `initialize()`

Yes, and I assume that this issue is existing independent from that "fix". I will check this again.

rex-schilasky:
mixed up bugfixes and code cleanups + did not really fix the issue

PR #1216, opened 2023-10-16
cleanup add/rem datawriter connection logic

rex-schilasky:
### Description
- `RemLocConnection` call for SHM removed from ecal_writer.cpp
  - this call is unnecessary (events are closed on `CSyncMemFile Destruction/Disconnect`)
  - this call disconnects to a process even there are still additional subscriber for the same topics in this process
- calls `AddLocConnection`/`AddExtConnection`, `RemLocConnection`/`RemExtConnection `for all layers (including tcp) even function are unimplemented
- changed signature for all `AddXXXConnection`/`RemXXXConnection `functions to provide now unambiguous information host name, process id, topic id and not only process id

### Cherry-pick to
- _none_

### DIFF
--- a/ecal/core/src/readwrite/ecal_writer.cpp
+++ b/ecal/core/src/readwrite/ecal_writer.cpp
@@ -668,8 +668,9 @@ namespace eCAL
     m_loc_subscribed = true;
 
     // add a new local subscription
-    m_writer.udp_mc.AddLocConnection (local_info_.process_id, reader_par_);
-    m_writer.shm.AddLocConnection    (local_info_.process_id, reader_par_);
+    m_writer.udp_mc.AddLocConnection (local_info_.process_id, local_info_.topic_id, reader_par_);
+    m_writer.shm.AddLocConnection    (local_info_.process_id, local_info_.topic_id, reader_par_);
+    m_writer.tcp.AddLocConnection(local_info_.process_id, local_info_.topic_id, reader_par_);
 
 #ifndef NDEBUG
     // log it
@@ -686,8 +687,9 @@ namespace eCAL
     }
 
     // remove a local subscription
-    m_writer.udp_mc.RemLocConnection (local_info_.process_id);
-    m_writer.shm.RemLocConnection    (local_info_.process_id);
+    m_writer.udp_mc.RemLocConnection (local_info_.process_id, local_info_.topic_id);
+    m_writer.shm.RemLocConnection    (local_info_.process_id, local_info_.topic_id);
+    m_writer.tcp.RemLocConnection    (local_info_.process_id, local_info_.topic_id);
 
 #ifndef NDEBUG
     // log it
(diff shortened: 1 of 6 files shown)


ecal/core/src/readwrite/ecal_writer_base.h
```diff
@@ -48,11 +48,11 @@ namespace eCAL
     virtual bool SetQOS(const QOS::SWriterQOS& qos_) { m_qos = qos_; return true; };
     QOS::SWriterQOS GetQOS() { return(m_qos); };
 
-    virtual bool AddLocConnection(const std::string& /*process_id_*/, const std::string& /*conn_par_*/) { return false; };
-    virtual bool RemLocConnection(const std::string& /*process_id_*/) { return false; };
+    virtual bool AddLocConnection(const std::string& /*process_id_*/, const std::string& /*topic_id_*/, const std::string& /*conn_par_*/) { return false; };
+    virtual bool RemLocConnection(const std::string& /*process_id_*/, const std::string& /*topic_id_*/) { return false; };
```
KerstinKeller, line 52: Why is the default behavior to return `false`? 
Since it shm doesn't override `RemLocConnection` any more, it will now return `false` instead of `true`.

ecal/core/src/readwrite/ecal_writer.cpp
```diff
@@ -726,8 +729,9 @@ namespace eCAL
     }
 
     // remove external subscription
-    m_writer.udp_mc.RemExtConnection (external_info_.host_name, external_info_.process_id);
-    m_writer.shm.RemExtConnection    (external_info_.host_name, external_info_.process_id);
+    m_writer.udp_mc.RemExtConnection (external_info_.host_name, external_info_.process_id, external_info_.topic_id);
+    m_writer.shm.RemExtConnection    (external_info_.host_name, external_info_.process_id, external_info_.topic_id);
+    m_writer.tcp.RemExtConnection    (external_info_.host_name, external_info_.process_id, external_info_.topic_id);
   }
```
KerstinKeller, line 735: Why do the functions have return values, if they are not checked? Just make the function `void`?

ecal/core/src/readwrite/ecal_writer.cpp
```diff
@@ -668,8 +668,9 @@ namespace eCAL
     m_loc_subscribed = true;
 
     // add a new local subscription
-    m_writer.udp_mc.AddLocConnection (local_info_.process_id, reader_par_);
-    m_writer.shm.AddLocConnection    (local_info_.process_id, reader_par_);
+    m_writer.udp_mc.AddLocConnection (local_info_.process_id, local_info_.topic_id, reader_par_);
+    m_writer.shm.AddLocConnection    (local_info_.process_id, local_info_.topic_id, reader_par_);
+    m_writer.tcp.AddLocConnection(local_info_.process_id, local_info_.topic_id, reader_par_);
```
KerstinKeller, line 674: Why do the functions have return values, if they are not checked? Just make the function `void`?

KerstinKeller (approved):
In general it looks good, but see my comments. If nothing is done with the return values, let's just make the functions return `void`.

KerstinKeller (approved):

PR #1222, opened 2023-10-23
new internal event functions with ownership handling (ownership used for linux implementation only)

### LINKED ISSUE #1218: broken pub/sub connection by dynamic recreation of a publisher or subscriber
### Problem Description

After introduction of the new fast unregistration mode (#1086) shm synchronization events between datawriter (sending process) and memfile observer (receiving process) are destroyed (shm_unlink) either by a stopped publisher or subscriber even the event is still used by "the other side" (either the shm data writer or the memfile observer).

This leads to a broken connection if a new publisher/subscriber is created dynamically using the same topic name again.

### How to reproduce

1. Start ecal_sample_person_snd
2. Start this adapted ecal_sampl_person_rec variant

```cpp
int main(int argc, char **argv)
{
  // initialize eCAL API
  eCAL::Initialize(argc, argv, "person subscriber");

  std::cout << std::endl << "FIRST RUN" << std::endl << std::endl; eCAL::Process::SleepMS(1000);
  {
    eCAL::protobuf::CSubscriber<pb::People::Person> sub("person");
    sub.AddReceiveCallback([](const char* /*topic_*/, const pb::People::Person& person_, long long /*time_*/, long long /*clock_*/, long long /*id_*/){std::cout << "person id : " << person_.id() << std::endl;});
    eCAL::Process::SleepMS(3000);
  }

  std::cout << std::endl << "SECOND RUN" << std::endl << std::endl; eCAL::Process::SleepMS(1000);
  {
    eCAL::protobuf::CSubscriber<pb::People::Person> sub("person");
    sub.AddReceiveCallback([](const char* /*topic_*/, const pb::People::Person& person_, long long /*time_*/, long long /*clock_*/, long long /*id_*/){std::cout << "person id : " << person_.id() << std::endl;});
    eCAL::Process::SleepMS(3000);
  }

  // finalize eCAL API
  eCAL::Finalize();

  return(0);
}
```

3. In the ´SECOND RUN´ the subscriber will not receive any data.

### How did you get eCAL?

I don't know

### Environment

- eCAL: 5.12
- OS: Ubuntu 20.04

### eCAL System Information

_No response_

rex-schilasky:
### Description
Introducing ownership flag for named event creation.

### Related issues
Fixes #1218

### Cherry-pick to
- 5.12 (current stable)

### DIFF
--- a/ecal/core/include/ecal/ecal_event.h
+++ b/ecal/core/include/ecal/ecal_event.h
@@ -44,9 +44,9 @@ namespace eCAL
   ECAL_API bool gOpenNamedEvent(eCAL::EventHandleT* event_, const std::string& event_name_, bool ownership_);
 
   /**
-   * @brief Open a named event with ownership.
+   * @brief Open an unnamed event.
    *
-   * @param [out] event_       Returned event struct.
+   * @param [out] event_  Returned event struct.
    *
    * @return  True if succeeded.
   **/
--- a/ecal/core/include/ecal/ecal_event.h
+++ b/ecal/core/include/ecal/ecal_event.h
@@ -26,10 +26,32 @@
 
 #include <ecal/ecal_os.h>
 #include <ecal/ecal_eventhandle.h>
+#include <ecal/ecal_deprecate.h>
+
 #include <string>
 
 namespace eCAL
 {
+  /**
+   * @brief Open a named event with ownership.
+   *
+   * @param [out] event_       Returned event struct.
+   * @param       event_name_  Event name.
+   * @param       ownership_   Event is owned by the caller and will be destroyed on CloseEvent
+   *
+   * @return  True if succeeded.
+  **/
+  ECAL_API bool gOpenNamedEvent(eCAL::EventHandleT* event_, const std::string& event_name_, bool ownership_);
+
+  /**
+   * @brief Open a named event with ownership.
+   *
+   * @param [out] event_       Returned event struct.
+   *
+   * @return  True if succeeded.
+  **/
+  ECAL_API bool gOpenUnnamedEvent(eCAL::EventHandleT* event_);
+
   /**
    * @brief Open a named or unnamed event.
    *
@@ -38,6 +60,7 @@ namespace eCAL
    *
    * @return  True if succeeded.
   **/
+  ECAL_DEPRECATE_SINCE_5_13("Use either gOpenNamedEvent or gOpenUnnamedEvent")
   ECAL_API bool gOpenEvent(eCAL::EventHandleT* event_, const std::string& event_name_ = "");
 
   /**
--- a/ecal/core/include/ecal/ecal_event.h
+++ b/ecal/core/include/ecal/ecal_event.h
@@ -26,32 +26,10 @@
 
 #include <ecal/ecal_os.h>
 #include <ecal/ecal_eventhandle.h>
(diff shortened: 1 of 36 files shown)


ecal/core/include/ecal/ecal_event.h
```diff
@@ -26,10 +26,32 @@
...
+   * @param       event_name_  Event name.
+   * @param       ownership_   Event is owned by the caller and will be destroyed on CloseEvent
+   *
+   * @return  True if succeeded.
+  **/
+  ECAL_API bool gOpenNamedEvent(eCAL::EventHandleT* event_, const std::string& event_name_, bool ownership_);
+
+  /**
+   * @brief Open a named event with ownership.
+   *
```
KerstinKeller, line 48: The comment is wrong. However, do unnamed Events have ownership? How do unnamed events communicate?

ecal/core/src/ecal_event.cpp
```diff
@@ -302,9 +321,10 @@ namespace eCAL
   class CNamedEvent
   {
   public:
-    explicit CNamedEvent(const std::string& name_) :
+    explicit CNamedEvent(const std::string& name_, bool ownership_) :
```
KerstinKeller, line 324: You don't have to change it, but I would like to make a remark. I strongly dislike the `bool` for ownership vs non-ownership.
This should be two different classed, e.g. `OwnedNamedEvent` and `NonOwnedNamedEvent`/`WeakNamedEvent`. 
This has two advantages:
a) To everyone reading the code that is using the events, it's clear from the type, if the event is the owner / creator or not
b) You can use inheritance. You don't need to do stuff like `if(m_owner) ...` which makes code clearer and more readable.

KerstinKeller (commented):
See my comments.
Also I wonder if we should deprecate the whole ecal event API and make it private for eCAL 6. 
There is no reason for it being public. One can think of it as a different library / API, but it has no business in the higher level API.
Maybe this was due to testing purposes? But there are other ways to achieve the same.

testing/ecal/pubsub_test/src/pubsub_test.cpp
```diff
@@ -923,8 +919,6 @@ TEST(IO, ZeroPayloadMessageTCP)
 }
 #endif
 
-#include <ecal/msg/string/publisher.h>
-#include <ecal/msg/string/subscriber.h>
```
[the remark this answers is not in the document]
rex-schilasky: > These includes should be needed but maybe we can move them to the top?

They are there and have been there before ..

ecal/core/src/ecal_event.cpp
rex-schilasky: I just wanted to "just fix the issue" and introduce as little change as possible. Let's redesign this in a separate issue. 

The same for the creation. The behavior is now exact the same as before. The event is created by the first call of OpenEvent independently from the new ownership attribute. The ownership is only plays a role in the destruction phase.

ecal/core/include/ecal/ecal_event.h
rex-schilasky: This is fixed ..

KerstinKeller (approved):
I guess it's the easiest way to fix our problems for now.

rex-schilasky:
Tested with the following setup https://github.com/rex-schilasky/ecal_dynamic_pubsub_test

PR #1227, opened 2023-10-26
switch off monitoring of other hosts if network_mode=local

### LINKED ISSUE #1191: Monitoring traffic from other hosts visibile, even if Network Mode = local
### Problem Description

Even when in `Network mode : local` the monitoring layer shows monitoring traffic from other hosts.
This is undesirable, since an actual communication between the other hosts will not be possible (as no monitoring info will be sent back).

### How to reproduce

Configure eCAL to be in local mode, then start eCAL traffic on another host, which has routes configured. Observe the output in eCAL Monitor.

### How did you get eCAL?

I don't know

### Environment

eCAL: v5.12.0
OS: Windows

### eCAL System Information

_No response_

rex-schilasky:
### Description
If network mode is set to local (`network_enabled=false`) in the ecal global configuration (`ecal.ini`) the traffic of other/external hosts should not monitored.

### Related issues
Fixes #1191 

### Cherry-pick to
- _none_

### DIFF
--- a/ecal/core/src/mon/ecal_monitoring_impl.cpp
+++ b/ecal/core/src/mon/ecal_monitoring_impl.cpp
@@ -31,7 +31,6 @@
 
 #include "../ecal_registration_receiver.h"
 
-
 namespace eCAL
 {
   ////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


rex-schilasky:
Should be solved with another approach on socket level.

PR #1231, opened 2023-10-26
switch off udp receiving of other hosts if network_mode=local

### LINKED ISSUE #1191: Monitoring traffic from other hosts visibile, even if Network Mode = local

rex-schilasky:
### Description
If network mode is set to local (`network_enabled=false`) in the ecal global configuration (`ecal.ini`) the receiving of any kind of udp samples is switched off.

### Related issues
Fixes #1191 

### Cherry-pick to
- _none_

### DIFF
--- a/ecal/core/src/io/udp_receiver_asio.cpp
+++ b/ecal/core/src/io/udp_receiver_asio.cpp
@@ -32,9 +32,9 @@ namespace eCAL
   CUDPReceiverAsio::CUDPReceiverAsio(const SReceiverAttr& attr_) :
     CUDPReceiverBase(attr_),
     m_created(false),
+    m_localhost(attr_.localhost),
     m_broadcast(attr_.broadcast),
     m_unicast(attr_.unicast),
-    m_localhost(attr_.localhost),
     m_socket(m_iocontext)
   {
     if (m_broadcast && m_unicast)
@@ -43,8 +43,11 @@ namespace eCAL
       return;
     }
 
+    // define endpoint
+    asio::ip::udp::endpoint listen_endpoint(asio::ip::udp::v4(), static_cast<unsigned short>(attr_.port));
+    if (m_localhost) listen_endpoint.address(asio::ip::address::from_string("127.0.0.1"));
+
     // create socket
-    const asio::ip::udp::endpoint listen_endpoint(asio::ip::udp::v4(), static_cast<unsigned short>(attr_.port));
     {
       asio::error_code ec;
       m_socket.open(listen_endpoint.protocol(), ec);
@@ -76,7 +79,7 @@ namespace eCAL
       }
     }
 
-    if (!m_unicast)
+    if (!m_localhost && !m_unicast)
     {
       // set loopback option
       const asio::ip::multicast::enable_loopback loopback(attr_.loopback);
@@ -102,14 +105,7 @@ namespace eCAL
     }
 
     // join multicast group
-    if (m_localhost)
-    {
-      AddMultiCastGroup("127.0.0.1");
-    }
-    else
-    {
-      AddMultiCastGroup(attr_.ipaddr.c_str());
-    }
+    AddMultiCastGroup(attr_.ipaddr.c_str());
 
     // state successful creation
     m_created = true;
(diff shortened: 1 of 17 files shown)


rex-schilasky:
THIS approach is not working :-(

KerstinKeller:
So it still receives traffic from other hosts?

rex-schilasky:
> So it still receives traffic from other hosts?

No, but not from your own host either. :-)

FlorianReimold:
I am writing this comment without checking back with the Internet, but I think "127.0.0.1" just isn't a multicast address. There is no multicast group you can join on that. To whom is this suppost to be multicasted to, anyways? there is just 1 machine on 127.0.0.1. You should be able to just unicast that address.

ecal/core/src/io/udp_receiver_asio.cpp
```diff
@@ -196,23 +223,43 @@ namespace eCAL
...
       {
         std::cout << "CUDPReceiverAsio: ipv4 address conversion failed." << std::endl;
       }
     }
 
+    // are we running in 'local host only' receiving mode ?
+    if (m_localhost)
+    {
+      // if this is not a local address, return 0
+      if (!isLocalAddress(m_sender_endpoint.address()))
```
FlorianReimold, line 257: Your definition of a "Local address" has nothing to do with a loopback address (-> localhost)

ecal/core/src/io/udp_receiver_asio.cpp
```diff
@@ -24,6 +24,40 @@
...
+    {
+      return true;
+    }
+
+    // Check if the address is in the private IP address ranges.
+    // For IPv4, the private address ranges are 10.0.0.0/8, 172.16.0.0/12, and 192.168.0.0/16.
+    if (address.is_v4())
+    {
+      uint32_t ip = address.to_v4().to_ulong();
+      return ((ip & 0xFF000000) == 0x0A000000) || // 10.0.0.0/8
```
FlorianReimold, line 42: I am puzzled why you would care about those. Why not just use a loopback address (127.0.0.0 - 127.255.255.255 https://de.wikipedia.org/wiki/Loopback)

ecal/core/src/io/udp_receiver_asio.cpp
rex-schilasky: Yes, I didn't name that unambiguously. What is meant is that the network packets come from your own machine.

ecal/core/src/io/udp_receiver_asio.cpp
rex-schilasky: That was my first approach. On my local system, however, I see packets that have my own IPv4 address as the sender 10.x.x.x instead of "127.x.x.x". I need to investigate this further - IT's A DRAFT :-)

rex-schilasky: Good news is that with this draft fix all gtest (including local UDP Pub/Sub) and all samples are running correctly with and without local mode setting and with and without activated npcap mode.

FlorianReimold: ok, but then you are probably using DNS to resolve your hostname. Just skip DNS and just send to 127.0.0.1 directly. Why use DNS, if you already know the IP.

Btw, there is no guarantee that DNS will resolve your hostname to a local IP. It could also be a public IP, if you have one. I guess you rather want to check whether the resolved IP address is a multicast or a unicast address. But still I think you don't need DNS at all to send to a loopback adapter.

ecal/core/src/io/udp_receiver_asio.cpp
FlorianReimold: But your `isLocalAddress` doesn't check if that IP belongs to your own machine. It checks whether the IP belongs to any machine in any local area network.

rex-schilasky: Now I got it .. yes that is true. So this approach will not work at all I think.

rex-schilasky:
This will fixed in an upcoming new udp layer implememtation.

PR #1242, opened 2023-11-06
Fix segfault in GetCpuCores

### LINKED ISSUE #1233: MMA crashes on arm64 platforms
### Problem Description

It seems to be the case that on some ARM64 Linux based platforms such as Buildroot Linux the MMA tool crashes with segmentation fault when it tries to query information about CPU utilization and other statistics.

### How to reproduce

Run ecal_mma on an ARM64-based Buildroot Linux operating system.

### How did you get eCAL?

### Environment

Version: eCAL 5.11.x and most likely all versions above.
OS: Buildroot Linux

### eCAL System Information

_No response_

jheeks:
### Description

The branch reading line 12 of /proc/cpuinfo is valid only for `__x86_64__`.
So, bugfix is to restrict this branch for `__x86_64__` only.
Or, simpler and more generic, use `get_ncores()`.

### Related issues

- Fixes #1233

### Cherry-pick to
- 5.11 (old stable)
- 5.12 (current stable)

### DIFF
--- a/contrib/mma/src/linux/mma_linux.cpp
+++ b/contrib/mma/src/linux/mma_linux.cpp
@@ -488,6 +488,7 @@ int MMALinux::GetCpuCores(void)
   std::list <std::string> cpuinfo_list = TokenizeIntoLines(FileToString("/proc/cpuinfo"));
   std::vector <std::string> cpucore_vect;
   auto idx = 0;
+#ifdef __x86_64__
   if(!is_vrm)
   {
     for (auto iter : cpuinfo_list)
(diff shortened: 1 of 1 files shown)


FlorianReimold:
What is with x86 (-> 32 bit)? And what will happen for non intel/amd platforms like arm? Wouldn't it be a better idea to use `nproc` instead of parsing some text that is meant to be interpretable by humans only?

jheeks:
I agree that `get_nprocs()` is much simpler and more generic. So, I used it.

FlorianReimold (approved):

PR #1252, opened 2023-11-14
Fixed wrong order of sender and receiver timestamp in make_frame() method

hannemn:
### Description
In some parts of the imeasurement class, timestamps of sender and receiver side have been accidentelly swapped. 

### Cherry-pick to
- 5.11 (old stable)
- 5.12 (current stable)


KerstinKeller (approved):
That's the danger with mixing up arguments.....

PR #1253, opened 2023-11-15
cleanup udp configuration and sender/receiver attributes

rex-schilasky:
### Description

In preparation for a new udp communication layer implementation there are some cleanups made in udp_confiuration and the sender and receiver udp attribute structures. Test under Windows and Ubuntu with local and network mode and for Windows with and without npcap.

### Related issues
Is not fixing any issue, should behave as before.

### Cherry-pick to
- _none_


KerstinKeller (approved):
To me it looks good, it's clear with the newly introduced functions.

PR #1259, opened 2023-11-16
Cleanup: Remove a ton of unnecessary ";" (that cause pedantic warnings).

KerstinKeller:


rex-schilasky (approved):
Some many semicolons had to die.

PR #1267, opened 2023-11-22
isolate eCAL cmake find logic from existing eCAL installation

rex-schilasky:
### Description
Fix for cmake eCAL package find logic (failed in case of an existing local eCAL installation)

### Cherry-pick to
-  5.12


KerstinKeller (approved):
Thank you. We need to cherry pick to 5.12, otherwise 5.12 has the same build issues.

PR #1269, opened 2023-11-23
cleanup core folder structure

rex-schilasky:
### Description
All eCAL core subfolder reorganized.

image

### Cherry-pick to
- _none_


ecal/core/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-# ========================= eCAL LICENSE =================================
+  # ========================= eCAL LICENSE =================================
```
KerstinKeller, line 1: Where does the space come from? :D

ecal/core/src/readwrite/shm/ecal_writer_shm.cpp
```diff
@@ -34,7 +34,7 @@
 #endif
 
 #include "ecal_def.h"
-#include "ecal_writer.h"
+#include "../ecal_writer.h"
```
KerstinKeller, line 37: Please no relative paths in our includes. rather "readwwrite/shm/ecal_writer.h" (so every include relative to the `src` (or `include` directory)

KerstinKeller (changes requested):
You can specify includes in two separate ways. With `""` which by convention means relative to the current file, or with `<>` which means relative to a specified include directory passed with `-I` or in cmake `target_include_directories`.

For portability, I'd prefer to use the `<>` option, and always specify relative to the include directories which we provide, which in CMake are the `include` directory, thus `#include<ecal/ecal.h>`, or relative to the source directory for private headers. `#include <config/ecal_config_reader.h>`.

I'd prefer not to have `#include "../header.h", as those can break more easily.

KerstinKeller (approved):

PR #1270, opened 2023-11-23
initialize random udp package id using nanosecond timestamp

rex-schilasky:
### Description
Random UDP package id now uses timestamp to start with a unique number for every process.

### Cherry-pick to
- 5.11 (old stable)
- 5.12 (current stable)

### DIFF
--- a/ecal/core/src/io/udp/snd_raw_buffer.cpp
+++ b/ecal/core/src/io/udp/snd_raw_buffer.cpp
@@ -21,6 +21,8 @@
  * @brief  raw message buffer handling
 **/
 
+#include <chrono>
+#include <mutex>
 #include <thread>
 
 #include "ecal_process.h"
@@ -31,8 +33,11 @@
 namespace
 {
   // random number generator
+  std::mutex xorshf96_mtx;
   unsigned long xorshf96(unsigned long& x, unsigned long& y, unsigned long& z)  // period 2^96-1
   {
+    const std::lock_guard<std::mutex> lock(xorshf96_mtx);
+
     unsigned long t;
     x ^= x << 16;
     x ^= x >> 5;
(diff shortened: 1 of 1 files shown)


ecal/core/src/io/udp/snd_raw_buffer.cpp
```diff
@@ -31,8 +33,11 @@
 namespace
 {
   // random number generator
+  std::mutex xorshf96_mtx;
   unsigned long xorshf96(unsigned long& x, unsigned long& y, unsigned long& z)  // period 2^96-1
   {
+    const std::lock_guard<std::mutex> lock(xorshf96_mtx);
+
```
KerstinKeller, line 40: Why do you have a mutex here? what do you want to protect? This function has no global variables which it might access, hence it is reentrant.

rex-schilasky: Yes that is right :)

KerstinKeller (approved):

ecal/core/src/io/udp/snd_raw_buffer.cpp
```diff
@@ -129,8 +131,18 @@ namespace eCAL
       msg_header.type = msg_type_header;
       {
         // create random number for message id
-        static unsigned long x = 123456789, y = 362436069, z = 521288629;
-        msg_header.id = xorshf96(x, y, z);
+        {
+          static std::mutex xorshf96_mtx;
+          const std::lock_guard<std::mutex> lock(xorshf96_mtx);
+
+          static unsigned long x = static_cast<unsigned long>(std::chrono::duration_cast<std::chrono::nanoseconds>(
```
KerstinKeller, line 138: @rex-schilasky I was wondering if std::random would be a bit more appropriate here, to indicate we need a random number to start with. If you started two processes at the exact same time, would they get the same numbers this way? Though still unlikely.

PR #1272, opened 2023-11-26
Python: Fix client remove response callback

DownerCase:
### Description
Fixed the python bindings for removing a response callback from a service client:

`ecal.core.service.py` :
- Client had `rem_response_callback` taking a method name parameter (which was unused) and calling non-existant `ecal.core.core.client_rem_response_callback`.

`ecal.core.core.py`:
- `client_rem_response_callback` doesn't exist because it was typoed in as `client_rem_method_callback`.

`ecal_wrap.cxx`:
- Implementation of `client_rem_response_callback` does exist
- Unfortunately, the `_ecal_methods` mapping binds the python API function `client_rem_response_callback` to the C++ function `client_add_response_callback`!

### Related issues

- None

### Cherry-pick to
- 5.11 (old stable)
- 5.12 (current stable)

### DIFF
--- a/lang/python/core/ecal/core/core.py
+++ b/lang/python/core/ecal/core/core.py
@@ -478,7 +478,7 @@ def client_add_response_callback(client_handle, callback):
   return _ecal.client_add_response_callback(client_handle, callback)
 
 
-def client_rem_method_callback(client_handle):
+def client_rem_response_callback(client_handle):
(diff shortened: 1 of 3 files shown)


KerstinKeller (commented):
Looks ok from my side.

FlorianReimold (approved):

PR #1284, opened 2023-12-18
Get mma linux more generic

jheeks:
### Description

* remove dependency to sysstat and iostat
* get most data from /proc
* consider all mounted disks
* results in better support of embedded devices

### Related issues

-


contrib/mma/src/linux/mma_linux.cpp
```diff
@@ -29,26 +29,36 @@
...
+#define MMA_NET_CMD "cat /proc/net/dev"
+#define MMA_DISK_CMD "cat /proc/diskstats"
+#define MMA_PS_CMD "cat /proc/[0-9]*/stat 2>/dev/null"
+
 MMALinux::MMALinux():
-  cpu_pipe_(std::make_unique<PipeRefresher>("sar -u 1 1",500)),
-  network_pipe_(std::make_unique<PipeRefresher>("ifstat 1 1",500)),
-  disk_pipe_(std::make_unique<PipeRefresher>("iostat -N 1 2",500)),
-  process_pipe_(std::make_unique<PipeRefresher>("ps -aux",500))
+  page_size(sysconf(_SC_PAGE_SIZE)),
```
hannemn, line 47: Please mark member variables with a '_' suffix like `page_size_ `.

contrib/mma/src/linux/mma_linux.cpp
```diff
@@ -29,26 +29,36 @@
...
+  ticks_per_second(sysconf(_SC_CLK_TCK)),
```
hannemn, line 48: Same here

hannemn (changes requested):
Thanks for reworking the MMA and for getting rid of sysstat and ifstat dependencies. I just did a quick test on my x86 machine and everything still works as expected. Once the coding style findings are reworked, we can bring your adaptions into the master.

contrib/mma/src/linux/mma_linux.cpp
```diff
@@ -257,19 +319,73 @@ ResourceLinux::ProcessStatsList MMALinux::GetProcesses()
...
+      if (delta_count > 0)
+        process_stats.cpu_load.cpu_load = (cur - prev) / (0.005 * delta_count);
+      item.utime = cur;
+      item.process_stats = process_stats;
+      proc_prev_map[process_stats.id] = item;
+    }
+    proc_prev_count_ = item.count;
+
+    // garbage collection
+    for (auto it = proc_prev_map.begin(), next = it; it != proc_prev_map.end(); it = next)
```
hannemn, line 378: Iterating on a map and at the same time removing certain elements can be dangerous. I am not 100% sure if your code can lead to unintended side-effects.

jheeks: The implementation follows a pattern given on Stackoverflow here: https://stackoverflow.com/a/39719450
The idea is to get the *next* iterator already before erasing the current entry.
The erased entries belong to old processes which are finished, means not listed in `/proc/[pid]` any more.

contrib/mma/src/linux/mma_linux.cpp
jheeks: This rule applies to the already available member `mutex` as well, correct?

hannemn: Yes, we were a bit careless with our naming conventions in the past. ;-)

contrib/mma/src/linux/mma_linux.cpp
hannemn: I see and acually thought there is a std::erase_if() functionality from the <algorithm> library available that can be used here. But it turned out this helper method comes along with C++20 which is too "new" for our code base.

hannemn (approved):

DownerCase:
> * consider all mounted disks

Thank you this specific change! I was recording to a "data drive" and was sad when mma wasn't reporting the usage of my recording drive..

PR #1285, opened 2023-12-20
fix wrapper function miss matching

GJYang:
### Description
Fix python wrapper function name
`client_set_host_name` to `client_set_hostname`

the method defined here:
https://github.com/eclipse-ecal/ecal/blob/28ef1a150bf7e27d0748c124d42a7381f72c4862/lang/python/core/src/ecal_wrap.cxx#L1633C22-L1633C22

### Related issues
-


hannemn (approved):
Thanks for contributing this fix @GJYang and you are totally right here. We made the underlying wrapper call accidentally with a small typo.

PR #1290, opened 2024-01-08
ensure correct udp socket concurrent handling by mutex

rex-schilasky:
### Description
Sporadic crashing of UDP layer shutdown (triggered by eCAL::Finalize ) by resetting the socket while still handling receives. Socket needs to be protected by mutex.


rex-schilasky:
Windows tests not passed anymore. Will not be applied to 5.11.x

PR #1291, opened 2024-01-08
1st draft for Python Measurement Writer.

KerstinKeller:
### Description
Add high level API for writing measurements in Python

This was a first draft used in some Python tooling, and I thought it would be useful to bring it to the repo.
However, not yet 100% happy with the API, wouldn't want to make it stable. However the functionality may be valuable.

### DIFF
--- a/lang/python/ecalhdf5/ecal/measurement/writer.py
+++ b/lang/python/ecalhdf5/ecal/measurement/writer.py
@@ -1,0 +1,105 @@
+# ========================= eCAL LICENSE =================================
+#
+# Copyright (C) 2016 - 2019 Continental Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+# 
+#      http://www.apache.org/licenses/LICENSE-2.0
+# 
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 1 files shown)


lang/python/ecalhdf5/ecal/measurement/writer.py
```diff
@@ -0,0 +1,105 @@
...
+
+def combine_encoding_and_type(encoding, type):
+  return encoding + ":" + type
+
+class ProtobufChannelWriter:
+    def __init__(self, channel_name: str, protobuf_type, meas: ecalhdf5.Meas):
+        self._protobuf_type = protobuf_type
+        self._binary_writer = BinaryChannelWriter(channel_name, meas=meas, encoding = self.encoding, type = self.type, descriptor = self.descriptor)
+
+    def write_entry(self, data, rcv_timestamp: float, snd_timestamp: float):
```
KerstinKeller, line 33:
```suggestion
    def write_entry(self, data, rcv_timestamp: int, snd_timestamp: int):
```

lang/python/ecalhdf5/ecal/measurement/writer.py
```diff
@@ -0,0 +1,105 @@
...
+        self._meas = meas
+        self._channel_name = channel_name
+        self._encoding = encoding
+        self._type = type
+        self._descriptor = descriptor
+
+        self._meas.set_channel_description(channel_name, descriptor)
+        self._meas.set_channel_type(channel_name, combine_encoding_and_type(encoding, type))
+
+    def write_entry(self, data, rcv_timestamp: float, snd_timestamp: float):
```
KerstinKeller, line 63:
```suggestion
    def write_entry(self, data, rcv_timestamp: int, snd_timestamp: int):
```

PR #1313, opened 2024-01-18
[Core] Remove unnecessary `new` operators ... 

KerstinKeller:
and replace with `make_shared` or `make_unique`.

### Description
There is really need for explicit use of `new` so replace with creation of shared/unique pointers.

### Cherry-pick to:
- None (changes public API slightly and will therefore be ABI compatible)

### DIFF
--- a/ecal/core/include/ecal/ecal_timer.h
+++ b/ecal/core/include/ecal/ecal_timer.h
@@ -37,13 +37,13 @@ namespace eCAL
    *
    * The CTimer class is used to realize simple time triggered callbacks.
   **/
-  class ECAL_API CTimer 
+  class CTimer 
   {
   public:
     /**
      * @brief Constructor. 
     **/
-    CTimer();
+    ECAL_API CTimer();
 
     /**
      * @brief Constructor. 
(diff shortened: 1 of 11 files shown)


KerstinKeller:
We should rather delete the move operator than to just default them, it might lead to problems.

FlorianReimold (approved):

KerstinKeller:
Ubuntu build fails with:
```
5: Test command: /home/runner/work/ecal/_build/bin/ecal_test_core-5.13.0
5: Working Directory: /home/runner/work/ecal/_build/bin
5: Test timeout computed to be: 10000000
5: Running main() from /home/runner/work/ecal/ecal/thirdparty/gtest/googletest/googletest/src/gtest_main.cc
5: [==========] Running 3 tests from 1 test suite.
5: [----------] Global test environment set-up.
5: [----------] 3 tests from Core
5: [ RUN      ] Core.MultipleInitializeFinalize
5: [       OK ] Core.MultipleInitializeFinalize (7006 ms)
5: [ RUN      ] Core.LeakedPubSub
5: [       OK ] Core.LeakedPubSub (5004 ms)
5: [ RUN      ] Core.CallbackDestruction
5: double free or corruption (out)
 5/14 Test  #5: test_core ........................Subprocess aborted***Exception:  16.16 sec
 ```
 
 Need to check if sporadic failure (even though - why?) or reproducible from the changes (though I wouldn't know / understand why...).

PR #1315, opened 2024-01-19
Feature/qt6 compatibility

### LINKED ISSUE #565: Qt6 compatiblity
The goal is to have all eCAL Applications building with Qt6. But as we will have to deal with Linux distributions using Qt 5 for a long time, the Qt 5 Compatibility must be maintained.
Qt 6.2 is the first LTS release and the first feature complete Version. Therefore, 6.0 / 6.1 are not important to support.

My expectation from this is future proofeness as well as better looks on Windows 11.

The Applications are:
- Mon
- Rec
- Play
- Sys
- Launcher

Qt Versions to support are:
- Qt 6.2
- Qt 5 down to the oldest still supported Ubuntu Version (18.04 with Qt 5.9 at the time of writing)

The way of tackeling this probably is:
1. Make sure that building against Qt 5.15.2 is warning-free. Qt 5.15 basically deprecated everything that will be removed in Qt 6, so a warning free build will greatly help with upgrading.
2. Modify the CMakeLists so they don't hard-coded look for Qt5

FlorianReimold:
### Description
Made eCAL Compatible with Qt6

- Compiles with both Qt5 and Qt6
- CMake option is now `HAS_QT`, but the legacy `HAS_QT5` option is still supported
- Min Supported Qt Version: 5.9 (that's the version Ubuntu 18.04 ships with)
- Max Supported Qt Version: 6.6.1 (probably higher, but that's the latest I tested with)
- Official Windows installer now ships with Qt6
- Qt6 dropped the QWinExtras that enabled the Thumbnail Taskbar buttons of eCAL Play and eCAL Rec on Windows. Therefore, those buttons had to be removed.
- Fixed the mon include regexp from the `ecal.ini`, so it doesn't need to be "fixed" in C++ code anymore (Added `^` and `$` to match entire string)
- Solved Qt Related warnigns
- Updated QWT submodule to the latest development version. 

### Related issues
- Fixes #565

### Cherry-pick to
- _none_


ecal/core/cfg/ecal.ini
```diff
@@ -158,15 +158,15 @@ protocol_v1                        = 1
...
+; filter_excl                      = ^__.*$                        Topics blacklist as regular expression (will not be monitored)
 ; filter_incl                      =                               Topics whitelist as regular expression (will be monitored only)
 ; filter_log_con                   = info, warning, error, fatal   Log messages logged to console (all, info, warning, error, fatal, debug1, debug2, debug3, debug4)
 ; filter_log_file                  =                               Log messages to logged into file system
 ; filter_log_udp                   = info, warning, error, fatal   Log messages logged via udp network
 ; --------------------------------------------------
 [monitoring]
 timeout                            = 5000
-filter_excl                        = __.*
+filter_excl                        = ^__.*$
```
FlorianReimold, line 169: @rex-schilasky Please leave a comment whether you think that this change can have bad side-effects

app/mon/mon_gui/CMakeLists.txt
```diff
@@ -217,9 +229,64 @@ target_include_directories(${PROJECT_NAME} PRIVATE src)
 target_include_directories(${PROJECT_NAME} PRIVATE $<TARGET_PROPERTY:eCAL::core,INCLUDE_DIRECTORIES>)
 
 
-if(WIN32)
+if ((WIN32 OR APPLE) AND (${QT_VERSION_MAJOR} GREATER_EQUAL 6))
```
KerstinKeller, line 232: We have this (or similar) code in many places. Could you please make cmake macros / functions for it?

FlorianReimold: These are just 3 commands (`qt_generate_deploy_app_script`, `add_custom_command`, `install`). The rest are comments to explain what is going on. Do you think that it is worth to make a script for that, discover the script file, include the script and then call it?
I feel like that would cause more code and be more difficult to understand, because you would have to go to that script to see what is going on. And one would have to implement some kind of argument parsing, if you intend to keep it flexible. It would also cause additional coupling of all eCAL Qt Application to some custom script that is stored somewhere. I was excited to remove custom deploy scripts for Qt6, so I am not sure if encapsulating 3 commands justify creating one, again.

KerstinKeller: But you have the same code in at least 4 places (mon_gui, play_gui, rec_gui, sys_gui, ...)
It's like arguing that you're not creating another header file and not using functions, because it's too complicated.

It's 60 lines of code, being (as far as I can see), exactly identical in at least mon_gui, play_gui, rec_gui.
If you want to keep it with the apps, put it in the app subdirectory, or define the macro in app/CMakeLists.txt.
But duplicated code is duplicated code and should be put in a function (or cmake macro, because that is in the end text replacement, and does not introduce new scopes which you may not want).

KerstinKeller: You could also make it part of `ecal_add_app_qt`, if it's something to be done for all ecal qt apps.

FlorianReimold: Hm ok, so your intention is to put the legacy Qt5 `install(CODE ...` in that script as well? I agree that that script could go in its own file, but the scope of this PR was to just add Qt6. Also the Qt5 script wasn't written by me and I honestly was afraid to touch it, so I just added the new commands for Qt6.

> It's like arguing that you're not creating another header file and not using functions, because it's too complicated.

I didn't write it would be too complicated and (for Qt6) the CMakeLists.txt does use functions. But yea, it kind of is similar to copying code from one to the next header file versus creating a separate project for it that you then have to reference to. It's always a tradeoff between coupling and code duplication.
I felt like Qt6 already encapsulated everything very well and those official qt functions are probably better understandable than a custom one. Thus I didn't want to hide those 3 calls with the downside of more coupling.

>You could also make it part of ecal_add_app_qt, if it's something to be done for all ecal qt apps.

Sure, could be a solution for the apps, as those are coupled to those functions anyways. It would get rid of the flexbility, but maybe we are ok with a preconfigured variant of the scripts.
We would also maybe need a preconfigured variant for samples, monitor plugins and libraries.

KerstinKeller (approved):
From my point we can merge it, but we need to schedule a general CMake cleanup issue at some point in the future.

PR #1316, opened 2024-01-19
[Core] Default move constructor/assignment for CPublisher/CSubscriber

KerstinKeller:
as the previous manual implementation was just plainly wrong.

Cherry-pick to:
- none

### DIFF
--- a/ecal/core/src/pubsub/ecal_publisher.cpp
+++ b/ecal/core/src/pubsub/ecal_publisher.cpp
@@ -71,12 +71,38 @@ namespace eCAL
   /**
    * @brief CPublisher are move-enabled
   **/
-  CPublisher::CPublisher(CPublisher&& rhs) noexcept = default;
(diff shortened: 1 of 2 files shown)


KerstinKeller:
@rex-schilasky @hannemn @FlorianReimold 
Let's discuss about this. it's not so easy and our move tests crash with the default generated move operators.

E.g. in the move assignement case, I'd argue that we should first call `Destroy()`, because the original Subscriber might have been actively communicating.

I am not quite sure if we should just delete them altogether, though that would be an API change (not semver compatible).
So I am not sure how they are implemented correctly.

FlorianReimold:
What's the reason for calling Destroy and not just moving all the variables over? I guess the communication implementation is probably in the `m_datareader`, right? Does the `m_datareader` have a pointer or reference to the original object? Because in that case I think moving should be entirely disabled (or the implementation will become extremely difficult). But otherwise it should be fine...

FlorianReimold (approved):

PR #1319, opened 2024-01-19
Fixed deprecation warning in eCAL Play

FlorianReimold:
### Description
Fixed deprecation warning in eCAL Play

### Cherry-pick to
- _none_


FlorianReimold:
Will be fixed with a new API function in HDF5 Meas

PR #1323, opened 2024-01-22
fix: yaml-cpp compatability with CMake < 3.18

### LINKED ISSUE #1322: yaml-cpp error with CMake < 3.18
### Problem Description

#1250 Introduced an alias to yaml-cpp if not already present. This works great on CMake versions 3.18 or greater.
With older versions CMake fails with the message:
```text
CMake Error at app/meas_cutter/CMakeLists.txt:27 (add_library):
  add_library cannot create ALIAS target "yaml-cpp::yaml-cpp" because target
  "yaml-cpp" is imported but not globally visible.
```

The ability to create a scoped ALIAS for non-global imported libraries was introduced in version 3.18.
[add_library(ALIAS) Documentation](https://cmake.org/cmake/help/git-master/command/add_library.html#alias-libraries)

This affects Ubuntu 20 and any other environment running CMake less than 3.18

This issue isn't caught by CI because [it uses the latest CMake version](https://github.com/actions/runner-images/blob/ubuntu20/20240116.2/images/ubuntu/Ubuntu2004-Readme.md#tools). 

To resolve this I could find two options:

1. Increase the minimum required version to at least 3.18 
  - Requires Ubuntu 20 to acquire new CMake versions, not that its difficult and the Ubuntu 18 instructions already have that procedure.
2. Take inspiration from [this GitLab issue](https://gitlab.kitware.com/cmake/cmake/-/issues/20641#note_797165) and do something like:
```cmake
if (NOT TARGET yaml-cpp::yaml-cpp AND TARGET yaml-cpp)
    add_library(yaml-cpp::yaml-cpp INTERFACE IMPORTED)
    target_link_libraries(yaml-cpp::yaml-cpp INTERFACE yaml-cpp)
endif()
```

I was able to test option 2; it built and `ldd` showed it linking to `libyaml-cpp.so` correctly. 😄 

I can make a PR with your preferred choice if you'd like.

### How to reproduce

1. Try to configure master using CMake <3.18 with meas_cutter included in the build (BUILD_APPS and HAS_HDF5)
2. Observe the failure.

### How did you get eCAL?

### Environment

- eCAL master (5.13)
  - **This will also affect 5.11 and 5.12** support branches if the current state of #1320 and #1321 gets merged!
- Primarily Ubuntu 20; but could affect any.
- **CMake < 3.18**

### eCAL System Information

_No response_

DownerCase:
### Description
- Fixes yaml-cpp compatability with CMake < 3.18
  - See #1322 for details
- Corrects the minimum required version to 3.15
  - Required for `list(APPEND)`

### Related issues

closes #1322

### Cherry-pick to
- none

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,7 @@
 #
 # ========================= eCAL LICENSE =================================
 
-cmake_minimum_required(VERSION 3.13)
+cmake_minimum_required(VERSION 3.15)
 
 include(CMakeDependentOption)
 
--- a/app/meas_cutter/CMakeLists.txt
+++ b/app/meas_cutter/CMakeLists.txt
@@ -24,7 +24,9 @@ find_package(yaml-cpp REQUIRED)
 
 #compatibility with yaml-cpp < 0.8.0
 if (NOT TARGET yaml-cpp::yaml-cpp AND TARGET yaml-cpp)
-  add_library(yaml-cpp::yaml-cpp ALIAS yaml-cpp)
+  # ALIASing a imported non-global library requires CMake 3.18 so we do this
+  add_library(yaml-cpp::yaml-cpp INTERFACE IMPORTED)
+  target_link_libraries(yaml-cpp::yaml-cpp INTERFACE yaml-cpp)
 endif()
 
 


FlorianReimold:
@KerstinKeller I would like to get a final approval from you 😀

KerstinKeller (approved):

Issue #1318, opened 2024-01-19
Sporadic Test Failure: Core.CallbackDestruction


KerstinKeller:
### Problem Description

Ubuntu build fails with:
```
5: Test command: /home/runner/work/ecal/_build/bin/ecal_test_core-5.13.0
5: Working Directory: /home/runner/work/ecal/_build/bin
5: Test timeout computed to be: 10000000
5: Running main() from /home/runner/work/ecal/ecal/thirdparty/gtest/googletest/googletest/src/gtest_main.cc
5: [==========] Running 3 tests from 1 test suite.
5: [----------] Global test environment set-up.
5: [----------] 3 tests from Core
5: [ RUN      ] Core.MultipleInitializeFinalize
5: [       OK ] Core.MultipleInitializeFinalize (7006 ms)
5: [ RUN      ] Core.LeakedPubSub
5: [       OK ] Core.LeakedPubSub (5004 ms)
5: [ RUN      ] Core.CallbackDestruction
5: double free or corruption (out)
 5/14 Test  #5: test_core ........................Subprocess aborted***Exception:  16.16 sec
```

### How to reproduce

Only happens sporadically.

### How did you get eCAL?

Custom Build / Built from source

### Environment

Ubuntu 20

PR #1324, opened 2024-01-22
[Core] Fix Race Conditions in CMsgSubscriber

KerstinKeller:
- Properly implement destructor (in inherited classes)! -> Destroy may not be called in base CMsgSubscriber class as this can lead to stack unwinding (most likely causes an exception in destructor, or leads to already deleted functions of derived class being called when they are already deleted)
- Correct order of RemReceiveCallback and resetting internal pointer
- protect function pointer with mutex.
- Make testcase "harder" to more easily trigger race conditions.

### Cherry-pick to
- 5.11 (old stable)
- 5.12 (current stable)

### DIFF
--- a/ecal/core/include/ecal/msg/capnproto/subscriber.h
+++ b/ecal/core/include/ecal/msg/capnproto/subscriber.h
@@ -73,7 +73,7 @@ namespace eCAL
       **/
       ~CBuilderSubscriber() override
       {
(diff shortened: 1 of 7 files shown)


KerstinKeller:
Fixes #1318

FlorianReimold (approved):

PR #1334, opened 2024-01-24
[Core] Fix: proper destructors / rule of 5 for everything involving std::thread

### LINKED ISSUE #1329: ecal_thread.h misses Destructor
https://github.com/eclipse-ecal/ecal/blob/master/ecal/core/src/util/ecal_thread.h

The class has an std::thread member variable that is not joined on destruction. Thus, the class needs a destructor calling join

KerstinKeller:
### Related issues
Fixes #1329  

### Cherry-pick to
- 5.11 (old stable)
- 5.12 (current stable)
(might not be easily cherry-pickable)

### DIFF
--- a/ecal/core/include/ecal/ecal_timed_cb.h
+++ b/ecal/core/include/ecal/ecal_timed_cb.h
@@ -68,6 +68,11 @@ namespace eCAL
     **/
     virtual ~CTimedCB() { Stop(); }
 
+    CTimedCB(const CTimedCB&) = delete;
+    CTimedCB& operator=(const CTimedCB&) = delete;
+    CTimedCB(CTimedCB&& rhs) = delete;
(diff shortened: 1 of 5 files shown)


ecal/core/include/ecal/ecal_timed_cb.h
```diff
@@ -68,6 +68,11 @@ namespace eCAL
     **/
     virtual ~CTimedCB() { Stop(); }
 
+    CTimedCB(const CTimedCB&) = delete;
+    CTimedCB& operator=(const CTimedCB&) = delete;
+    CTimedCB(CTimedCB&& rhs) = delete;
+    CTimedCB& operator=(CTimedCB&& rhs) = delete;
```
FlorianReimold, line 74: Does this cause an ABI break? 🤔

FlorianReimold (approved):

PR #1358, opened 2024-02-09
[new-core-alignment] service includes cleaned up

rex-schilasky:
### Description
Service header clean up, see  see https://github.com/eclipse-ecal/ecal-core/pull/15.

Based on changes by @FlorianReimold):

_I tried to apply the IWYU pragma. I mainly focused on std headers, as those don't come with umbrella headers._

### Cherry-pick to
- None

### DIFF
--- a/ecal/service/ecal_service/include/ecal/service/client_manager.h
+++ b/ecal/service/ecal_service/include/ecal/service/client_manager.h
@@ -19,10 +19,14 @@
 
 #pragma once
 
-#include <memory>
+#include <cstddef>
+#include <cstdint>
 #include <map>
+#include <memory>
 
 #include <ecal/service/client_session.h>
+#include <mutex>
+#include <string>
 
 namespace eCAL
 {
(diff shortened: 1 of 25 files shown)


KerstinKeller (commented):
👍 
I think if we do cleanup like this, we should try to come up with a rule, e.g. first include your own header (e.g. `client_session.cpp` first include is `client_session.h`, then *all* standard headers, then all other thirdparty headers, and then all needed ecal headers? 

Not 100% sure, but I would mix it like in `server_manager.h` 😄

FlorianReimold:
I usually include the class declaration `.h` (in cases of `.cpp`) first, then stl headers, then custom headers. But in this case it took a considerable amount of time to just go through the clang-tidy output and decide for each warning if I wanted to apply the fix or not (there were a couple thousand warnings).
And as the order of includes are not supposed to matter anyways, I just clicked the fix-it checkboxes. I think the manual labor of re-sorting this is not worth the time, I would rather merge this and then further improve the includes. I am certain that I missed a lot of includes that should be there and I didn't fix any unnecessary includes, yet (i.e. remove them). A further improvement would also be to mark umbrella headers accordingly with an IWYU export comment.

KerstinKeller (approved):

PR #1359, opened 2024-02-11
CMake: automatic venv creation for docs

DownerCase:
### Description
- Teaches CMake to automatically create a Python virtual environment when building the docs is enabled.
  - This only happens if it is missing.
  - If `venv` fails, CMake is halted and the user is informed they may manually create a virtual environment at the expected location. (Eg: If their version of Python does not include a fully functioning `venv` module.)
  - The virtual environment is the only place searched for a compatible (3.7+) Python version.
- Dependencies are installed at build time and will be reinstalled whenever the `requirements.txt` is changed.
  - Use of `--require-virtualenv` is a failsafe to ensure we are, in fact, in a virtual environment. The flag was introduced in Python 3.6, and I've tested with Python 3.7.0 and pip 10.0.1, so no concerns there.
- Removes `FindSphinx.cmake` in favor of directly invoking sphinx with `python -m sphinx` (ensuring we use the version we expect)
- Refactors IDE folder handling by using [CMAKE_FOLDER](https://cmake.org/cmake/help/latest/variable/CMAKE_FOLDER.html#cmake-folder) to set the default `FOLDER` property to targets.
- Updates CI to use this new mechanism (remove manual venv creation/activation)
  - This is safe as you proceed to (re)-build the wheels with Python versions sourced from their own venvs
- Swaps the existing `find_package` calls for Python to explicitly use Python 3.

#### Motivation

I went to build the documentation by setting `BUILD_DOCS=ON`; CMake complained so I had a look at what it was trying to do (not knowing about the Python requirements); saw the original abandoned attempt at implementing this; decided I wanted that, and made it so.

### Related issues

- None

### Cherry-pick to
- _none_


KerstinKeller:
Thanks for this PR! It looks like you put a lot of work into it. :+1:
 
In general, I am not so sure how much I like the idea that CMake is responsible for *everything*, including venv Generation, but in the end, it's convenient to be able to run it reliably, all from one place...

FlorianReimold:
Hey @DownerCase, thanks for the PR. However, we see multiple problems here:

- CMake should not be responsible for fetching dependencies. There must always be a way to provide dependencies from the outside.
- The PR disables building without any venv. There are valid reasons to just build with the default system-wide python environment.
- It also disables building without having pip. On Ubuntu for instance you often want to apt-get python dependencies.
- This forces the build machine to have access to the internet.
- I have seen a hardcoded Python3.7 version somewhere, which is obviously bad if you want to build for other python environments

DownerCase:
Thanks Florian, I respect your decision but would like to respond to you points. 

Remember this is specifically scoped to *only the documentation building*, not general wheels. Nothing has changed if ‘BUILD_DOCS’ is ‘OFF’ (except changing to explicitly look only for Python3)

1. CMake will only create the venv if it does not already exist where it expects. You can create it in the expected location and the CMake will just use it. I could modify the script to make the external location an explicit user-facing option.

2. I haven’t checked but are all your version pinned dependencies available in all supported versions of APT? There are very good reasons to _not_ “pip install” on the system distribution. Such as it not being possible in the upcoming Ubuntu release and other modern distros due to [PEP 668](https://peps.python.org/pep-0668/)

3. As mentioned, I haven’t check but don’t think APT has your specific version of Sphinx-book-theme, for example. You certainly don’t take this approach on GitHub actions at least.


4. Internet is only required to install the dependencies if the venv does not already exist. Otherwise it will blindly try and use whatever python environment is at the location. Again, can make the venv path a user-facing option (or to use any active venv).

5. You saw the 3.7 in a ‘find_package’ call. It follows standard CMake behaviour and is a **minimum**. Your documentation requirements need at least Python 3.7.

PR #1370, opened 2024-02-14
[new core alignment] integrate new core

rex-schilasky:
### Description
This is the first PR to redesign the ecal core. You can find the motivation behind it and the current status here

https://github.com/eclipse-ecal/ecal-core


CMakeLists.txt
```diff
@@ -91,6 +91,32 @@ option(BUILD_STANDALONE_PY_WHEEL               "Build eCAL python binding as sta
...
+# ------------------------------------------------------------------------------------------------------------------------------------------------------------------
+# core transport layer options (change with care)
+# ------------------------------------------------------------------------------------------------------------------------------------------------------------------
+option(ECAL_CORE_TRANSPORT_UDP                           "Enables the eCAL to transport payload via UDP multicast"                                               ON)
+option(ECAL_CORE_TRANSPORT_TCP                           "Enables the eCAL to transport payload via TCP"                                                         ON)
+option(ECAL_CORE_TRANSPORT_SHM                           "Enables the eCAL to transport payload via local shared memory"                                         ON)
+# ------------------------------------------------------------------------------------------------------------------------------------------------------------------
```
KerstinKeller, line 116: @rex-schilasky Can you please create these options in CMakeLists.txt in the `ecal` subfolder (or whatever is the main CMakeLists.txt for the core component).
This will allow to have options which are relevant for eCAL core, but they are not necessary set from the top level project, aka the whole CMake solution.

rex-schilasky: Done

ecal/core/include/ecal/msg/subscriber.h
```diff
@@ -231,27 +204,14 @@ namespace eCAL
...
-
-    ECAL_DEPRECATE_SINCE_5_13("Please use SDataTypeInformation GetDataTypeInformation() instead. This function will be removed in eCAL6.")
-    virtual std::string GetDescription() const
-    {
-      return GetDataTypeInformation().descriptor;
-    };
-
     // We cannot make it pure virtual, as it would break a bunch of implementations, who are not (yet) implementing this function
-    virtual SDataTypeInformation GetDataTypeInformation() const { return SDataTypeInformation{}; }
+    virtual struct SDataTypeInformation GetDataTypeInformation() const { return SDataTypeInformation{}; }
```
KerstinKeller, line 208: Why do you need the sttruct keyword here? This should not be necessary? Is it not known or are we maybe missing an include?

rex-schilasky: I rechecked it and it's not needed for sure. Not sure how this change happened.

ecal/core/src/io/mtx/ecal_named_mutex_base.h
```diff
@@ -47,9 +47,7 @@ namespace eCAL
   class CNamedMutexStubImpl : public CNamedMutexImplBase
   {
   public:
-    ~CNamedMutexStubImpl()
-    {
-    }
+    ~CNamedMutexStubImpl() override = default;
```
KerstinKeller, line 50: I think the constructor can be removed altogether - otherwise rule of 5.

KerstinKeller (approved):
I wI will approve it even though i have not been able to review all files. we need to make sure to test things appropriately before an eCAL6 release.

Issue #1292, opened 2024-01-09
Repurpose ID as unique publisher ID


KerstinKeller:
### Context

The publisher has a function to set an ID https://github.com/eclipse-ecal/ecal/blob/0f16b39e964ebe5e2c64fbb2a8ddeea519f19f6b/ecal/core/include/ecal/ecal_publisher.h#L365, in order to distinguish topics that are send on the same topic name, but coming from different publishers.
By default this topic id is 0.
By giving a unique default ID to the publisher, this ID can later be used e.g. to store / connect descriptor information with a given publisher, also in measurements.

### Tasks and updates

Discussion points:
- Do  we want to deprecate the SetId function?

cmallardm:
Awesome. Initially, I resolved this by working with a previous commit of the master branch that a teacher had tested weeks ago and suggested I work with it. As expected, it worked.

And with this new information, ill be working with the mentined branches. Thanks for getting in touch

Issue #1076, opened 2023-04-19
Support MCAP as Measurement File format


KerstinKeller:
### Context

Our current hdf5 measurement format has a few drawbacks:
- We don't support compression :(
- HDF5 can be hard to crosscompile
- New bug limit of message descriptor size (#1062)

The [MCAP format](https://foxglove.dev/mcap) is an alternative recording format, designed explicitly for the data recording usecase.

### Proposal

Let's support MCAP as a measurement output!

It can be configurable, which file format to write, with the default being HDF5. In the long run, this can be switched to MCAP and at some point (if all goes well) we might deprecate HDF5.

All reading applications (player / meas_cutter / C++ High-Level-API / Python API) should be able to read both formats.

### Tasks and updates

- [ ] Introduce a common base class with the old `eCALMeas` interface, so that the alternative implementation can serve as a drop-in
- [ ] create configuration item in eCAL.ini for eCAL Player / other apps (to decide which format to write)
- [ ] make an implementation for the base class based on mcap (what about options, etc (-> config items?) and itegrate dependencies via submodules
- [ ]  make a Reader factory function to create an appropriate Reader based on the actual measurement type.

git-pi-e:
Will this be part of the 5.14 release?

AntoineBassoul:
Any updates ? ROS still a gigantic over opinionated pain, but we need native recording visualization with foxglove xD

PR #1375, opened 2024-02-16
[measurement] HDF5 v6 - Store channel ID in measurement

KerstinKeller:
### Description
This PR introduces a modified HDF5Meas API, based on Channel objects (channel + id) instead of channel names only.
The old API is still available as `eCAL::eh5::v2::HDF5Meas`, whereas the new API is available as `eCAL::eh5::HDF5Meas` respective `eCAL::eh5::v3HDF5Meas`, using an `v3` inline namespace.

This does *not* change the recorder or any app to record the new format. Replay is possible, but with the limitation of not distinguishing channels with the same name.

Open for discussion:
Namespaces: Do the namespaces make sense? Should we adapt them to the includes (e.g. `ecalhdf5` namespace?

### Related issues
This is loosely related to #1076, and also to #1292 

### Cherry-pick to
No cherry picking. However, we can manually sync the contrib/ecalhdf5 folder with the v5.x support branch and use the new api with the `v2::` namespace.
This allows to read v6 measurements in eCAL 5.x.

### Docs???

### DIFF
--- a/contrib/ecalhdf5/CMakeLists.txt
+++ b/contrib/ecalhdf5/CMakeLists.txt
@@ -42,9 +42,15 @@ set(ecalhdf5_src
     src/eh5_meas_file_v4.h
     src/eh5_meas_file_v5.cpp
     src/eh5_meas_file_v5.h
+    src/eh5_meas_file_v6.cpp
+    src/eh5_meas_file_v6.h
     src/eh5_meas_file_writer_v5.cpp
     src/eh5_meas_file_writer_v5.h
+    src/eh5_meas_file_writer_v6.cpp
+    src/eh5_meas_file_writer_v6.h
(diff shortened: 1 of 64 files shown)


KerstinKeller:
Will be reimplemented.

contrib/ecalhdf5/CMakeLists.txt
```diff
@@ -67,7 +78,10 @@ target_include_directories(${PROJECT_NAME}
 
 target_compile_definitions(${PROJECT_NAME}
   PRIVATE
-    $<$<BOOL:${MSVC}>:_UNICODE>)
+    $<$<BOOL:${MSVC}>:_UNICODE>
```
Peguen, line 81: Remove

PR #1380, opened 2024-02-17
Add CMakePresets.json

### LINKED ISSUE #528: CMake Presets
Currently we have in-CmakeLists.txt defined configurations to distinguish between Windows and Non-Windows builds. CMake does provide the presets-technique that one can use to hierarchically describe default configurations for different scenarios:

- Windows
- Linux
- Core only
- CLI apps only
- Complete with GUI
- etc.

Having Cmake presets

**Describe alternatives you've considered**
Not having them

DownerCase:
### Description

Adds CMakePresets.json with the most common user-facing build profiles:
- `core`: Minimal build. Focuses on building just the core libraries and nothing else.
- `wheel`: Builds the Python wheel.
  - Simplifies building the wheel to: `cmake --preset wheel && cmake --build --preset wheel`
- `cli`: Extends `core` to include the CLI/TUI applications
- `gui`: Extends `cli` to include the QT GUI applications too
- `docs`: Extends `wheel` to also build the Sphinx documentation
  - Made much more ergonomic to use via #1359 so you don't need to set up the documentation virtual environment

I did not include a "full" build, testing, or C# presets as I was focusing on the most common feature levels users (from my perspective) would want. These presets should evolve over time to meet end-user, contributor, and developer needs.

With this PR it is much easier to build the different feature levels; especially for users that may not want a complete build.

Additionally, users may write their own "CMakeUserPresets.json" to extend the set of presets with something tailored to their use and preference. ( I, personally, will be using this feature)

Lastly, IDEs, such as Visual Studio, understand CMakePresets and allow for quickly switching between the different presets, aiding development and checking the different feature levels.

#### Technical notes

The builds default the `CMAKE_BUILD_TYPE` to Release to prevent the combinatorial explosion of having explicit debug/release(/relwithdebinfo/minsizerel) presets.
This is only relevant for single-config generators (like "Unix Makefiles" or regular "Ninja") and does not affect "Visual Studio *" or "Ninja Multi-Config" generators.

This PR includes a CMake option to default `BUILD_SHARED_LIBS` to `OFF` **only** on "win32" builds. This setting is always required when building with MSVC due to the current DLL situation.
I do not define `BUILD_SHARED_LIBS` on non-win32 builds to match existing behaviour.

I bumped the minimum CMake version of CMakeFunctions solely because it was giving a [deprecation message about support for CMake < 3.5.](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#policy-settings)

[CMake Preset documentation](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html)

### Related issues

- closes #528 

### Cherry-pick to
- Whatever the team is happy for, it would be nice to have for 5.13 but you will have to change `HAS_QT` back to `HAS_QT5`. I don't think any other changes would be needed.


CMakeLists.txt
```diff
@@ -121,12 +121,14 @@ if(WIN32)
   option(ECAL_THIRDPARTY_BUILD_CURL            "Build CURL with eCAL"                                              ON)
   option(ECAL_THIRDPARTY_BUILD_HDF5            "Build HDF5 with eCAL"                                              ON)
   cmake_dependent_option(ECAL_THIRDPARTY_BUILD_QWT "Build qwt::qwt with eCAL"                                      ON  "HAS_QT" OFF)
+  option(BUILD_SHARED_LIBS                     "Build shared libraries; required off on MSVC"                     OFF)
```
KerstinKeller, line 124: Why does it have to be set explicitly? `BUILD_SHARD_LIBS` is already a CMake option, defaulting to `OFF` on Windows platforms?

DownerCase: I believe the cURL submodule was turning it on if it wasn't set causing problems for eCAL as curl goes first. This also matches what is done elsewhere in the project ([\[1\]](https://eclipse-ecal.github.io/ecal/development/building_ecal_from_source.html#fa-windows-windows-build), [\[2\]](https://github.com/eclipse-ecal/ecal/blob/master/.github/workflows/build-windows.yml#L116)) where BUILD_SHARED_LIBS is explicitly set `OFF` for Windows.

DownerCase: And this isn't in the CMakePresets because then you'd need separate Windows and non-Windows presets to handle the mandatory platform-dependent setting. The presets themselves don't handle platform dependent things very well. Its best at setting high-level options which the build system understands and can do what you ask (like handling whether BUILD_SHARED_LIBS needs to be set on your platform or which thirdparty submodules to build from source)

FlorianReimold (approved):

FlorianReimold:
Thanks for this PR! 👍

PR #1384, opened 2024-02-19
GH Actions Windows: Removed downgrade nuget

FlorianReimold:
### Description
Removed the downgrade-nuget step that was introduced in 2021

### Cherry-pick to
_Let's see if it works, first_


KerstinKeller (approved):

KerstinKeller:
Looks like it works 😄

FlorianReimold:
Hm, then cherry pick it to all supported branches, too, i guess?

Issue #1161, opened 2023-07-25
eCAL Packages on PyPi


KerstinKeller:
### Context

We want to be able to do 
```
pip install ecal
```
for a seamless python integration.
However the wheels that we produce are not  PyPi ready.

### Proposal

We need to make a few changes to the build in order to make it pypi ready:

1. Drive the build from setup.py. Atm, CMake generates setup.py and then calls it. We need to reverse this process. We need setup.py that calls cmake. Only that way can we make source distributions of eCAL.
2. Use either https://scikit-build-core.readthedocs.io/en/latest/index.html or https://github.com/tttapa/py-build-cmake to drive the build.
3.  Setup build jobs that build releases (and then push to pypi).
     For linux build, build inside a manylinux docker image:
     https://github.com/pypa/manylinux

FlorianReimold:
For windows x64 we should be PyPi ready, as Windows has the most amazing binary compatibility of all Operating systems. So there really isn't an issue with publishing .whl files containing native .dlls

For Linux we currently depend on too many system packages, I guess. The .whl files are just not self contained enough for being functional on a wide variety of Linux distros.

DownerCase:
You've probably already done the investigation; but it is certainly possible to generate the pb files as part of setup.py, I've been able to do a PoC of that process. Mostly involves injecting an extra subcommand into the main build command.

You can look to Google's [protobuf distutils](https://pypi.org/project/protobuf-distutils/) and the newer (and pyproject compatible) [setuptools-protobuf](https://pypi.org/project/setuptools-protobuf/) projects for inspiration (if they're not directly suitable).

KerstinKeller:
@DownerCase thanks for this great hint! Setuptools-protobuf seems really promising.
We have to see if it plays together with py-build-cmake, but worst case, we'll make two different packages, one with the binary extensions, and one with the _pb2 files, but let's see..

DownerCase:
That said I was commenting from a pure python standpoint.

Given you’re calling into a CMake build system you should be able to set up an `add_custom_command` to call protoc with the normal `—python-out` argument to generate the python bindings as part of the build.

KerstinKeller:
The good thing is that during build, we already have protobuf available. That's challenging with `setuptools-protobuf` as the user has to provide a protoc (which needs to be compatible with the protobuf version that he wants to use).

TheCynosure:
[...]

PR #1395, opened 2024-02-25
[python] Don't link C extensions to libpython

DownerCase:
### Description

Fixes the Python C extensions having an explicit link dependency on libpython. In environments without libpython this causes problems. The easiest way to encounter this is through pyenv.
Eg: Building the wheel using a Python install that has libpython and trying to install and run it in a Python without it (static build).

The first hint of the problem is running `ldd` on the shared object when using a pyenv installation (for example):

``` 
ldd out/wheel/build/python/ecal/_ecal_core_py.so
        linux-vdso.so.1 (0x00007ffefb729000)
        libpython3.12.so.1.0 => not found
        ...
```

This will actually run fine if it is run in a Python interpreter that has a shared libpython as the wheels inherit the symbols from Python (see the PEP 513 link below). However; trying to run this eCAL wheel in a statically built Python will cause an error like:

```
python samples/python/benchmarks/latency_snd/latency_snd.py
Traceback (most recent call last):
  File "/home/downercase/dev/ecal/samples/python/benchmarks/latency_snd/latency_snd.py", line 24, in <module>
    import ecal.core.core as ecal_core
  File "/home/downercase/dev/ecal/.venv_312_static/lib/python3.12/site-packages/ecal/core/core.py", line 26, in <module>
    import ecal._ecal_core_py as _ecal
ImportError: libpython3.12.so.1.0: cannot open shared object file: No such file or directory
```

[PEP 513 supporting material](https://peps.python.org/pep-0513/#libpythonx-y-so-1)

Fixing this is one small step closer to redistributable wheels.

### Related issues

- This would be problematic for #1161 

### Cherry-pick to
- 5.12 (old stable)
- 5.13 (current stable)

### DIFF
--- a/lang/python/core/CMakeLists.txt
+++ b/lang/python/core/CMakeLists.txt
@@ -40,7 +40,6 @@ ecal_add_python_module(${PROJECT_NAME} SOURCES ${ecal_lang_py_src} PYTHON_CODE $
 
 target_link_libraries(${PROJECT_NAME}
     PRIVATE
-        Python::Python
         eCAL::core
         eCAL::core_pb
 )
--- a/lang/python/ecalhdf5/CMakeLists.txt
+++ b/lang/python/ecalhdf5/CMakeLists.txt
@@ -41,7 +41,6 @@ ecal_add_python_module(${PROJECT_NAME} SOURCES ${ecal_lang_py_src} PYTHON_CODE $
 
 target_link_libraries(${PROJECT_NAME}
     PRIVATE
-        Python::Python
         eCAL::hdf5
 )
 


DownerCase:
Update: Need to do some more testing/investigation on this... Looks like it may only be a CPython 3.8+ solution.
https://bugs.python.org/issue21536
https://github.com/python/cpython/pull/12946

KerstinKeller:
Does this actually work? I though linkage was required, but since the build is green, it's not necessary?
Maybe correct linkage is already assumed by using the CMake helper functions to create Python extensions?

However, in general, the whole Python build process should be triggered through Python instead of through CMake.
E.g. Python should trigger CMake. I started to investigate some Python build backends (scikit-build-core, ...) but in the end I somehow got stuck because I couldn't quite model the correct behavior.

It's definately on our list to have proper pip support and universal linux wheels and most effort should go in that direction.
But per se - if that works, let's just remove those two linking lines 😄

DownerCase:
> Does this actually work? I though linkage was required, but since the build is green, it's not necessary? Maybe correct linkage is already assumed by using the CMake helper functions to create Python extensions?

[Your helper here](https://github.com/eclipse-ecal/ecal/blob/0bc41abdf2348e7045de97228bf895ec4590559c/cmake/helper_functions/ecal_python_functions.cmake#L47) correctly calls `Python_add_library` with `module` which implicitly adds a dependency to [`Python::Module`](https://cmake.org/cmake/help/latest/module/FindPython.html#commands)

I reproduced the error on Linux using Python 3.12 and confirmed that removing the link fixes it for "build with shared -> use with static".

From the research it looks like this was only made possible with Python 3.8 so just need to check behaviour under 3.7 and re-verify Windows too.

> However, in general, the whole Python build process should be triggered through Python instead of through CMake. E.g. Python should trigger CMake. I started to investigate some Python build backends (scikit-build-core, ...) but in the end I somehow got stuck because I couldn't quite model the correct behavior.
> 
> It's definately on our list to have proper pip support and universal linux wheels and most effort should go in that direction. But per se - if that works, let's just remove those two linking lines 😄

This is definitely something I would love to see too. Happy to help out where possible to make it happen for eCAL 6 😄

DownerCase:
TLDR: Everything looks good 😄 

Windows:
Static Python builds on Windows isn't really a thing as far as I can tell. [eg](https://github.com/python/cpython/issues/110234#ref-pullrequest-1918151996)
This patch removes duplicate linking entries.
This is the only difference in the Ninja build rules. Left is before, right is after.
image

Ubuntu 22.04 LTS (WSL) with Python 3.7 built via pyenv:

Before:
Shared Python show links to `libpython*.so.1.0` as you'd expect
Static Pythons do not.
✅ Using static built wheel in shared
❌ Using shared built wheel in static

After:
✅ Static build wheel in shared
✅ Shared built wheel in static

Inspecting the build differences again we see a very similar thing, expect that the libpython linkage is completely removed.
image


To summarize everything.
- The `ecal_add_python_module` helper correctly calls `Python_add_library` with the `MODULE` argument.
- This creates a CMake library target correctly linked for Python C extensions (and includes the Python.h include directory).
- Prior to this patch eCAL was also linking again, explictly to `libpython` via `Python::Python`
- For Windows this had no real consequence.
- Under Linux this adds a dependency of the shared object on `libpython`
- As a result, any statically linked versions of Python (eg: custom source build) cannot use an eCAL wheel built by a shared (i.e: most package manager versions) Python build.
- This patch removes the unneeded dependency on `libpython` on Linux
- This is fine as C extensions inherit symbol definitions from the parent Python execution environment. (Technical terms probably not quite right, but you get the idea)

(This is probably the most effort I've put into a 2 line change 😆)

KerstinKeller:
Thanks for all the investigation that you put into this topic 👍 
On Windows you can really see, that the linker argument to the Python lib was present twice.
On Linux I am not quite sure I get the part about inheriting the symbols definitions, but as long, as they are there, that sounds good.

I will merge your PR, thanks for contributing!

KerstinKeller (approved):

PR #1406, opened 2024-02-28
[Qwt] Default to static build

### LINKED ISSUE #1379: Make sure Qt Versions of eCAL Mon GUI and Signal Plotting QWT plugin match
### Problem Description

When compiling eCAL master on Ubuntu 24.04 against Qt6, the eCAL Monitor crashes when trying to load the Signal Plotting plugin. The reason for that is, that the only available `libqwt` is a Qt5 version. Even though it compiles, it will still create a runtime issue.

```
[Ecalmon] Updating monitor
QLayout: Attempting to add QLayout "" to PluginWidget "PluginWidget", which already has a layout
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
44	./nptl/pthread_kill.c: No such file or directory
```

We probably have to introduce an option to compile QWT from the submodule and link statically against it.

### How to reproduce

- Use Ubuntu 24.04
- Install dependencies (notice the version missmatch between `qt6-base-dev` and `libqwt-qt5-dev`)
    ```
    sudo apt install qt6-base-dev
    sudo apt-get install git cmake doxygen graphviz build-essential zlib1g-dev
    sudo apt install libhdf5-dev libprotobuf-dev libprotoc-dev protobuf-compiler libcurl4-openssl-dev libqwt-qt5-dev libyaml-cpp-dev
    ```
- Compile eCAL
- Run eCAL Mon
- Open Reflection window

### How did you get eCAL?

### Environment

- ecal master from 2024-02-16
- Ubuntu 24.04

### eCAL System Information

_No response_

FlorianReimold:
### Description
- QWT can now be built static
- static is the default option
- GH Actions for Ubuntu now build QWT from the submodule
- This fixes issues when building with Qt6 on Linux

### Related issues
- Fixes #1379

### Cherry-pick to
- 5.13 (current stable)

### DIFF
--- a/cpack/innosetup/ecal_setup.iss.in
+++ b/cpack/innosetup/ecal_setup.iss.in
@@ -81,7 +81,6 @@ Source: "{#ComponentStagingDir}\configuration\cfg\ecaltime.ini"; DestDir: "{comm
 
 ; applications
 Source: "{#ComponentStagingDir}\app\*";               DestDir: "{app}";                   Flags: ignoreversion recursesubdirs;     Components: applications
-Source: "{#ComponentStagingDir}\qwt_runtime\*";       DestDir: "{app}";                   Flags: ignoreversion recursesubdirs;     Components: applications
 
 ; samples
 Source: "{#ComponentStagingDir}\samples\*";           DestDir: "{app}";                   Flags: ignoreversion recursesubdirs;     Components: samples


FlorianReimold:
Something seems to be wrong with the innosetup. I guess it cannot find qwt.dll anymore, as that is now statically integrated into the plugin.

KerstinKeller (approved):
👍

PR #1410, opened 2024-03-01
[core] deprecated proto fields ttype and tdesc removed

rex-schilasky:
### Description
1. The two protobuf `Topic` fields `ttype` (topic type name) and `tdesc` (topic descriptor) are deprecated (replaced by `tdatatype` (and its sub elements `name`, `encoding `and `desc`). For the future eCAL6 version they are finally removed. This affects:

* core / serialization / registration
* monitoring apps
* testing
* monitoring samples

2. After replacing internal google protobuf serialization with nanopb all core modules are using eCAL Sample structs instead protobuf objects. Based on that fact the new Sample struct element `DataTypeInformation` could be replaced by the existing `SDataTypeInformation`.

### Cherry-pick to
- _none_


app/mon/mon_tui/src/model/monitor.hpp
```diff
@@ -199,8 +199,8 @@ class MonitorModel
       topic.id = std::move(*t.mutable_tid());
       topic.name = std::move(*t.mutable_tname());
       topic.direction = TopicDirection(t.direction());
-      topic.type = std::move(*t.mutable_ttype());
-      topic.type_descriptor = std::move(*t.mutable_tdesc());
+      topic.type = std::move(*t.mutable_tdatatype()->mutable_name());
+      topic.type_descriptor = std::move(*t.mutable_tdatatype()->mutable_desc());
```
KerstinKeller, line 203: I'm not quite sure, it might be that in mon_tui, we're now missing the topic encoding?

app/rec/rec_client_core/src/monitoring_thread.cpp
```diff
@@ -91,7 +91,7 @@ namespace eCAL
             // Evaluate the quality of the current descriptor information
             int this_topic_info_quality = 0;
 
-            if (!topic.tdesc().empty())
+            if (!topic.tdatatype().desc().empty())
             {
```
KerstinKeller, line 95: I think in this file, I had the whole discussion with @FlorianReimold  - when changing it like this, the behavior is actually different. however, before we release 6.0, we will get rid of the topic qualities, so I'd vote that this is ok to merge atm.

app/rec/rec_server_core/src/monitoring_thread.cpp
```diff
@@ -129,15 +129,15 @@ namespace eCAL
             if (topic_info_map_it != topic_info_map_.end())
             {
               // Only update the values if there are information available
-              if (!topic.ttype().empty() || !topic.tdesc().empty())
+              if (!topic.tdatatype().name().empty() || !topic.tdatatype().desc().empty())
```
KerstinKeller, line 132: Same here as above. Technically, not the same as before. But should be unnecessary when descgate can store mutliple information.

ecal/core/src/monitoring/ecal_monitoring_impl.cpp
```diff
@@ -269,8 +269,8 @@ namespace eCAL
         }
       std::string topic_datatype_encoding = sample_topic.tdatatype.encoding;
       std::string topic_datatype_name     = sample_topic.tdatatype.name;
-      std::string topic_datatype_desc     = sample_topic.tdatatype.desc;
-      auto attr           = sample_topic.attr;
+      std::string topic_datatype_desc     = sample_topic.tdatatype.descriptor;
```
KerstinKeller, line 272: Can it now be called descriptor? i thought the name was blocked?

KerstinKeller (commented):
I have a few remarks which we should discuss before merging this.

ecal/core/src/monitoring/ecal_monitoring_impl.cpp
rex-schilasky: Yes, not in the proto definition but the structure element can be called `descriptor`.

app/mon/mon_tui/src/model/monitor.hpp
rex-schilasky: Changed back to old behavior by using old combined type = tdatatype.encoding:tdatatype.name.

app/rec/rec_client_core/src/monitoring_thread.cpp
rex-schilasky: The tdesc here is okay, the other ones are changed back to old behavior.

app/rec/rec_server_core/src/monitoring_thread.cpp
rex-schilasky: Changed back to old behavior by using old combined type = tdatatype.encoding:tdatatype.name.

KerstinKeller (approved):
Now it looks good!

PR #1414, opened 2024-03-01
[core v5.13] deprecate function declaration added

rex-schilasky:
### Description

- deprecate function declaration added
- minor fix in windows build batch file (HAS_QT instead HAS_QT5)


FlorianReimold (approved):
Changes don't really fit together in 1 commit, but whatever 😋

PR #1422, opened 2024-03-06
set unit name empty string and nullptr handling fixed

rex-schilasky:
### Description
SetUnitName should not be able to set the eCAL global unit name to an empty string and nullptr argument should be handled as well.

### Cherry-pick to
- 5.12 (old stable)
- 5.13 (current stable)


hannemn (approved):
It's okay!

PR #1423, opened 2024-03-06
[core] core c test added / core cpp test cleaned up

rex-schilasky:
### Description
eCAL C API is currently not tested. Therefore a core C API test is added and the existing C++ API test is aligned with that.

### Cherry-pick to
- _none_

### DIFF
--- a/ecal/tests/c/core_test/src/core_test.cpp
+++ b/ecal/tests/c/core_test/src/core_test.cpp
@@ -1,0 +1,123 @@
+/* ========================= eCAL LICENSE =================================
+ *
+ * Copyright (C) 2016 - 2019 Continental Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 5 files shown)


ecal/tests/c/core_test/src/core_test.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+  // finalize eCAL API we expect return value 0 because it will be finalized
+  EXPECT_EQ(0, eCAL_Finalize(0));
+}
+
+TEST(core_c, Core_eCAL_Ok)
+{
+  // check uninitialized eCAL, should not be okay
+  EXPECT_EQ(0, eCAL_Ok());
+
```
KerstinKeller, line 111: Nothing to change atm, but it's super counter intuitive, that `eCAL_Initialize()` returns `0` upon success, but on the other hand `eCAL_Ok()` returns `0` upon failure, and `1` upon success (e.g. boolean values casted..).

So one behaves like a C style function while the other one doesn't....

KerstinKeller (approved):
:+1:

PR #1429, opened 2024-03-07
[core] create FrequencyCalculator class for sender / receiver frequency calculation.

### LINKED ISSUE #1404: Frequency column in eCAL Monitor is showing weird values for slow publishers
### Problem Description

grafik

I created a simple Python publisher that publishes machine state from my machines at home like memory and CPU consumption. I configured the publisher to send out the data roughly once per second, which probably corresponds to a frequency of less than one Hz in reality.

Now the Frequency column in the eCAL Monitor toggles slowly between 0 and 0.2, which are both not really helpful values.

While I understand the 0 for cases in which no message is published during the calculation window for the frequency, the 0.2 makes no sense at all to me.

This is not a big issue, but maybe it's somehow possible to have the calculation window to adjust itself automatically so that the values for slow publishers are more accurate?

### How to reproduce

1. Create a very slow (Python) publisher
2. Check it's frequency in eCAL Monitor

### How did you get eCAL?

### Environment

- eCAL Version: 5.12.4
- OS: Arch Linux
- Python Version: 3.11

### eCAL System Information

```text
------------------------- SYSTEM ---------------------------------
Version                  : v5.12.4 (2024-02-08 09:01:15 +0100)
Platform                 : linux

------------------------- CONFIGURATION --------------------------
Default INI              : /etc/ecal/ecal.ini

------------------------- NETWORK --------------------------------
Host name                : jtb-desktop
Host group name          : jtb-desktop
Network mode             : local
Network ttl              : 2
Network sndbuf           : 5 MByte
Network rcvbuf           : 5 MByte
Multicast cfg version    : v1
Multicast group          : 239.0.0.1
Multicast mask           : 0.0.0.15
Multicast ports          : 14000 - 14010
Multicast join all IFs   : off
Bandwidth limit (udp)    : not limited

------------------------- TIME -----------------------------------
Synchronization realtime : "ecaltime-localtime"
Synchronization replay   : 
State                    :  synchronized 
Master / Slave           :  Master 
Status (Code)            : "everything is fine." (0)

------------------------- PUBLISHER LAYER DEFAULTS ---------------
Layer Mode INPROC        : off
Layer Mode SHM           : auto
Layer Mode TCP           : off
Layer Mode UDP MC        : auto

------------------------- SUBSCRIPTION LAYER DEFAULTS ------------
Layer Mode INPROC        : on
Layer Mode SHM           : on
Layer Mode TCP           : on
Layer Mode UDP MC        : on
Npcap UDP Reciever       : off

------------------------- EXPERIMENTAL ---------------------------
SHM Monitoring           : off
SHM Monitoring (Domain)  : ecal_mon
SHM Monitoring (Queue)   : 1024
Network Monitoring       : on
Drop out-of-order msgs   : off
```

KerstinKeller:
### Description
The frequency calculation for reader / writer frequencies was moved from reader.cpp / writer.cpp into it's own class.
The calculator properly handles frequencies below 1Hz and is able to reset if no data is coming in.

### Discussion
The calculator is not thread safe. Therefore mutexes are needed for synchronization.

The non-resettable calculator can probably be made threadsafe by moving the two members with are accessed concurrently into a struct with a size of < 8 bytes and proper memory alignment. Both functions could atomically load/store the values and a lock-free thread safety can be achieved.

The resettable calculator sees a bit harder. The unique_ptr needs to be handled and atomic unique_pointers are not available. That pointer may be created in `addTick` and destroyed in `getFrequency`, and I have low hopes of getting that part threadsafe easily without using locks.

### Related issues
Fixes #1404 

### Cherry-pick to
To be discussed, potentially 5.13, as it is not yet released.
- 5.13 (current stable)

### DIFF
--- a/ecal/core/src/readwrite/ecal_reader.cpp
+++ b/ecal/core/src/readwrite/ecal_reader.cpp
@@ -70,8 +70,7 @@ namespace eCAL
                  m_read_time(0),
                  m_receive_time(0),
                  m_clock(0),
-                 m_clock_old(0),
-                 m_freq(0),
+                 m_frequency_calculator(3.0f),
                  m_message_drops(0),
                  m_loc_published(false),
                  m_ext_published(false),
@@ -97,10 +96,7 @@ namespace eCAL
     m_topic_name    = topic_name_;
     m_topic_id.clear();
     m_topic_info    = topic_info_;
-    m_clock         = 0;
-    m_clock_old     = 0;
     m_message_drops = 0;
-    m_rec_time      = std::chrono::steady_clock::time_point();
     m_created       = false;
 #ifndef NDEBUG
     // log it
@@ -163,11 +159,7 @@ namespace eCAL
 
     // reset defaults
     m_created                 = false;
-    m_clock                   = 0;
-    m_clock_old               = 0;
-    m_freq                    = 0;
     m_message_drops           = 0;
-    m_rec_time                = std::chrono::steady_clock::time_point();
 
     m_use_udp_mc_confirmed    = false;
     m_use_shm_confirmed       = false;
@@ -299,7 +291,10 @@ namespace eCAL
     ecal_reg_sample_topic.pname         = m_pname;
     ecal_reg_sample_topic.uname         = Process::GetUnitName();
     ecal_reg_sample_topic.dclock        = m_clock;
-    ecal_reg_sample_topic.dfreq         = m_freq;
+    {
+      std::lock_guard<std::mutex> lock(m_frequency_calculator_mutex);
+      ecal_reg_sample_topic.dfreq = static_cast<int32_t>(m_frequency_calculator.getFrequency(std::chrono::steady_clock::now()) * 1000);
+    }
     ecal_reg_sample_topic.message_drops = static_cast<int32_t>(m_message_drops);
 
     // we do not know the number of connections ..
@@ -476,6 +471,12 @@ namespace eCAL
 
     // increase read clock
     m_clock++;
+
+    // Update frequency calculation
+    {
+      std::lock_guard<std::mutex> freq_lock(m_frequency_calculator_mutex);
+      m_frequency_calculator.addTick(std::chrono::steady_clock::now());
+    }
 
     // reset timeout
     m_receive_time = 0;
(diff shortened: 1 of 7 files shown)


ecal/core/src/readwrite/ecal_reader.cpp
```diff
@@ -477,6 +474,12 @@ namespace eCAL
     // increase read clock
     m_clock++;
 
+    // Update frequency calculation
+    {
+      std::lock_guard<std::mutex> freq_lock(m_frequency_calculator_mutex);
```
FlorianReimold, line 479: Should be const

ecal/core/src/readwrite/ecal_reader.cpp
```diff
@@ -299,7 +293,10 @@ namespace eCAL
     ecal_reg_sample_topic.pname         = m_pname;
     ecal_reg_sample_topic.uname         = Process::GetUnitName();
     ecal_reg_sample_topic.dclock        = m_clock;
-    ecal_reg_sample_topic.dfreq         = m_freq;
+    {
+      std::lock_guard<std::mutex> lock(m_frequency_calculator_mutex);
```
FlorianReimold, line 297: Should be const

ecal/core/src/readwrite/ecal_writer.cpp
```diff
@@ -408,6 +401,12 @@ namespace eCAL
 
   size_t CDataWriter::Write(CPayloadWriter& payload_, long long time_, long long id_)
   {
+    {
+      // we should think about if we would like to potentially use the `time_` variable to tick with (but we would need the same base for checking incoming samples then....
+      std::lock_guard<std::mutex> lock(m_frequency_calculator_mutex);
+      m_frequency_calculator.addTick(std::chrono::steady_clock::now());
```
FlorianReimold, line 407: It would technically be better to grab the current time before aquiring the mutex. But I guess in this case it doesn't matter

ecal/core/src/readwrite/ecal_writer.cpp
```diff
@@ -770,6 +743,12 @@ namespace eCAL
 
   std::string CDataWriter::Dump(const std::string& indent_ /* = "" */)
   {
+    double frequency;
+    {
+      std::lock_guard<std::mutex> lock(m_frequency_calculator_mutex);
+      frequency = m_frequency_calculator.getFrequency(std::chrono::steady_clock::now()) * 1000000;
```
FlorianReimold, line 749: Why is this in µHz, while all other values are in mHz?

ecal/core/src/readwrite/ecal_writer.cpp
```diff
@@ -785,6 +764,7 @@ namespace eCAL
     out << indent_ << "m_topic_info.desc:        " << m_topic_info.descriptor << '\n';
     out << indent_ << "m_id:                     " << m_id << '\n';
     out << indent_ << "m_clock:                  " << m_clock << '\n';
+    out << indent_ << "frequency [mHz]:          " << frequency << '\n';
```
FlorianReimold, line 767: It says mHz here, but you multiply with 1,000,000, which should result in µHz

FlorianReimold (approved):

PR #1433, opened 2024-03-07
[core] gtest testsuite names harmonized

rex-schilasky:
### Description
All ecal-core gtest and ecal-service gtest testsuite names harmonized to get better ordering (e.g. Visual Studio Test Explorer) and easier maintenance.

image

### Cherry-pick to
- _none_


ecal/tests/CMakeLists.txt
```diff
@@ -44,7 +46,9 @@ if(ECAL_CORE_REGISTRATION AND ECAL_CORE_PUBLISHER AND ECAL_CORE_SUBSCRIBER)
     add_subdirectory(cpp/pubsub_test)
   endif()
   if(ECAL_CORE_BUILD_TESTS_PROTOBUF)
-    add_subdirectory(cpp/pubsub_proto_test)
+    add_subdirectory(cpp/pubsub_proto_dyn_test)
```
KerstinKeller, line 49: Does it make sense really to split them? Pub doesn't make sense without sub. maybe different cpp files, but different executables seems a bit overkill?

ecal/tests/cpp/pubsub_proto_dyn_test/src/proto_dyn_subscriber_test.cpp
```diff
@@ -63,6 +65,7 @@ class ProtoDynSubscriberTest : public ::testing::Test {
 
   std::atomic<int> received_callbacks;
 };
+using core_cpp_pubsub_proto = ProtoDynSubscriberTest;
```
KerstinKeller, line 68: What's with the `using` here? are you not renaming them now? does it even matter?

ecal/tests/cpp/pubsub_proto_pub_test/src/protobuf/animal.proto
```diff
@@ -0,0 +1,28 @@
+/* ========================= eCAL LICENSE =================================
```
KerstinKeller, line 1: now we have yet two more `animal.proto`, `person.proto` and `house.proto`. It's unnecessary. I don't want to count how many times we have them in the repo already.

KerstinKeller (commented):
In general, I'd vote rather for less distinct CMake projects than more.
Thanks to tests explorers, you can trigger individual tests, also the command line lets you do that.

Maybe even go so far: per one library (core) - one test executable, with multiple cpp files.
This also makes it way easier to add tests and group them logically.

If need be, we can have one test executable for public API testing, and then one test executable for internal utilities.
And maybe one more for each language extensions (e.g. protobuf pubsub, string pubsub)...

ecal/tests/cpp/pubsub_proto_dyn_test/src/proto_dyn_subscriber_test.cpp
rex-schilasky: Bringing the test cases in the test suite name "core_cpp_pubsub_proto"

ecal/tests/CMakeLists.txt
rex-schilasky: GTest does not allow to have multiple test base classes with the same name in one executable, no matter if I make them static or in an separate namespace. IF I want to use common test suite names for all tests I need either to not use test base classes, or split the tests in separate executables.

ecal/tests/cpp/pubsub_proto_pub_test/src/protobuf/animal.proto
rex-schilasky: I would vote for a separate test_proto library that provides this logic to all tests.

KerstinKeller (approved):

PR #1435, opened 2024-03-07
[core] new public api getlogging as struct

rex-schilasky:
### Description
New public API function int GetLogging(Logging::SLogging& log_) to get log messages as struct without the need to serialize and deserialize it via protobuf.

### Cherry-pick to
- _none_


KerstinKeller (approved):
Looks good to me. As discussed, we'll take a look at the public API when we get closer to an eCAL6 release.

PR #1437, opened 2024-03-08
Cleanup: Remove a ton of unnecessary ";" (that cause pedantic warnings).

KerstinKeller:
### Description
Apply the patch to remove `;` that cause pedantic warnings with gcc.


rex-schilasky (approved):
LGTM, let's check if we get any other issues in the toolchain later based on the correction of these magic comment lines starting with ';'.

PR #1439, opened 2024-03-08
[python] python api fixed to work again based on new c++/c core api

rex-schilasky:
### Description
The new c++ core api is missing a few functions (set/get_qos, set_layer, set_type_name, set_description ..) and has some new functionalies like separated encoding and type naming.

This PR is fixing the python wrapper that way, that all samples are working again almost without any changes (two samples latency_rec/rec_cb needed small adaptions because of different (now correct) behavior of ecal_ok()).

### Cherry-pick to
- _none_


KerstinKeller (approved):
:+1:

PR #1444, opened 2024-03-11
[core] Silence deprecation warnings in core(_c) build.

KerstinKeller:
### Description
No deprecation warnings during eCAL build.


rex-schilasky (approved):
LGTM, we need a second cleanup step regarding ecal_clang.h/cpp 👍

PR #1445, opened 2024-03-11
[samples/tests] folder structure cleaned up

rex-schilasky:
### Description
ecal-core tests and samples and ecal main solution tests and samples did not reflect in the same order inside a VS solution. Now they do

image
.


ecal/samples/c/pubsub/string/minimal_snd/CMakeLists.txt
```diff
@@ -40,4 +40,4 @@ target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_14)
 
 ecal_install_sample(${PROJECT_NAME})
 
-set_property(TARGET ${PROJECT_NAME} PROPERTY FOLDER core/samples/c/pubsub/string)
+set_property(TARGET ${PROJECT_NAME} PROPERTY FOLDER samples/c/pubsub/string)
```
KerstinKeller, line 43: shouldn't it be `samples/c/core/pubsub/string` now?

KerstinKeller (approved):
I wish test instead of tests, but at some point we can do it 😄

Issue #1187, opened 2023-08-21
Support newer Protobuf


FlorianReimold:
### Context

Currently, we have Protobuf 3.11 in the submodule, which is used for Windows. That version is quite old and newer versions would have better support for C#. Also, the old version of protobuf currently forces users to install old python protobuf wheels, as we ship the eCAL wheel with pre-generated pb2.py files.

### Proposal

The goal here is to update the protobuf submodule to something more recent.

### Tasks and updates

- [ ] Update Protobuf to 3.21 in the submodule
- [ ] Fix Windows build
- [ ] Test python wheels

FlorianReimold:
Looks like C# doesn't like the new protobuf. Maybe we just need to re-generate the files?

```
D:\a\ecal\_build\complete\lang\csharp\samples\pubsub\protobuf\protobuf\protobuf\animal.cs(41,13): error CS0234: The type or namespace name 'IBufferMessage' does not exist in the namespace 'Google.Protobuf' (are you missing an assembly reference?) [D:\a\ecal\_build\complete\lang\csharp\samples\pubsub\protobuf\protobuf\person_protobufs_cs.csproj]
D:\a\ecal\_build\complete\lang\csharp\samples\pubsub\protobuf\protobuf\protobuf\house.cs(41,13): error CS0234: The type or namespace name 'IBufferMessage' does not exist in the namespace 'Google.Protobuf' (are you missing an assembly reference?) [D:\a\ecal\_build\complete\lang\csharp\samples\pubsub\protobuf\protobuf\person_protobufs_cs.csproj]
D:\a\ecal\_build\complete\lang\csharp\samples\pubsub\protobuf\protobuf\protobuf\house.cs(150,14): error CS0234: The type or namespace name 'IBufferMessage' does not exist in the namespace 'Google.Protobuf' (are you missing an assembly reference?) [D:\a\ecal\_build\complete\lang\csharp\samples\pubsub\protobuf\protobuf\person_protobufs_cs.csproj]
D:\a\ecal\_build\complete\lang\csharp\samples\pubsub\protobuf\protobuf\protobuf\house.cs(210,14): error CS0234: The type or namespace name 'IBufferMessage' does not exist in the namespace 'Google.Protobuf' (are you missing an assembly reference?) [D:\a\ecal\_build\complete\lang\csharp\samples\pubsub\protobuf\protobuf\person_protobufs_cs.csproj]
[...]

PR #1454, opened 2024-03-12
[python] Scikit-build-core driven wheel creation

### LINKED ISSUE #700: Cannot build Python wheel for RelWithDebugInfo on Windows
I cannot build a python wheel for that configuration.
1. The pyd files end up in the wrong directory
2. They are not contained in the final wheel

image

### LINKED ISSUE #1394: Rename python binding from pyhton3-ecal5 to eclipse-ecal
### Context

With eCAL 6.0 being on its way, we should drop the "5" in the name

### Proposal

Rename python binding to eclipse-ecal (whl) and python3-eclipse-ecal (apt)

### Tasks and updates

_No response_

DownerCase:
### Description

Python wheels are now built via [scikit-build-core](https://scikit-build-core.readthedocs.io/en/latest/index.html)! 🎉 
This work is based off of Kerstin's attempt a few months ago. 

**You should have write permissions to this branch if you wish to make modifications 😄** 

Changes of this PR:

#### Removed
- BREAKING CHANGE: Removed all protobuf bindings. Shipping protoc generated python code is difficult if not impossible to do whilst being maximally compatible. 
  - Eg: protocolbuffers/protobuf#11123
- Old wheel mechanism removed and replaced with scikit-build-core
- The CMake option `BUILD_STANDALONE_PY_WHEEL` as it's not longer needed
- `ecal_python_functions.cmake` as no functions from it are used now

#### Updated
- Runtime dependencies (ecal_core, hdf5) copied at install time (not build time) via CMake's `install` command
- Wheels are built and uploaded by [cibuildwheel](https://cibuildwheel.pypa.io/en/stable/)
- CMake 3.18 or newer is required if building the Python modules.
  - Access to the `Development.Module` component is strictly required for this workflow.
  - This is only enforced if `BUILD_PY_BINDING` is on.
- The "wheel" CMake preset has been renamed to "python" and updated to reflect how wheels are now built (i.e: static where possible)
- The eCAL version may be provided by scikit-build-core via the `SKBUILD_PROJECT_VERSION` CMake variable, this is the only change to the CMake scripts that interacts with scikit-build-core.
- protobuf >=3.8 but < 4.X is listed as dependency.
  - By not shipping generated code, greater compatibility is possible. Should be able to extend to the python protobuf packages v4 and v5?

#### Added
- The wheels now contain an autogenerated `_version.py` file containing the package version information
- CMake has a new `python` component which has to be explicitly listed to CPack to be installed.

#### Identified issues:
- [x] [setuptools-scm](https://setuptools-scm.readthedocs.io/en/latest/) which provides the dynamic version metadata rejects the `v6.0.0-nightly` tag.
  - I fixed this by creating a `v6.0.0.dev0` tag to mark "start of v6 development". The tools were happy with that.
  - Probably related: pypa/setuptools_scm#825
  - **Resolved**: New tag pushed

- [x] How to handle the protobuf bindings.
  - I don't understand where the core proto messages would be used, especially after #1439 
  - IMO the apps are the most likely ones users could want / be using.
  - In a pinch users can grab the .proto files and compile themselves to have generated code that matches the versions they want to use.
  - > **Resolved**: [Deemed not an issue](https://github.com/eclipse-ecal/ecal/pull/1454#issuecomment-1994167540)
- [ ] The wheel includes multiple versioned SO's on Linux, namely the [SOVERSION](https://cmake.org/cmake/help/latest/prop_tgt/SOVERSION.html) one and the [VERSION](https://cmake.org/cmake/help/latest/prop_tgt/VERSION.html) one. Eg:
  - `libecal_core.so.6` and `libecal_core.so.6.0.0`
  - `libhdf5.so.103.4.1` and `libhdf5.103`
  - I have no good idea on how to resolve this, at least perhaps until scikit-build/scikit-build-core#652 is in a release and we can exclude `*.so.*.*.*` from the wheel?
- [ ] cibuildwheel has been restricted to one architecture type (64-bit) as otherwise the second architecture to get built has linking problems
  - Somehow we need to tell CMake the architecture
  - Eg: https://github.com/pypa/cibuildwheel/discussions/844#discussion-3589343
- [ ] cibuildwheel does not build musl wheels. eCAL does not compile with musl libc.

### Related issues

- Partly resolves #1394, the deb packaging isn't in this repo.
- Fixes #700 by deleting the problematic code

Related but may not close:
- #1187
- #1161


DownerCase:
Ok. I'm being silly, I literally wrote that this won't work here due to the "v6.0.0-nightly" tag.

```packaging.version.InvalidVersion: Invalid version: 'v6.0.0-nightly'```

**Comment by KerstinKeller** [comment]
Hi, thanks so much for your PR! This is great, especially that you managed to pull in the dynamic versions 😄 
For me it's ok, not to ship the compiled protos.

We will try to retag with an alpha tag. But to my understanding, shouldn't 'v6.0.0-nightly' still be semver compliant?

**Comment by DownerCase** [comment]
> Hi, thanks so much for your PR, especially the great documentation of all issues. I really tried gettings things to work myself for a day or two, but I ran into so many issues, that I moved on to other topics.

Yup, this took a lot longer than I expected too, but worth it.

> We will try to retag with an alpha tag. But to my understanding, shouldn't 'v6.0.0-nightly' still be semver compliant?

Yes, but its not Python version specification compliant.
setuptools-scm uses [this regex](https://setuptools-scm.readthedocs.io/en/latest/config/#setuptools_scm._config.DEFAULT_TAG_REGEX) by default to match PEP440 style versions so for Python its strictly enforced for the version to meet [this format](https://packaging.python.org/en/latest/specifications/version-specifiers/#public-version-identifiers) (`[N!]N(.N)*[{a|b|rc}N][.postN][.devN]`). There is another section further down that talks about schemes for [developmental release](https://packaging.python.org/en/latest/specifications/version-specifiers/#developmental-releases) (eg: X.Y.devN)

**Comment by DownerCase** [comment]
@KerstinKeller  Do you need anything further from me here? Everything is passing green now with the new tag (thank you!) except the cherry picker due to the missing label. If some pytests would lighten the validation load for you all and help get this merged I can do that, no problem.

**Comment by KerstinKeller** [comment]
Sorry I was really busy last week. The PR looks already really good. We wanted to do some lokal tests before merging.
Pytests would totally make sense (we can also help write some), I guess they can also be executed as part of the wheel building process?
Anyways, we really appreciate your contribution, and we will likely merge it later this weeks.

**Comment by KerstinKeller** [comment]
One more quick question: is there a reason that the ubuntu builds are done on Ubuntu 20.04 which is soon going to be deprecated. I think cibuildwheel will execute the actual build in a docker container? Or did I get something wrong here?

**Comment by DownerCase** [comment]
> One more quick question: is there a reason that the ubuntu builds are done on Ubuntu 20.04 which is soon going to be deprecated. I think cibuildwheel will execute the actual build in a docker container? Or did I get something wrong here?

Correct. cibuildwheel does all build in manylinux docker containers. https://cibuildwheel.pypa.io/en/stable/#how-it-works
Those platforms are chosen because I didn't change them when importing what you had previously done.

Will update to ubuntu-latest and windows-latest as it shouldn't affect anything :smile:

**Comment by DownerCase** [comment]
> Sorry I was really busy last week. The PR looks already really good. We wanted to do some lokal tests before merging. Pytests would totally make sense (we can also help write some), I guess they can also be executed as part of the wheel building process? Anyways, we really appreciate your contribution, and we will likely merge it later this weeks.

No worries. Yes, the tests would run as part of cibuildwheel for every wheel

**Comment by FlorianReimold** [comment]
I tested the whl files for Windows Python 3.9 and Ubuntu 22.04 Python 3.10 and everything worked out of the box. I tested the HelloWorld and HelloWorld Protobuf sample from the documentation. The systems were clean and didn't have eCAL installed.

File: CMakeLists.txt [hunk 1]
```diff
@@ -159,7 +158,18 @@ option(CPACK_PACK_WITH_INNOSETUP               "Create Innosetup installer for t
 
 set(ECAL_INSTALL_PYTHON_DIR "bin" CACHE PATH "Location to install the Python extension modules. Might be set by setupdtools install.")
 
-set(ECAL_BUILD_VERSION "0.0.0" CACHE STRING "Inject a build version if not building from a git repository")
+if(DEFINED SKBUILD_PROJECT_VERSION)
```
KerstinKeller, line 161: I'm not a 100% friend of this, i'd prefer if the build could just inject `ECAL_BUILD_VERSION` from the outside.
But this may be cleaned up later on

lang/python/CMakeLists.txt
```diff
@@ -16,54 +16,34 @@
...
-
-if (Protobuf_VERSION VERSION_LESS "3.19")
-  set(Protobuf_required_versions ">=${Protobuf_VERSION},<=3.20")
-else()
-  set(Protobuf_required_versions ">=${Protobuf_VERSION}")
-endif()
+# 3.18 required for Development.Module.
+# prior versions using just Development would also search for artifacts for
+# embedding, which the manylinux containers don't provide
+cmake_minimum_required(VERSION 3.18...3.26)
```
KerstinKeller, line 22: Do you think we should limit the python version to max 3.26? Or should we assume that CMake sticks to not breaking functionality?

KerstinKeller (approved):
I added two remarks, but I think they should prevent us from merging the PR.

CMakeLists.txt
DownerCase: As far as I could tell, this is the only way to get the package version from setuptools-scm into CMake, let scikit-core handle it.

lang/python/CMakeLists.txt
DownerCase: Thankfully, that's not what it means. The 3.26 just says, we support the new versions of all policies introduced up to 3.26.

I had just copied it from the documentation. Technically one _should_ specify a max policy version, like here, and test with that to get the changes from new CMake versions without raising the minimum version.

DownerCase: That is to say, this still runs with the latest 3.29 CMake, it just won't use any new features/policies from CMake >3.26

FlorianReimold:
Hi @DownerCase,

I tried to build the python wheel locally on my PC and realized that I have no idea how to do that. Is there a guide somewhere how that can be done?

DownerCase:
There are a few methods but I do

```sh
python -m pip install build
python -m build
```

cibuildwheels can also be run locally if you desire.

https://packaging.python.org/en/latest/tutorials/packaging-projects/#generating-distribution-archives

FlorianReimold:
Thanks, works like a charm!

Currently, we have the following python packages:
- .whl files for Ubuntu amd64 and Windows amd64
- .deb files for Ubuntu amd64 / x86 / arm64 / armhf

I checked whether I can reuse the old scripts to generate the .deb packages and that did not work (as expected). While I wouldn't really mind dropping the amd64 .deb installers, I would still like to offer the arm64 users a way to install the python binding.

So, do you by any chance know how to
a) build arm64 whl files with the GH Actions or
b) build .deb packages for this python binding?

In the past I used the following script for generating the deb package:

```
override_dh_auto_install-arch:
	dh_auto_install --arch

	# Python install.
	set -e; \
	cd _build/python && for python in $(shell py3versions -r); do \
		$$python setup.py install \
			--install-layout=deb --no-compile \
			--root=$(CURDIR)/debian/python3-ecal5; \
	done
	find $(CURDIR)/debian/python3-ecal5 -name 'ecal-*-nspkg.pth' -delete
```

DownerCase:
> a) build arm64 whl files with the GH Actions 

Yes! cibuildwheel can do it: [docs](https://cibuildwheel.pypa.io/en/stable/faq/#emulation). I'll set that up here too

 >b) build .deb packages for this python binding?

That is outside of my knowledge. Theoretically that method could work on new enough versions of pip to support the build backend standard. You may need to add the minimal stub setup.py.

DownerCase:
Hmm. Over an hour to get all green CI ticks is unacceptable. Will rework to only trigger the emulated compilation on releases? IMO, I only need to beat the 37m windows build for daily PR workflows.

FlorianReimold:
Is there a way to have them trigger on releases and on "manual trigger" or something similar? I was also thinking about the same thing.

DownerCase:
> Is there a way to have them trigger on releases and on "manual trigger" or something similar? I was also thinking about the same thing.

Theoretically I've done just that now :smile: It's set to only do the arm64 wheels on (pre-)releases being published or when the workflow is manually started.
See here for how to manually trigger a workflow: https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow

FlorianReimold (approved):

PR #1455, opened 2024-03-12
[Doc] Updated book theme to 1.1.2

### LINKED ISSUE #1303: Update version of sphinx-book-theme
### Context

[sphinx-book-theme](https://github.com/executablebooks/sphinx-book-theme) is fixed to an old version.
This leads to more and more problems with newer sphinx related packages, that lead to version conflicts.

We need to try and upgrade that package, such that we can use newer sphinx version.
This might require manual adaptions to some auto-generated files, to keep the documentation in eCAL orange.

### Proposal

Go to doc/requirements.txt and remove version requirements. Try and build documentation.

### Tasks and updates

_No response_

FlorianReimold:
### Description
Updated the Sphinx-book-theme

Fixes #1303


KerstinKeller (commented):
Nice it looks like you're able to remove a lot of code this way 👍

FlorianReimold:
I mostly removed autogenerated css code that I had to modify after generation. But that's still a good thing and probably makes upgrading to future releases way easier.
But it looks like we have python 3.9 now as minimum required version. Ubuntu 20.04 by default comes with python 3.8, but as python 3.9 is installable we should be fine.

rex-schilasky (approved):

PR #1456, opened 2024-03-13
include/ecal/ecal_clang.h deprecated

rex-schilasky:
### Description
Public header ecal_clang.h deprecated. Will be finally removed after renewal of the python API wrapper.


KerstinKeller (commented):
Does this cause deprecation warnings? only in python build?

rex-schilasky:
Yes, it's nowhere else included.

FlorianReimold (approved):

PR #1457, opened 2024-03-13
[core] ecal_clang.cpp/h moved to python wrapper

rex-schilasky:
### Description
ecal_clang.h and ecal_clang.cpp moved from ecal core to python wrapper project


KerstinKeller (approved):
Makes a lot of sense to move it to python!

PR #1459, opened 2024-03-13
[core] Bugfix: descgate incorrectly locked with shared locks.

KerstinKeller:
### Description
Incorrectly used shared_locks, when internally, data was modified.
Also deprecate the map on all function calls.


rex-schilasky:
Recreated as PR #1462 on master to cherry pick later on v5.13

PR #1464, opened 2024-03-13
[doc] v5.13 deprecation.rst fixed

rex-schilasky:
### Description
Syntax fixes for `versions/5.13/deprecation.rst`.


FlorianReimold (changes requested):
I think the indentations are wrong. Indentations are part of the syntax. It's usually a good idea to check the compiled output in a preview. I have written a tutorial how that can be done: https://eclipse-ecal.github.io/ecal/advanced/documentation.html

FlorianReimold (approved):

PR #1476, opened 2024-03-18
[core] description gate refactoring

rex-schilasky:
### Description
DescGate logic refactored:

- SHM and UDP registrations collected in common SampleList
- ApplyDescription/Type whole quality logic moved to desc_gate.cpp unit
- eCAL Initialize flag ProcessReg removed (should be set always)

### Related issues
Needs to be written.

### DIFF
--- a/ecal/core/src/registration/ecal_registration_provider.cpp
+++ b/ecal/core/src/registration/ecal_registration_provider.cpp
@@ -104,8 +104,7 @@ namespace eCAL
     // stop cyclic registration thread
     m_reg_sample_snd_thread->stop();
 
-    // prepare unregistration
-    ClearSampleList();
+    // add process unregistration sample
     AddSample2SampleList(GetProcessUnregisterSample());
 
     if (m_use_registration_udp)
--- a/ecal/core/src/registration/ecal_registration_provider.cpp
+++ b/ecal/core/src/registration/ecal_registration_provider.cpp
@@ -43,58 +43,11 @@
 #include <mutex>
 #include <string>
 
-namespace
-{
-  // TODO: remove me with new CDescGate
-  bool ApplyTopicDescription(const std::string& topic_name_, const eCAL::SDataTypeInformation& topic_info_, bool topic_is_a_publisher_)
-  {
-    if (eCAL::g_descgate() != nullptr)
-    {
-      // calculate the quality of the current info
-      eCAL::CDescGate::QualityFlags quality = eCAL::CDescGate::QualityFlags::NO_QUALITY;
-      if (!topic_info_.encoding.empty() || !topic_info_.name.empty())
-        quality |= eCAL::CDescGate::QualityFlags::TYPE_AVAILABLE;
-      if (!topic_info_.descriptor.empty())
-        quality |= eCAL::CDescGate::QualityFlags::DESCRIPTION_AVAILABLE;
-      if (topic_is_a_publisher_)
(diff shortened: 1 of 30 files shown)


ecal/core/src/registration/ecal_registration_provider.cpp
```diff
@@ -173,143 +131,171 @@ namespace eCAL
...
-    SampleMapT::iterator iter;
-    const std::lock_guard<std::mutex> lock(m_topics_map_sync);
-    iter = m_topics_map.find(topic_name_ + topic_id_);
-    if(iter != m_topics_map.end())
-    {
-      m_topics_map.erase(iter);
-      return(true);
+      // broadcast (updated) sample list over shm
+      SendSampleList2SHM();
     }
```
KerstinKeller, line 158: Don't you need `#ifdef` here like above? (`#if ECAL_CORE_REGISTRATION_SHM`,...)

ecal/core/src/registration/ecal_registration_provider.cpp
```diff
@@ -173,143 +131,171 @@ namespace eCAL
...
+#if ECAL_CORE_SERVICE
+    // refresh server registration
+    if (g_servicegate() != nullptr) g_servicegate()->RefreshRegistrations();
+
+    // refresh client registration
+    if (g_clientgate() != nullptr) g_clientgate()->RefreshRegistrations();
+#endif
+
+    // send out sample list over udp
+    SendSampleList2UDP();
```
KerstinKeller, line 282: again, here you need the `#if ...` ....

KerstinKeller: Btw, here would be a great point for a new refactoring (not in this PR!) - Have a list of Registration Layers, over which you iterate - this saves you the `#if` in many places.

ecal/core/src/registration/ecal_registration_provider.h
```diff
@@ -54,64 +53,49 @@ namespace eCAL
...
+    bool ApplySample(const Registration::Sample& sample_, bool force_);
 
-    bool RegisterServer(const std::string& service_name_, const std::string& service_id_, const Registration::Sample& ecal_sample_, bool force_);
-    bool UnregisterServer(const std::string& service_name_, const std::string& service_id_, const Registration::Sample& ecal_sample_, bool force_);
-
-    bool RegisterClient(const std::string& client_name_, const std::string& client_id_, const Registration::Sample& ecal_sample_, bool force_);
-    bool UnregisterClient(const std::string& client_name_, const std::string& client_id_, const Registration::Sample& ecal_sample_, bool force_);
+    using ApplySampleCallbackT = std::function<void(const Registration::Sample&)>;
+    void SetCustomApplySampleCallback(const std::string& customer_, const ApplySampleCallbackT& callback_);
+    void RemCustomApplySampleCallback(const std::string& customer_);
```
KerstinKeller, line 63: I think API-wise, something like this would be cleaner:
```
struct CallbackID
{
private: 
CallbackID();
int ID;
} 
// You can now also omit the customer string, but if you potentially want to know, keep it
CallbackID SetCustomApplySampleCallback(const std::string& customer_, const ApplySampleCallbackT& callback_);
void RemCustomApplySampleCallback(const CallbackID & customer_);
```
Pros: 
- You don't need a map with a string (which are more expensive for lookup than ints). 
- Also you cannot try to remove something that you have never added - harder to missuse

-> No need for this PR but in general...

KerstinKeller (approved):
Much cleaner :+1:

PR #1481, opened 2024-03-21
[core] register, unregister logic adapted to prevent data races

rex-schilasky:
### Description
In the destruction phase of a publisher, subscriber, client, server it could happen that parallel to the unregistration of the entity a cyclic registration is triggered by the registration provider refresh thread. This should be avoided by adaption the creation flag logic and making this state atomic.


FlorianReimold:
Those cherry pick labels only work for pushes to master

KerstinKeller (approved):
:+1:

PR #1493, opened 2024-03-27
[core] [gui] Basic monitoring support for client entities

hannemn:
This implementation enriches the registration/monitoring layer with meta information such as call methods and its request/response type/descriptor strings, call counts, etc.  from client enties in the eCAL network.

Parts of the meta information can be passed on construction time of a client instance with following new public API method:
`CServiceClient(const std::string& service_name_, const ServiceMethodInformationMapT& method_information_map_)`
In this case` ServiceMethodInformationMapT` contains unique pairs of _method name as key_ and  _request/response type/descriptor string as value_ which are called by the client for the respective service. 

All those new monitoring information coming with this PR are displayed in the GUI version of the eCAL monitor as well in order to get even a more deeper insight into client-service communication during runtime.

image


ecal/core/include/ecal/msg/protobuf/client.h
```diff
@@ -189,6 +198,43 @@ namespace eCAL
...
+
+          // get message type names
+          const std::string input_type_name = method_descriptor->input_type()->name();
+          const std::string output_type_name = method_descriptor->output_type()->name();
+
+          // get message type descriptors
+          std::string input_type_desc;
+          std::string output_type_desc;
+
+          dyn_decoder.GetServiceMessageDescFromType(service_descriptor, input_type_name, input_type_desc, error_s);
```
KerstinKeller, line 225: rename request_type_name / response_type_name

ecal/core/include/ecal/msg/protobuf/client.h
```diff
@@ -189,6 +198,43 @@ namespace eCAL
...
+          dyn_decoder.GetServiceMessageDescFromType(service_descriptor, output_type_name, output_type_desc, error_s);
+
+
+          method_information_map[method_name] = SServiceMethodInformation({
+            {input_type_name, "protobuf", input_type_desc} ,
+            {output_type_name, "protobuf", output_type_desc}
```
KerstinKeller, line 231: If we want to be compatible with pubsub, this `"protobuf"` should be renamed to `"proto"`.

ecal/core/src/service/ecal_service_client_impl.cpp
```diff
@@ -684,6 +718,25 @@ namespace eCAL
...
+        method.mname = method_name;
+        method.req_type = method_information.request_type.name;
+        method.req_desc = method_information.request_type.descriptor;
+        method.resp_type = method_information.response_type.name;
+        method.resp_desc = method_information.response_type.descriptor;
+        method.call_count = m_method_call_count_map.at(method_name);
+        service_client.methods.push_back(method);
+      }
+    }
+
```
KerstinKeller, line 739: Unregister does not need to unregister methods.

rex-schilasky (approved):
LGTM 👍

PR #1505, opened 2024-04-04
[config] Implementation of new configuration management

Peguen:
### Description
Introduced config management via structs which opens the possibilty for users to configure eCAL in code. In addition, evaluation of the inputs will be easily possible.

If assigned configuration values can be evaluated and are incorrect, the program will exit with an error.

Internally also the new config structs will be used.


KerstinKeller (commented):
First part of review.

ecal/core/src/ecal_process.cpp
```diff
@@ -164,35 +166,36 @@ namespace eCAL
...
+      sstream << "Network rcvbuf           : " << GetBufferStr(Config::GetCurrentConfig()->transport_layer_options.mc_options.recbuf.get()) << '\n';
+      sstream << "Multicast cfg version    : v" << static_cast<uint32_t>(Config::GetCurrentConfig()->transport_layer_options.mc_options.config_version) << '\n';
+      sstream << "Multicast group          : " << Config::GetCurrentConfig()->transport_layer_options.mc_options.group.get() << '\n';
+      sstream << "Multicast mask           : " << Config::GetCurrentConfig()->transport_layer_options.mc_options.mask.get() << '\n';
+      const int port = Config::GetCurrentConfig()->transport_layer_options.mc_options.port.get();
       sstream << "Multicast ports          : " << port << " - " << port + 10 << '\n';
-      sstream << "Multicast join all IFs   : " << (Config::IsUdpMulticastJoinAllIfEnabled() ? "on" : "off") << '\n';
-      sstream << '\n';
+      sstream << "Multicast join all IFs   : " << (Config::GetCurrentConfig()->transport_layer_options.mc_options.join_all_interfaces ? "on" : "off") << '\n';
+
```
KerstinKeller, line 194: (also not in this PR but later):
Maybe, instead of printing everything here, for each Config(sub)object there be a print / tostring / operator<< method (preferrably the last one, but as non-member function), that takes care of knowing how to print itself.

ecal/core/include/ecal/types/ecal_monitoring_options.h
```diff
@@ -0,0 +1,70 @@
...
+    struct SHMMonitoringOptions
+    {
+      std::string shm_monitoring_domain;
+      size_t      shm_monitoring_queue_size;  
+    };
+
+    struct MonitoringOptions
+    {
+      eCAL_MonitoringMode_Filter monitoring_mode;
+      LimitSize<1000, 1000>      monitoring_timeout;
```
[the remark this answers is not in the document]
Peguen: No, the 2nd 1000 is the step. Its just allowed to be 1000 * x

ecal/core/src/io/udp/ecal_udp_configurations.cpp
```diff
@@ -79,20 +80,20 @@ namespace eCAL
...
+      const bool local_only = !Config::GetCurrentConfig()->transport_layer_options.network_enabled;
       if (local_only)
       {
         return GetLocalBroadcastAddress();
       }
 
       // both in v1 and v2, the multicast group is returned as the adress for the registration layer
-      return Config::GetUdpMulticastGroup();
+      return Config::GetCurrentConfig()->transport_layer_options.mc_options.group.get();
     }
```
KerstinKeller, line 91: To be honest, I don't like that you have to call `get()` here. Some options have to be treated differently form others.
Maybe this type could define a `operator std::string()` so you can implicitly convert to string?

ecal/core/src/readwrite/tcp/ecal_reader_tcp.cpp
```diff
@@ -76,7 +76,7 @@ namespace eCAL
     // add new session and activate callback if we add the first session
     if (new_session)
     {
-      m_subscriber->addSession(host_name_, port_, Config::GetTcpPubsubMaxReconnectionAttemps());
+      m_subscriber->addSession(host_name_, port_, static_cast<int>(Config::GetCurrentConfig()->transport_layer_options.tcp_options.max_reconnections));
```
KerstinKeller, line 79: why do you need a `static_cast`? what did `GetTcpPubsubMaxReconnectionAttemps` return?

ecal/tests/cpp/config_test/src/config_test.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+  // Test registration option assignment, default timeout is 60000U and default refresh is 1000U
+  EXPECT_EQ(registration_timeout, eCAL::Config::GetCurrentConfig()->registration_options.getTimeoutMS());
+  EXPECT_EQ(registration_refresh, eCAL::Config::GetCurrentConfig()->registration_options.getRefreshMS());
+
+  // Finalize eCAL API
+  EXPECT_EQ(0, eCAL::Finalize());
+}
+
+TEST(ConfigDeathTest, user_config_death_test)
```
KerstinKeller, line 106: I'd suggest a few more testcases on the `IPAddressV4` object itself: `0.0.0.0`, `000.000.000.000` (are they both valid), and some invalid corner cases (`256.0.0.0`, ...)

KerstinKeller (commented):
I am not 100% sure. Just an idea.
Maybe for the first PR, we keep all these:
```
    if ((Config::IsServiceProtocolV1Enabled()) && (server_tcp_port_v1 == 0)) return;
    if ((Config::GetCurrentConfig()->service_options.protocol_v1) && (server_tcp_port_v1 == 0)) return;
```
functions, and have them return the member variables of the config object. This way, we can more easily iterate on the structure of the config object without having to change code in too many places.
And if then we are sure, we do the replacement.
But I totally understand if you alread did all the work that you'd like to keep it like this.

ecal/core/src/readwrite/tcp/ecal_reader_tcp.cpp
Peguen: It returned an int. Now it changed to unsigned int. Two questions:

1. Would rather instead be size_t the better choice?
2. Change the addSession method to take a size_t/uint instead of int?

ecal/core/src/ecal_globals.h
```diff
@@ -64,12 +65,16 @@ namespace eCAL
...
 
+    void SetUserConfig(Config::eCALConfig& user_config_);
+    
     unsigned int GetComponents() const { return(components); };
 
     int Finalize(unsigned int components_);
 
     const std::unique_ptr<CConfig>&                                       config()                 { return config_instance; };
     const std::unique_ptr<CLog>&                                          log()                    { return log_instance; };
+    const std::unique_ptr<Config::eCALConfig>&                            ecal_config()            { return ecal_config_instance; };
```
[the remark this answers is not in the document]
Peguen: CConfig is the wrapper for the simpleini config reading. It will be probably removed when confirmed from the globals.

ecal/core/src/config/ecal_config_initializer.cpp
```diff
@@ -0,0 +1,268 @@
...
+    Configuration::Configuration(std::vector<std::string> args_)
+    {
+      args_.emplace(args_.begin(), eCAL::Process::GetProcessName());
+      std::vector<const char*> argv(args_.size());
+      std::transform(args_.begin(), args_.end(), argv.begin(), [](std::string& s) {return s.c_str();});
+      
+      Init(static_cast<int>(argv.size()), const_cast<char**>(argv.data()));
+    }
+
+    void Configuration::Init(int argc_ , char **argv_)
```
hannemn, line 236: Maybe it's safer to use Init(std::vector<std::string>) as main overload - Init(int, char**) calls Init(std::vector..) etc. - to prevent the usage of const_cast<>.

hannemn (approved):
Reviewed together with @Peguen. Further restructuring of configuration object is needed but can be handled in a subsequent PR.

PR #1506, opened 2024-04-04
[mma] add missing include.

KerstinKeller:
### Description
`#include<ctime>` was missing, although `std::time_t` is used in the header file.


hannemn (approved):
Awesome!

PR #1514, opened 2024-04-08
[core/test] util_test (ParallelGetTopics) was excluded due to incorrect CMake setting

rex-schilasky:
### Description
Minimal fix: ParallelGetTopics (core/util_test) gtest was excluded due to ECAL_CORE_PUBLISHER condition (that should not used in test cpp modules, because all ECAL_CORE_XY variables are not forwarded as compiler defines.)


Peguen (approved):
lgtm

hannemn (approved):

PR #1520, opened 2024-04-10
[core] active entity unregistration in descgate

rex-schilasky:
### Description
DescGate is currently collecting descriptions (type encodings, type names, type descriptors) for publisher, subscriber and services. 

This PR:
* extends it to collect descriptions for clients too
* introduces active removal of descriptions triggered by unregistration events (samples)

The existing removal is based on expiration, configured with a default timeout of 5s (monitoring timeout). The new active removal removes entries immediately.

### DIFF
--- a/ecal/core/src/ecal_descgate.h
+++ b/ecal/core/src/ecal_descgate.h
@@ -72,26 +72,20 @@ namespace eCAL
     bool GetDataTypeInformation(const std::string& topic_name_, SDataTypeInformation& topic_info_);
 
     void GetServices(std::map<std::tuple<std::string, std::string>, SServiceMethodInformation>& service_info_map_);
-    void GetServiceNames(std::vector<std::tuple<std::string, std::string>>& service_method_names_);
+    void GetServiceMethodNames(std::vector<std::tuple<std::string, std::string>>& service_method_names_);
     bool GetServiceTypeNames(const std::string& service_name_, const std::string& method_name_, std::string& req_type_name_, std::string& resp_type_name_);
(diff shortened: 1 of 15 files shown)


ecal/core/src/ecal_util.cpp
```diff
@@ -168,46 +298,277 @@ namespace eCAL
...
-      g_descgate()->GetTopics(topic_info_map_);
+      QualityTopicInfoMultiMap multi_map;
+      if (g_descgate() == nullptr) return multi_map;
+
+      // insert publisher into target multimap
+      for (const auto& topic : g_descgate()->GetPublishers())
+      {
+        multi_map.insert(std::pair<std::string, SQualityTopicInfo>(topic.first.topic_name, topic.second));
+      }
+
```
KerstinKeller, line 311: what is the difference of the type the DescGate returns, and that the Util function returns?
Because we transform the map twice now? once from expired_map -> regular map -> multimap.
Does it make sense that the descgate directly returns a multimap?

KerstinKeller (commented):
Sorry I have some more remarks. some might be from yesterday, and outdated, but others are from today.

ecal/core/src/ecal_descgate.h
```diff
@@ -26,114 +26,117 @@
...
+
+  struct SServiceIdKey
+  {
+    std::string     service_name;
+    std::string     method_name;
+    Util::ServiceId service_id;
+
+    bool operator<(const SServiceIdKey& other) const
+    {
+      return std::tie(service_name, service_id) < std::tie(other.service_name, other.service_id);
```
KerstinKeller, line 59: Did you "forget" the `method_name` in the comparison, or is this by intent?

KerstinKeller (approved):
Looks good, now.

PR #1527, opened 2024-04-15
[core] Fix possible race condition

KerstinKeller:
ecal_registration_provider.cpp: m_sample_list_sync should also protect m_sample_list_buffer.


KerstinKeller:
Cherry pick manually.

PR #1528, opened 2024-04-15
[build]  make eCAL::proto part of ecal subfolder (ecal_core project)

KerstinKeller:
- changed include folder from "ecal/protobuf/..." to "ecal/msg/protobuf/..." to be in line with the includes from core (e.g. subscriber.h / publisher.h)
- target naming: eCAL::core_protobuf to be discussed.


conanfile.py
```diff
@@ -6,51 +6,68 @@
 
 class eCALConan(ConanFile):
     name = "ecal"
-    settings = "os", "compiler", "arch", "build_type"
```
KerstinKeller, line 9: (!!!!!!)

rex-schilasky (approved):

PR #1536, opened 2024-04-16
[Submodules] ecaludp include

Peguen:
### Description
Implemented ecaludp as submodule.


thirdparty/ecaludp/build-ecaludp.cmake
```diff
@@ -0,0 +1,15 @@
...
+set(ECALUDP_ENABLE_NPCAP        ${ECAL_CORE_NPCAP_SUPPORT})
+set(ECALUDP_BUILD_SAMPLES       OFF)
+set(ECALUDP_BUILD_TESTS         OFF)
+set(ECALUDP_USE_BUILTIN_ASIO    OFF)
+set(ECALUDP_USE_BUILTIN_RECYCLE OFF)
+set(ECALUDP_USE_BUILTIN_UDPCAP  OFF)
+set(ECALUDP_LIBRARY_TYPE        STATIC)
+
+# Add ecaludp library from subdirectory
+add_subdirectory(${CMAKE_CURRENT_LIST_DIR}/ecaludp thirdparty/ecaludp EXCLUDE_FROM_ALL SYSTEM)
```
KerstinKeller, line 12: I think we could also directly include `${CMAKE_CURRENT_LIST_DIR}/ecaludp/ecaludp`, and then we don't need to set alld the options, but I guess it's not actually necesssary.

PR #1538, opened 2024-04-16
new - ecaludp based - udp send/receive handling

rex-schilasky:
### Description
Replace ecal's udp implementation by ecaludp.


ecal/core/src/io/udp/ecal_udp_sample_receiver.h
```diff
@@ -43,43 +40,17 @@ namespace eCAL
...
-        explicit CSampleDefragmentation(CSampleReceiver* sample_receiver_);
-        ~CSampleDefragmentation() override;
-
-        int OnMessageCompleted(std::vector<char>&& msg_buffer_) override;
-
-      protected:
-        CSampleReceiver* m_sample_receiver;
-      };
+    private:
+      std::shared_ptr<CSampleReceiverAsio>  m_receiver_asio;
```
KerstinKeller, line 49: I would recommend that these are `std::unique_ptr`.

ecal/core/src/io/udp/ecal_udp_receiver_attr.h
```diff
@@ -18,21 +18,28 @@
...
+      std::string address;
+      int         port      = 0;
+      bool        broadcast = false;
+      bool        loopback  = true;
+      int         rcvbuf    = 1024 * 1024;
+    };
 
-    using TransmitCallbackT = std::function<size_t(const void*, const size_t)>;
-    size_t SendFragmentedMessage(char* buf_, size_t buf_len_, const TransmitCallbackT& transmit_cb_);
+    using HasSampleCallbackT   = std::function<bool(const std::string& sample_name_)>;
```
KerstinKeller, line 42: I am a bit curious about the naming of `HasSampleCallbackT`. When is this called and what should it return?
If this is true, then `ApplySampleCallback` is called?

ecal/core/src/io/udp/ecal_udp_sample_receiver_asio.cpp
```diff
@@ -0,0 +1,253 @@
...
+      AddMultiCastGroup(attr_.address.c_str());
+
+      // run the io context
+      m_io_thread = std::thread([this] { m_io_context->run(); });
+
+      // start receiving
+      Receive();
+    }
+
+    CSampleReceiverAsio::~CSampleReceiverAsio()
```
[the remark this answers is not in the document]
KerstinKeller: If the action bot is right here, you might want to surround this with a `try / catch` block. @FlorianReimold do you know under which circumstances this may throw?

ecal/core/src/io/udp/ecal_udp_sample_receiver_asio.cpp
```diff
@@ -0,0 +1,253 @@
...
+      if (m_io_thread.joinable())
+        m_io_thread.join();
+    }
+
+    bool CSampleReceiverAsio::AddMultiCastGroup(const char* ipaddr_)
+    {
+      if (!m_broadcast)
+      {
+#ifdef __linux__
+        // TODO: NPCAP - How to implement this ? native_handle() not provided
```
KerstinKeller, line 75: Is this still relevant? Otherwise remove?

ecal/core/src/io/udp/ecal_udp_sample_receiver_asio.cpp
```diff
@@ -0,0 +1,253 @@
...
+      }
+      return(true);
+    }
+
+    bool CSampleReceiverAsio::RemMultiCastGroup(const char* ipaddr_)
+    {
+      if (!m_broadcast)
+      {
+        // Leave multicast group
+#ifdef __linux__
```
KerstinKeller, line 103: why `__linux__` and not `ECAL_OS_LINUX`?

ecal/core/src/io/udp/ecal_udp_sample_receiver_npcap.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+              // apply the sample payload
+              m_apply_sample_callback(payload_buffer, payload_buffer_size);
+            }
+          }
+
+          // recursively call Receive() to continue listening for data
+          this->Receive();
+        });
+    }
```
KerstinKeller, line 172: There is a lot of duplicated code between this class and the asio class. It might make sense to actually compare them, and see what code may be reused between both classes.

ecal/core/src/io/udp/ecal_udp_sample_sender.cpp
```diff
@@ -22,51 +22,113 @@
...
+    CSampleSender::CSampleSender(const SSenderAttr& attr_) :
+      m_destination_endpoint(asio::ip::make_address(attr_.address), static_cast<unsigned short>(attr_.port))
+    {
+      m_io_context = std::make_shared<asio::io_context>();
+
+      // create the socket and set all socket options
+      InitializeSocket(attr_);
+    }
+
+    CSampleSender::~CSampleSender()
```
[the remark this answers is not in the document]
KerstinKeller: Same as above, can this throw?

ecal/core/src/io/udp/ecal_udp_sample_sender.cpp
```diff
@@ -22,51 +22,113 @@
...
+      // close socket
+      m_socket->close();
     }
 
-    size_t CSampleSender::Send(const std::string& sample_name_, const std::vector<char>& serialized_sample_)
+    void CSampleSender::InitializeSocket(const SSenderAttr& attr_)
     {
-      if (!m_udp_sender) return(0);
+      // create socket
+      m_socket = std::make_shared<ecaludp::Socket>(*m_io_context, std::array<char, 4>{'E', 'C', 'A', 'L'});
```
KerstinKeller, line 52: This is the thirds place, where I see exactly this line of code:
```
m_socket = std::make_shared<ecaludp::Socket>(*m_io_context, std::array<char, 4>{'E', 'C', 'A', 'L'});
```
Make it can be put inside a free function?

ecal/core/src/io/udp/ecal_udp_sample_receiver_asio.cpp
KerstinKeller: also didn't you change the namespace `IO::UDP` -> `eCAL::UDP` ? 🤔

KerstinKeller (commented):
I think we also need a review from @FlorianReimold, because I cannot judge the asio content.

ecal/core/src/io/udp/ecal_udp_sample_receiver_asio.cpp
rex-schilasky: I kept __linux__ here. Maybe its used because the OS_XY macros should not be used in low level IO layers .. can check this later. Namespace here is still IO::UDP

ecal/core/src/io/udp/ecal_udp_sample_receiver_asio.cpp
rex-schilasky: Still relevant.

ecal/core/src/io/udp/ecal_udp_receiver_attr.h
rex-schilasky: It's a callback called after reading the sample_name and before deserializing the payload. So the later "consumer" of the payload can state its interest on the payload.

ecal/core/src/io/udp/ecal_udp_sample_receiver.h
rex-schilasky: Done

ecal/core/src/io/udp/ecal_udp_sample_sender.cpp
```diff
@@ -22,51 +22,113 @@
...
+      // ------------------------------------------------
+      const unsigned short s1 = static_cast<unsigned short>(sample_name_.size()) + 1 /*'\0'*/;
+      const size_t         s2 = serialized_sample_.size();
+      const asio::const_buffer sample_name_size_asio_buffer(&s1, 2);
+      const asio::const_buffer sample_name_asio_buffer(sample_name_.c_str(), s1); // we need to use c_str() here to guarantee  trailling \'0'
+      const asio::const_buffer serialized_sample_asio_buffer(serialized_sample_.data(), s2);
+
+      const asio::socket_base::message_flags flags(0);
+      asio::error_code ec;
+      size_t sent = m_socket->send_to({ sample_name_size_asio_buffer, sample_name_asio_buffer, serialized_sample_asio_buffer }, m_destination_endpoint, flags, ec);
```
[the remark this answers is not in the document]
rex-schilasky: Done

ecal/core/src/io/udp/ecal_udp_sample_receiver_asio.cpp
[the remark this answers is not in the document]
rex-schilasky: Solved by using the asio::error_code variants of socket close/cancel.

ecal/core/src/io/udp/ecal_udp_sample_sender.cpp
rex-schilasky: I would leave it that way and not introduce an extra function for the 2 receivers (asio and npcap) and the one sender.

ecal/core/src/io/udp/ecal_udp_sample_receiver.h
```diff
@@ -23,63 +23,29 @@
...
-#include <string>
-#include <unordered_map>
-#include <vector>
 
 namespace eCAL
 {
   namespace UDP
   {
-    class CSampleReceiver
+    class CSampleReceiver : public CSampleReceiverBase
```
KerstinKeller, line 39: This class does not need to inherit from the Base class! This is a wrapper class for the base class. I theory, you might not even need this class at all.

ecal/core/src/io/udp/ecal_udp_sample_receiver.h
```diff
@@ -23,63 +23,29 @@
...
 
-#include "io/udp/sendreceive/udp_receiver.h"
-#include "io/udp/fragmentation/rcv_fragments.h"
-#include "util/ecal_thread.h"
+#include "io/udp/ecal_udp_sample_receiver_base.h"
+
+#include "ecal_udp_sample_receiver_asio.h"
+#ifdef ECAL_CORE_NPCAP_SUPPORT
+#include "ecal_udp_sample_receiver_npcap.h"
+#endif
```
KerstinKeller, line 32: the includes at this point are not necessary. in the idea of `Pimpl` you might even remove the `ecal_udp_sample_receiver_base.h` include and here just say `class CSampleReceiverBase;`

KerstinKeller (commented):
I have two remarks, which make sense to fix atm.

KerstinKeller (approved):
Looks good! :+1:

PR #1541, opened 2024-04-17
[doc] downgrade Sphinx to < 7.3

KerstinKeller:


rex-schilasky (approved):

KerstinKeller:
We should be able to remove the restriction again.

PR #1557, opened 2024-04-23
Python 3.9 downgrade for macOS GH runner removed

rex-schilasky:
### Description
As the title says .. Python 3.9 ARM64 no more available for macOS GH runner. See https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -25,7 +25,7 @@ jobs:
         target: 'desktop'
 
     - name: Install Dependencies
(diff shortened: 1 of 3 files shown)


FlorianReimold:
You are probably referring to amd64, not arm64, right? to my knowledge, arm was never available.

rex-schilasky:
> You are probably referring to amd64, not arm64, right? to my knowledge, arm was never available.

No, in the error log it is missing an ARM64 package .. even we do not use that one.

image

PR #1561, opened 2024-04-23
[Core] publisher config API

rex-schilasky:
### Description
Draft implementation of a publisher configuration API in preparation of the new global eCAL configuration API.


ecal/core/src/readwrite/ecal_writer.cpp
```diff
@@ -68,60 +67,35 @@ namespace std
...
     m_topic_size(0),
-    m_buffering_shm(PUB_MEMFILE_BUF_COUNT),
-    m_zero_copy(PUB_MEMFILE_ZERO_COPY),
-    m_acknowledge_timeout_ms(PUB_MEMFILE_ACK_TO),
+    m_config(config_),
     m_connected(false),
     m_id(0),
     m_clock(0),
-    m_frequency_calculator(3.0f),
+    m_frequency_calculator(0.0f),
```
KerstinKeller, line 82: is this change on purpose?

ecal/core/src/pubsub/ecal_publisher.cpp
```diff
@@ -35,6 +35,27 @@
...
+    shm.memfile_reserve_percent = eCAL::Config::GetMemfileOverprovisioningPercentage();
+    shm.memfile_buffer_count    = eCAL::Config::GetMemfileBufferCount();
+
+    // udp config
+    udp.send_mode               = eCAL::Config::GetPublisherUdpMulticastMode();
+    udp.sndbuf_size_bytes       = eCAL::Config::GetUdpMulticastSndBufSizeBytes();
+
+    // tcp config
+    tcp.send_mode               = eCAL::Config::GetPublisherTcpMode();
+  }
```
KerstinKeller, line 57: This introduces a tight coupling between the publisher config, and `eCAL::Config`.
I'd prefer the constructor to have hardcoded values, but then an `eCAL::Config::GetPublisherConfig` which returns then config object, filled with the default values from `eCAL::Config`.

ecal/core/src/readwrite/ecal_writer.h
```diff
@@ -162,10 +148,7 @@ namespace eCAL
     SDataTypeInformation                   m_topic_info;
     std::map<std::string, std::string>     m_attr;
     size_t                                 m_topic_size;
-
-    size_t                                 m_buffering_shm;
-    bool                                   m_zero_copy;
-    long long                              m_acknowledge_timeout_ms;
+    CPublisher::Config                     m_config;
```
KerstinKeller, line 151: I am a bit unsure about this here. Suddenly, ecal_writer needs to know about `CPublisher` (you had to add an include), and then your low level implementations depends on the high level implementation. 
If Publisher knows Writer, and Writer knows publisher, this leads to circular includes, which is a bad thing.

So there are two alternative options:
1) he Configuration needs to be extracted from the publisher, into it's own header file. 
2) Actually the cleanest would be a `WriterConfiguration`. and then a function where the `PublisherConfiguration` is mapped to the `WriterConfiguration`, even if in our case they are exactly the same.

ecal/core/src/readwrite/shm/ecal_writer_shm.cpp
```diff
@@ -74,6 +68,13 @@ namespace eCAL
       return false;
     }
 
+    // prepare memfile attributes
+    SSyncMemoryFileAttr memory_file_attr = {};
+    memory_file_attr.min_size        = m_config.memfile_min_size_bytes;
+    memory_file_attr.reserve         = m_config.memfile_reserve_percent;
+    memory_file_attr.timeout_open_ms = PUB_MEMFILE_OPEN_TO;
+    memory_file_attr.timeout_ack_ms  = m_config.acknowledge_timeout_ms;
+
```
KerstinKeller, line 77: Maybe extract this into its own function?

ecal/core/src/readwrite/shm/ecal_writer_shm.cpp
```diff
@@ -112,12 +113,12 @@ namespace eCAL
...
 
     // adapt number of used memory files if needed
-    if (attr_.buffering != m_buffer_count)
+    if (attr_.buffering != m_config.memfile_buffer_count)
     {
       SetBufferCount(attr_.buffering);
 
       // store new buffer count and flag change
-      m_buffer_count = attr_.buffering;
+      m_config.memfile_buffer_count = attr_.buffering;
```
KerstinKeller, line 121: it feels a bit wrong if you change the config internally? the config is set from outside, maybe here you need a  variable "actual_buffer_count" or something?

ecal/core/src/readwrite/ecal_writer.cpp
KerstinKeller: @rex-schilasky

ecal/core/src/readwrite/udp/ecal_writer_udp.cpp
```diff
@@ -44,7 +44,7 @@ namespace eCAL
     attr.port      = UDP::GetPayloadPort();
```
KerstinKeller, line 44: Other question: I think we indirectly access the config here? does it make sense to move also this information to the `Publisher::UDP::Configuration`? Although we would never want the user to change those.... (just out of curiosity, not relevant for this PR).

KerstinKeller (approved):
Good job 😄 we're moving forward.

ecal/core/src/readwrite/ecal_reader.cpp
```diff
@@ -59,65 +59,35 @@ namespace eCAL
...
                  m_frequency_calculator(3.0f),
-                 m_message_drops(0),
-                 m_loc_published(false),
-                 m_ext_published(false),
-                 m_use_ttype(true),
-                 m_use_tdesc(true),
-                 m_use_udp_mc_confirmed(false),
-                 m_use_shm_confirmed(false),
-                 m_use_tcp_confirmed(false),
+                 m_share_ttype(Config::IsTopicTypeSharingEnabled()),
```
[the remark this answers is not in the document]
rex-schilasky: yes, will be part of the new CSubscriber::Configuration

PR #1579, opened 2024-05-02
[Core / Service]: Fixed crash that occured when communicating with eCAL 5.12 services

### LINKED ISSUE #1578: eCAL Rec crash with eCAL 5.13.0 + 5.12.4
### Problem Description

eCAL Rec GUI 5.13.0 terminates when you start a ecal_sample_rec_server_service_gui

### How to reproduce

- install eCAL 5.13.0 and 5.12.4 in parallel
- Start eCAL Rec GUI from 5.13.0
- Start ecal_sample_rec_server_service_gui from 5.12.4 (copy to bin dir for finding Qt DLLs)
- eCAL Rec GUI disappears

### How did you get eCAL?

Download from Release Page

### Environment

- eCAL 5.13.0 + 5.12.4
- Windows 11

### eCAL System Information

_No response_

FlorianReimold:
### Description
The crash occured when a v0 client connected to a service server.

### Related issues
- Fixes #1578

### DIFF
--- a/ecal/service/ecal_service/src/server_session_impl_v0.cpp
+++ b/ecal/service/ecal_service/src/server_session_impl_v0.cpp
@@ -75,6 +75,18 @@ namespace eCAL
     ///////////////////////////////////////////////
     void ServerSessionV0::start()
     {
+      // Call the handle_start with the io_service
+      // It is important to async call handle_start(), as it will call a
+      // user-defined callback. As we have no influence what that callback will
+      // be, we must call it from another thread to make sure to not double-lock
+      // mutexes from the server_impl, if the callback should itself call a
+      // server_impl api function.
+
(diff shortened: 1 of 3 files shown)


KerstinKeller (approved):
I don't quite understand why start() already calls user defined callbacks, and how that solves the problem. Is `start()` called by the user themselves? Maybe a flow diagram would help me understand. otherwise looks good :+1:

FlorianReimold:
ServerSessionV0::Start() is called by the factory create. As the buggy v0 protocol completely lacks any connection establishment, it immediately assumes that it is connected and therefore calls the  EventCallback with a connected event. Before, it did that in the same thread. Now it does that in the io_context thread.

v1 never had that issue, as the EventCallback is called after the connection has been established. Thus, the event callback has always been called from the io_context thread.

PR #1581, opened 2024-05-03
[cmake] move find_package(Protobuf) to bottom

KerstinKeller:
... as abseil (required in newer protobuf versions) somehow messes with paths, and then the following includes fail.


FlorianReimold:
Warning: This PR goes to 5.13. the 5.13 branch doesn't have a cherry-pick action at the moment. The PR is labled as cherry-pick to 5.13 though, which makes me assume that the base branch is wrong.

KerstinKeller:
you are right. das nervt ein bisschen, wenn man seine änderungen direkt aus der weboberfläche macht.
I will make another PR based on master...

PR #1592, opened 2024-05-15
[Core, Services, Protobuf] Fixed compatibility with Ubuntu 18.04 (gcc7 / protobuf 3.0)

FlorianReimold:
### Description
- fixed lambda caputres (not sure why I had to do that in those places but not in others)
- Removed reserved fields from protobuf enums


rex-schilasky (approved):
LGTM

PR #1604, opened 2024-05-17
[core] subscriber configuration

rex-schilasky:
### Description
CSubscriber configuration introduced.

### DIFF
--- a/ecal/core/src/pubsub/ecal_subscriber.cpp
+++ b/ecal/core/src/pubsub/ecal_subscriber.cpp
@@ -1,6 +1,6 @@
 /* ========================= eCAL LICENSE =================================
  *
- * Copyright (C) 2016 - 2019 Continental Corporation
+ * Copyright (C) 2016 - 2024 Continental Corporation
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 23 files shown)


ecal/core/src/pubsub/ecal_subscriber.cpp
```diff
@@ -96,6 +91,11 @@ namespace eCAL
     return(m_created);
   }
 
+  bool CSubscriber::Create(const std::string& topic_name_)
```
KerstinKeller, line 94: @rex-schilasky why did you add this function?

ecal/core/src/pubsub/ecal_subscriber_config.cpp
```diff
@@ -0,0 +1,43 @@
...
+    Configuration::Configuration()
+    {
+      // shm config
+      shm.enable = eCAL::Config::IsShmRecEnabled();
+
+      // udp config
+      udp.enable = eCAL::Config::IsUdpMulticastRecEnabled();
+
+      // tcp config
+      tcp.enable = eCAL::Config::IsTcpRecEnabled();
```
KerstinKeller, line 40: I think this is a bit strange, because you give default values in `ecal_subscriber_config.h` but then in the constructor, assign the values from `eCAL::Config`. 
We need to align with @Peguen but I agree maybe it makes sense that the eCAL::Config has a way to fill them with values from the Ini file...

KerstinKeller (approved):
I approve, but we need to keep thinking about the two comments which I made.

PR #1608, opened 2024-05-22
send frequency calculation fixed in case of unconnected publisher

### LINKED ISSUE #1603: Frequency Calculation broken if non Subscriber present
### Problem Description

The frequency calculation in eCAL is broken, if no subscriber is present.

image

This is confirmed to be working in eCAL 5.12, but it's broken in eCAL 5.13 and master.

### How to reproduce

It can be seen in eCAL Mon GUI, start `person_snd` but no subscribers and see that the frequency is displayed to be 0.

### How did you get eCAL?

### Environment

- eCAL: master and 5.13

### eCAL System Information

_No response_

rex-schilasky:
### Related issues
Fixes #1603


KerstinKeller (approved):
👍

PR #1615, opened 2024-05-24
[core] new-pub-sub-matching

rex-schilasky:
### Description
- layer states (enabling) exchanged between pub/sub
- layer activation based on layer priority lists for local and remote connections


ecal/tests/cpp/serialization_test/src/registration_generate.cpp
```diff
@@ -95,6 +95,7 @@ namespace eCAL
       TLayer layer;
       layer.type      = static_cast<eTLayerType>(rand() % (tl_all + 1));
       layer.version   = rand() % 100;
+      layer.enabled   = rand() % 2 == 1;
       layer.confirmed = rand() % 2 == 1;
       return layer;
```
KerstinKeller, line 100: 😨 Tests should be reproducible!
if you are generating numbers for a tests, you should do so with a seed, so they are always generating the same.
And in this case, combinations would maked more sense instead of just random numbers.

ecal/core_pb/src/ecal/core/pb/layer.proto
```diff
@@ -62,10 +63,12 @@ enum eTLayerType                                // transport layer
...
 {
-  reserved 4;
+  // Reserved fields in enums are not supported in protobuf 3.0
+  // reserved 4;
 
   eTLayerType      type               =   1;    // transport layer type
   int32            version            =   2;    // transport layer version
-  bool             confirmed          =   3;    // transport layer used ?
+  bool             enabled            =   6;    // transport layer enabled ?
+  bool             confirmed          =   3;    // transport layer usage confirmed ?
```
KerstinKeller, line 72: Can you explain the logic again?

KerstinKeller (commented):
Just a remark: I think we really need to discuss about starting layers even though no subscribers are there. What is the current behavior? If noone is subscribed, nothing happens.
If someone is subscribed on udp, is the publisher also always sending on shm, even though noone is listening on SHM?

What are the latencies, when a Publisher is present, a subscriber wants to subscriber, until the subscriber receives the first message, assuming e.g. the publisher sends at 10 Hz frequency. How is it in eCAL5?

rex-schilasky:
> Just a remark: I think we really need to discuss about starting layers even though no subscribers are there. What is the current behavior? If noone is subscribed, nothing happens. If someone is subscribed on udp, is the publisher also always sending on shm, even though noone is listening on SHM?
> 
> What are the latencies, when a Publisher is present, a subscriber wants to subscriber, until the subscriber receives the first message, assuming e.g. the publisher sends at 10 Hz frequency. How is it in eCAL5?

The old behavior is that write layers are enabled (instantiated) at the construction phase of the `CDataWriter` if their mode is configured as `off` or `auto`. This leads to the fact that a memory file is created, a socket is opened .. even there is no local/external subscription.

The new behavior is to not instantiate any of theses layers if their is no matching subscription for this kind of transport. The timing is still insides 2 registration loops as guaranteed in eCAL5.

ecal/core_pb/src/ecal/core/pb/layer.proto
rex-schilasky: `enabled` == enabled by the (user) publisher/subscriber logic
`confirmed` == data were successful send over the layer finally (used by the monitoring mainly)

rex-schilasky:
replaced by new PR new-pub-sub-matching-2

PR #1631, opened 2024-06-13
[core] expmap performance improvement (#1626)

rex-schilasky:
* performance improvement of update_timestamp for large maps (tested with > 25 k entries)
* monitoring_get_topics sample slightly improved to allow time updates of internal (large) expmaps in case of subsequent readings


hannemn (approved):
lgtm as discussed!

PR #1632, opened 2024-06-19
[core] add/improved IsPublished/IsSubscribed logic (5.13.x)

rex-schilasky:
### Description
* additional check for CSubscriber::GetPublisherCount() using the connection state of the matching subscriber to ensure that GetPublisherCount is increased only if the matching publisher is able to send data
* test added to test new behavior

### DIFF
--- a/ecal/core/src/pubsub/ecal_subgate.cpp
+++ b/ecal/core/src/pubsub/ecal_subgate.cpp
@@ -1,6 +1,6 @@
 /* ========================= eCAL LICENSE =================================
  *
- * Copyright (C) 2016 - 2019 Continental Corporation
+ * Copyright (C) 2016 - 2024 Continental Corporation
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


ecal/core/src/pubsub/ecal_subgate.cpp
```diff
@@ -295,8 +295,12 @@ namespace eCAL
 
         iter->second->ApplyLocLayerParameter(process_id, topic_id, tlayer.type(), writer_par);
       }
-      // inform for local publisher connection
-      iter->second->ApplyLocPublication(process_id, topic_id, topic_info);
+      // we only inform the subscriber when the publisher has already recognized at least on local subscriber
```
KerstinKeller, line 298: at least on -> at least one

ecal/core/src/pubsub/ecal_subgate.cpp
```diff
@@ -345,8 +349,12 @@ namespace eCAL
         const std::string writer_par = tlayer.par_layer().SerializeAsString();
         iter->second->ApplyExtLayerParameter(host_name, tlayer.type(), writer_par);
       }
-      // inform for external publisher connection
-      iter->second->ApplyExtPublication(host_name, process_id, topic_id, topic_info);
+      // we only inform the subscriber when the publisher has already recognized at least on external subscriber
```
KerstinKeller, line 352: At least on -> at least one

KerstinKeller (commented):
Let's remove IsPublished() for 5.13 and add it only for 5.14.

ecal/core/src/pubsub/ecal_subgate.cpp
rex-schilasky: fixed

ecal/core/src/pubsub/ecal_subgate.cpp
rex-schilasky: fixed

PR #1634, opened 2024-06-20
[core] add/improved IsPublished/IsSubscribed logic (5.12.x)

rex-schilasky:
### Description
* additional check for CSubscriber::GetPublisherCount() using the connection state of the matching subscriber to ensure that GetPublisherCount is increased only if the matching publisher is able to send data
* test added to test new behavior


KerstinKeller (approved):
:+1:

PR #1635, opened 2024-06-20
[CMake] Remove references to ${CMAKE_PROJECT_DIR} ...

KerstinKeller:
..  and upward references to source files.

Usage of ${CMAKE_PROJECT_DIR} causes the eCAL build to break when eCAL is included via `add_subdirectory`.


rex-schilasky (approved):
lgtm

PR #1636, opened 2024-06-20
[core] Expmap takes additional template parameter for clock type.

KerstinKeller:
This improves testability and allows to expire based on different time base, if required.


rex-schilasky (approved):
much easier to test !

PR #1646, opened 2024-07-04
[core] asio server per default initialized with ipv6 

KerstinKeller:
(which can fallback to ipv4 usage).

### Description
On systems which use IPv6 per default, there is currently a significant "waiting time" during the startup period of tasks which are using eCAL Services.
This is due to tthe circumstance, that the asio server is currently initialized with an IPv4 address and thus doesn't accept IPv6 requests.
This PR fixes that.

### Related issues
##1644


rex-schilasky (approved):
small change with a big impact

PR #1647, opened 2024-07-05
Feature/registration provider refactoring

KerstinKeller:
### Related issues

-


ecal/core/src/registration/ecal_registration_provider.cpp
```diff
@@ -150,12 +125,14 @@ namespace eCAL
     // if registration is forced
     if (force_)
     {
+      SendSampleList();
+
       // send single registration sample over udp
-      SendSample2UDP(sample_);
+      //SendSample2UDP(sample_);
```
KerstinKeller, line 132: The behavior here is actually different than before, let's discuss if it's a problem

rex-schilasky (approved):
Intermediate version, will be refactored further.

PR #1655, opened 2024-07-12
[config] cleanup

Peguen:
### Description
Cleanup of unused member in transportlayer configuration as well as removing API call functions that relate to that.

Changed executor reader/writer amount configuration to publisher and subscriber.


rex-schilasky (approved):
LGTM

PR #1659, opened 2024-07-15
[gh-actions] kill-xprotect-scan-macos

rex-schilasky:
### Description
Kill XProtect to avoid malware scan of dmg image.

### Related issues
See https://github.com/actions/runner-images/issues/7522


hannemn (approved):
But only as a temporary workaround until we find another, cleaner way ;-)

PR #1662, opened 2024-07-16
Feature/registration receiver logic

KerstinKeller:
### Related issues

-


ecal/core/src/registration/shm/ecal_registration_receiver_shm.cpp
```diff
@@ -27,47 +27,48 @@
 
 #include "ecal_globals.h"
```
KerstinKeller, line 28: check if include is necessary / instead include config?

rex-schilasky (approved):
LGTM

PR #1665, opened 2024-07-18
[core] new-pub-sub-matching-compatibility-5

rex-schilasky:
### Description
Transport layer version is increased from 1 to 2 for eCAL6. This allows to recognize registrations of old transport layer configurations not containing the new dynamic enabling feature. In this case the transport layer is interpreted as enabled and the communication eCAL5 <> eCAL6 is kept compatible.


Peguen (approved):
lgtm

PR #1666, opened 2024-07-18
[sample] benchmark-sample-massive_pub_sub-added

rex-schilasky:
### Description
Sample to compare pub/sub creation performance over different eCAL versions.

### DIFF
--- a/samples/cpp/benchmarks/massive_pub_sub/src/massive_pub_sub.cpp
+++ b/samples/cpp/benchmarks/massive_pub_sub/src/massive_pub_sub.cpp
@@ -1,0 +1,128 @@
+/* ========================= eCAL LICENSE =================================
+ *
+ * Copyright (C) 2016 - 2024 Continental Corporation
+ *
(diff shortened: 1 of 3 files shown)


samples/cpp/benchmarks/massive_pub_sub/src/massive_pub_sub.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+#include <ecal/ecal.h>
+
+#include <iostream>
+#include <chrono>
+#include <sstream>
+#include <thread>
+#include <vector>
+
+const int subscriber_number                    (5000);
```
Peguen, line 28: could use constexpr instead of const

Peguen (approved):
lgtm

samples/cpp/benchmarks/massive_pub_sub/src/massive_pub_sub.cpp
rex-schilasky: correct, next time / iteration I will fix this

PR #1667, opened 2024-07-19
[core] registration-loopback-logic-fixed-for-service-registrations

rex-schilasky:
### Description
* loopback logic (EnabledLoopback) was ignored for service registration
* some refactoring of the logic whether to process a sample registration or not
* user registration callback performance fix
  * samples will be reserialized only if the corresponding callback is set

### DIFF
--- a/ecal/core/src/registration/ecal_registration_receiver.cpp
+++ b/ecal/core/src/registration/ecal_registration_receiver.cpp
@@ -139,17 +139,9 @@ namespace eCAL
       }
     }
 
-    std::string reg_sample;
-    if (m_callback_pub
-      || m_callback_sub
-      || m_callback_service
-      || m_callback_client
-      || m_callback_process
-      )
-    {
-      SerializeToBuffer(sample_, reg_sample);
-    }
-
+    // forward registration to defined gates
+    // and store user registration callback
+    RegistrationCallbackT reg_callback(nullptr);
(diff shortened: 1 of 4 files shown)


KerstinKeller (commented):
I think in general this is a good cleanup.
However I want to mention that we still have the following "problems":

- Callbacks are always applied, not only if `ShouldProcessRegistration()==true`. Do we want that behavior? Should the user be informed of a publisher, that he would never receive data from, even if he created a publisher for it?
- Same for monitoring / descgate map. Feels wrong to have information available, if they are not applicable for the process. OR we should additional save the information e.g. it appears in the monitoring but we can see from the settings / another flag that this process won't be able to receive data from this entity because of the current configuration.

rex-schilasky:
> * Callbacks are always applied, not only if `ShouldProcessRegistration()==true`. Do we want that behavior? Should the user be informed of a publisher, that he would never receive data from, even if he created a publisher for it?

No, that is not the wanted behavior. I will check for this and figure out how to fix this too. Thank you :-)

KerstinKeller (approved):

KerstinKeller:
@rex-schilasky Take a look at the last commit, I think we can merge it like this.

rex-schilasky:
> @rex-schilasky Take a look at the last commit, I think we can merge it like this.

LGTM, Danke :-)

PR #1671, opened 2024-07-24
[core] active-registration-logic-getter

rex-schilasky:
### Description
Registration logic from RefreshRegistration -> ApplySample to GetRegistration changed.


KerstinKeller (approved):
:+1:

PR #1674, opened 2024-07-26
[core] registration-sender-datarace-fix

rex-schilasky:
### Description
Class member variable (sample list buffer) moved into local function scope to avoid dataraces when calling SendSampleList() out of different threads.


KerstinKeller (approved):
:+1:

PR #1675, opened 2024-07-26
Feature/unregistration provider

KerstinKeller:
### Description
Provide unregistration for timed out participants centrally in one place!


KerstinKeller:
The client gate does not yet provide an Unregistration method. Hence it keeps the expiration map and needs to be considered at a later point.

rex-schilasky (approved):
LGTM

PR #1676, opened 2024-07-26
[core] check-sample-applier-process-id-fix

rex-schilasky:
### Description
Check process id for pub/sub, service, client separately in sample applier logic.


KerstinKeller (changes requested):
What about process samples? they also have a pid :)

KerstinKeller (approved):

PR #1677, opened 2024-07-26
[core] registration-provider-refactoring

rex-schilasky:
### Description
* registration provider single threaded send logic
* Register(false) calls removed finally

### DIFF
--- a/ecal/core/src/registration/ecal_registration_provider.cpp
+++ b/ecal/core/src/registration/ecal_registration_provider.cpp
@@ -158,7 +158,7 @@ namespace eCAL
   void CRegistrationProvider::TriggerRegisterSendThread()
   {
     {
-      std::lock_guard<std::mutex> lock(m_reg_sample_snd_thread_cv_mtx);
+      const std::lock_guard<std::mutex> lock(m_reg_sample_snd_thread_cv_mtx);
       m_reg_sample_snd_thread_trigger = true;
     }
     m_reg_sample_snd_thread_cv.notify_one();
--- a/ecal/core/src/registration/ecal_registration_provider.cpp
+++ b/ecal/core/src/registration/ecal_registration_provider.cpp
@@ -97,7 +97,7 @@ namespace eCAL
     m_reg_sample_snd_thread->stop();
 
     // send process unregistration sample
-    //m_reg_sender->SendSample(Registration::GetProcessUnregisterSample());
+    ProcessSingleSample(Registration::GetProcessUnregisterSample());
 
     // delete registration sender
     m_reg_sender.reset();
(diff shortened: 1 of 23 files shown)


ecal/tests/cpp/util_test/src/util_getclients.cpp
```diff
@@ -22,15 +22,18 @@
 #include <gtest/gtest.h>
 
 enum {
-  CMN_MONITORING_TIMEOUT_MS   = (5000),
-  CMN_REGISTRATION_REFRESH_MS = (1000 * 2)
+  CMN_MONITORING_TIMEOUT_MS   = (5000 + 100),
```
KerstinKeller, line 25: Should we read the config the timeout here? or later set the timeout via the config objects? 
We could create much shorter timeouts -> much quicker test execution.

rex-schilasky: Yes, lets do this in a new PR.

ecal/core/src/registration/ecal_registration_provider.cpp
```diff
@@ -84,7 +83,7 @@ namespace eCAL
 
     // start cyclic registration thread
     m_reg_sample_snd_thread = std::make_shared<CCallbackThread>(std::bind(&CRegistrationProvider::RegisterSendThread, this));
-    m_reg_sample_snd_thread->start(std::chrono::milliseconds(Config::GetRegistrationRefreshMs()));
+    m_reg_sample_snd_thread->start(std::chrono::milliseconds(0));
```
KerstinKeller, line 86: @rex-schilasky Is this supposed to be started with 0 timeout? doesn't seem right!

ecal/core/src/registration/ecal_registration_provider.h
```diff
@@ -50,31 +46,22 @@ namespace eCAL
...
 
-    static std::atomic<bool>                    m_created;
-
-    std::unique_ptr<CRegistrationSender>        m_reg_sender;
-
-    std::shared_ptr<CCallbackThread>            m_reg_sample_snd_thread;
-
-    std::mutex                                  m_sample_list_mtx;
-    Registration::SampleList                    m_sample_list;
+    static std::atomic<bool>             m_created;
```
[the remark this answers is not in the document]
rex-schilasky: false positive

ecal/core/src/registration/ecal_registration_provider.cpp
rex-schilasky: extent the CThread implementation to support trigger functionality

KerstinKeller (approved):

PR #1680, opened 2024-07-30
[config] configuration yaml generation

Peguen:
### Description
Use defines to generate a yaml stringstream


ecal/core/include/ecal/config/configuration.h
```diff
@@ -48,31 +48,30 @@ namespace eCAL
...
-    Cli::Configuration               command_line_arguments{};
+    TransportLayer::Configuration transport_layer{};
+    Registration::Configuration   registration{};
+    Monitoring::Configuration     monitoring{};
+    Subscriber::Configuration     subscriber{};
+    Publisher::Configuration      publisher{};
+    Time::Configuration           timesync{};
+    Service::Configuration        service{};
+    Application::Configuration    application{};
+    Logging::Configuration        logging{};
```
KerstinKeller, line 59: You don't need to initialize objects explicitly, the will be default initialized. I guess it's only necessary when they need a non default constructor. However, it's also not "wrong" to do it.

ecal/core/include/ecal/config/registration.h
```diff
@@ -39,39 +39,39 @@ namespace eCAL
...
+        UDP::Configuration udp;                  /*!< UDP based registration configuration */
       };
     }
 
     struct Configuration
     {
-      unsigned int         registration_timeout;    //!< Timeout for topic registration in ms (internal) (Default: 60000)
-      unsigned int         registration_refresh;    //!< Topic registration refresh cylce (has to be smaller then registration timeout!) (Default: 1000)                                   
+      unsigned int         registration_timeout { 60000U }; //!< Timeout for topic registration in ms (internal) (Default: 60000)
+      unsigned int         registration_refresh { 1000U };  //!< Topic registration refresh cylce (has to be smaller then registration timeout!) (Default: 1000)
```
KerstinKeller, line 67: In a following PR we could change timeouts to std::Chrono objects? At least it might make it easier for the user if he sets the values programatically...

ecal/core/src/config/default_config.h
```diff
@@ -42,12 +42,12 @@ const std::string default_config = R"(#  _____     _ _
...
       port: 14000
-
+  
 
 # Monitoring configuration
 monitoring:
-  # Timeout for topic monitoring in ms (Default: 5000), increase in 1000er steps
-  timeout: 5000
+  # Timeout for topic monitoring in ms (Default: 1000), increase in 1000er steps
+  timeout: 1000
```
KerstinKeller, line 50: Didn't rex just change this back to 5000

KerstinKeller (commented):
Will have to review again together .
I like the generation of the yam file, however it will be problematic during cross compilation.
We should consider to still have it as part of the repo (checked in)...

CMakeLists.txt
```diff
@@ -103,7 +103,7 @@ option(ECAL_USE_CLOCKLOCK_MUTEX                "Use native mutex with monotonic
 set(ECAL_CORE_HAS_PROTOBUF                                                                                         ON)
 set(ECAL_CORE_BUILD_SAMPLES                                                                          ${BUILD_SAMPLES})
 set(ECAL_CORE_BUILD_TESTS                                                                         ${BUILD_ECAL_TESTS})
-set(ECAL_CORE_CONFIGURATION                                                                                        ON)
+option(ECAL_CORE_CONFIGURATION                                                                                     ON)
```
KerstinKeller, line 106: For our build, this should not be an option, we *always* what to build with the config -> yaml-cpp will be a fixed dependency for the core build.

ecal/core/include/ecal/config/configuration.h
```diff
@@ -48,33 +48,29 @@ namespace eCAL
...
+    Service::Configuration        service;
+    Application::Configuration    application;
+    Logging::Configuration        logging;
+    Cli::Configuration            command_line_arguments;
 
     ECAL_API Configuration();
     ECAL_API Configuration(int argc_, char** argv_);
     ECAL_API Configuration(const std::vector<std::string>& args_);
 
     ECAL_API void InitConfigWithDefaultYaml();
```
KerstinKeller, line 66: Naming is a bit missleading. `InitFromConfig`, `InitConfigFromFile(const std::string& yaml_path_)` -> `InitFromFile`.

-> Also, do they still need to be member functions? Maybe free functions are better suited?

ecal/core/src/config/ecal_config_initializer.cpp
```diff
@@ -201,52 +220,38 @@ namespace eCAL
       const std::string yaml_path = checkForValidConfigFilePath(yaml_path_);
       if (!yaml_path.empty())
       {
+#ifdef ECAL_CORE_CONFIGURATION
         eCAL::Config::YamlFileToConfig(yaml_path, *this);
         ecal_yaml_file_path = yaml_path;
+#else
+        std::cout << "Yaml file found at " << yaml_path << " but eCAL core configuration is not enabled." << "\n";
```
KerstinKeller, line 227: prefer eCAL log over cout?

ecal/core/src/ecal_def.h
```diff
@@ -27,6 +27,7 @@
 #include <ecal/ecal_tlayer.h>
 #include <ecal/ecal_log_level.h>
 #include <ecal/ecal_os.h>
+#include <array>
```
KerstinKeller, line 30: not necessary

ecal/core/src/types/ecal_custom_data_types.cpp
```diff
@@ -44,7 +45,7 @@ namespace eCAL
   {
 
     // IpAddressV4 definitions
-    IpAddressV4::IpAddressV4() : IpAddressV4(NET_UDP_MULTICAST_GROUP) {};
+    IpAddressV4::IpAddressV4() : IpAddressV4( "192.168.0.1" ) {};
```
KerstinKeller, line 48: we should remove this constructor....

ecal/tests/cpp/config_test/CMakeLists.txt
```diff
@@ -21,30 +21,34 @@ project(test_config)
 find_package(Threads REQUIRED)
 find_package(GTest REQUIRED)
 find_package(tclap REQUIRED)
-find_package(yaml-cpp REQUIRED)
+
+if(ECAL_CORE_CONFIGURATION)
+  find_package(yaml-cpp REQUIRED)
+endif()
 
 include(${ECAL_PROJECT_ROOT}/thirdparty/yaml-cpp/compatibility-yaml-cpp.cmake)
```
KerstinKeller, line 29: this should be moved inside `if (ECAL_CORE_CONIFGURATION)` condition

KerstinKeller (commented):
Let's address the changes...

ecal/core/include/ecal/config/configuration.h
Peguen: As we want to store the loaded file path in a protected member, so that the user cannot modify it, we need it currently as member function.

ecal/core/src/ecal.cpp
```diff
@@ -132,11 +132,6 @@ namespace eCAL
   **/
   int Initialize(eCAL::Configuration& config_, const char *unit_name_ /*= nullptr*/, unsigned int components_ /*= Init::Default*/)
   {
-    if (g_globals() == nullptr)
-    {
-      InitGlobals();
-    }
-
```
KerstinKeller, line 139: @Peguen why is this removed?

Peguen: I have put the InitGlobals() in the Configuration object. Reason was to use the eCAL logging mechanism even when eCAL::Initialize() was not called before.

KerstinKeller (commented):
Can you create a test that will induce the crash?
Or at least later for validation?

rex-schilasky:
> Can you create a test that will induce the crash? Or at least later for validation?

We should check this later in the common config validation function.

KerstinKeller (approved):
As discussed, we need to look at the initialization phase after.

PR #1699, opened 2024-08-07
[config] Added config object to monitoring initialization.

Peguen:
### Description
Introducing usage of config object for monitoring implementation.

### DIFF
--- a/ecal/core/src/monitoring/ecal_monitoring_def.h
+++ b/ecal/core/src/monitoring/ecal_monitoring_def.h
@@ -38,7 +38,7 @@ namespace eCAL
   class CMonitoring
   {
   public:
-    CMonitoring(const Monitoring::Configuration& = {});
+    CMonitoring(const Monitoring::Configuration& config_);
     ~CMonitoring();
     
     void Start();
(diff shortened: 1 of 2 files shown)


ecal/core/src/monitoring/ecal_monitoring_impl.h
```diff
@@ -47,7 +48,7 @@ namespace eCAL
   class CMonitoringImpl
   {
   public:
-    CMonitoringImpl();
+    CMonitoringImpl(const Monitoring::Configuration& = {});
```
KerstinKeller, line 51: I would not provide a default constructor here.
It's not public API so we don't need to be backwards compatible.

ecal/core/src/monitoring/ecal_monitoring_def.h
```diff
@@ -37,7 +38,7 @@ namespace eCAL
   class CMonitoring
   {
   public:
-    CMonitoring();
+    CMonitoring(const Monitoring::Configuration& = {});
```
KerstinKeller, line 41: Same here, no default arguments please 😄

KerstinKeller (approved):

PR #1701, opened 2024-08-08
[core, python] json-monitoring-api-added

rex-schilasky:
### Description
JSON monitoring API added for C++ and Python.


ecal/core/src/monitoring/ecal_monitoring_impl.cpp
```diff
@@ -553,6 +567,18 @@ namespace eCAL
...
   }
 
+  void CMonitoringImpl::GetMonitoringJSON(std::string& monitoring_, unsigned int entities_)
+  {
+    monitoring_.clear();
+
+    eCAL::pb::Monitoring monitoring_pb;
+    GetMonitoringPb(monitoring_pb, entities_);
+
+    google::protobuf::util::JsonPrintOptions options;
```
KerstinKeller, line 577: I think we have some other options which we always set, like to print empty (0) fields. Maybe take a look at the json subscriber?

rex-schilasky:
```
google::protobuf::util::JsonOptions options;
#if GOOGLE_PROTOBUF_VERSION >= 5026000
options.always_print_fields_with_no_presence = true;
#else
options.always_print_primitive_fields = true;
#endif

```

KerstinKeller: jo

lang/python/core/src/ecal_wrap.cxx
```diff
@@ -1525,6 +1525,16 @@ PyObject* mon_monitoring(PyObject* /*self*/, PyObject* /*args*/)
...
 }
 
+/****************************************/
+/*      mon_monitoring_json             */
+/****************************************/
+PyObject* mon_monitoring_json(PyObject* /*self*/, PyObject* /*args*/)
+{
+  std::string monitoring_s;
+  eCAL::Monitoring::GetMonitoringJSON(monitoring_s);
+  return(Py_BuildValue("s", ecal_getversion()));
```
KerstinKeller, line 1535: this is never, ever correct 😄

rex-schilasky: it's not :-/

rex-schilasky: fixed

rex-schilasky:
we can not merge this without API break

PR #1702, opened 2024-08-08
[config] Adding registration configuration object

Peguen:
### Description
Added configuration objects to all registration related classes.

### DIFF
--- a/ecal/core/include/ecal/config/registration.h
+++ b/ecal/core/include/ecal/config/registration.h
@@ -25,6 +25,7 @@
 #pragma once
 
 #include "ecal/ecal_os.h"
+#include "ecal/ecal_process.h"
 
 #include <stdexcept>
 #include <string>
@@ -63,14 +64,14 @@ namespace eCAL
 
     struct Configuration
     {
-      unsigned int         registration_timeout { 60000U }; //!< Timeout for topic registration in ms (internal) (Default: 60000)
-      unsigned int         registration_refresh { 1000U };  //!< Topic registration refresh cylce (has to be smaller then registration timeout!) (Default: 1000)                                   
+      unsigned int         registration_timeout { 60000U };                           //!< Timeout for topic registration in ms (internal) (Default: 60000)
+      unsigned int         registration_refresh { 1000U };                            //!< Topic registration refresh cylce (has to be smaller then registration timeout!) (Default: 1000)                                   
 
-      bool                 network_enabled      { false };  /*!< true  = all eCAL components communicate over network boundaries
-                                                                 false = local host only communication (Default: false) */
-      bool                 loopback             { true };   //!< enable to receive udp messages on the same local machine (Default: true)
-      std::string          host_group_name      { "" };     /*!< Common host group name that enables interprocess mechanisms across 
-                                                                 (virtual) host borders (e.g, Docker); by default equivalent to local host name (Default: "") */
+      bool                 network_enabled      { false };                            /*!< true  = all eCAL components communicate over network boundaries
+                                                                                           false = local host only communication (Default: false) */
+      bool                 loopback             { true };                             //!< enable to receive udp messages on the same local machine (Default: true)
+      std::string          host_group_name      { eCAL::Process::GetHostName() };     /*!< Common host group name that enables interprocess mechanisms across 
+                                                                                           (virtual) host borders (e.g, Docker); by default equivalent to local host name (Default: CurrentHostName) */
       Layer::Configuration layer;
     };
   }
--- a/ecal/core/include/ecal/config/transport_layer.h
+++ b/ecal/core/include/ecal/config/transport_layer.h
@@ -25,6 +25,7 @@
 #pragma once
 
 #include <ecal/types/ecal_custom_data_types.h>
+#include <ecal/ecal_os.h>
 
 namespace eCAL
 {
@@ -39,6 +40,16 @@ namespace eCAL
           Types::IpAddressV4 group { "239.0.0.1" }; //!< UDP multicast group base (Default: 239.0.0.1)
           unsigned int       ttl   { 3U };          /*!< UDP ttl value, also known as hop limit, is used in determining 
                                                          the intermediate routers being traversed towards the destination (Default: 3) */
+        };
+      }
+
+      namespace Local
+      {
+        struct Configuration
+        {
+          Types::IpAddressV4 group { "127.255.255.255" }; //!< UDP multicast group base (Default: 127.255.255.255)
+          unsigned int       ttl   { 1U };                /*!< UDP ttl value, also known as hop limit, is used in determining 
+                                                               the intermediate routers being traversed towards the destination (Default: 1) */
         };
       }
 
@@ -61,6 +72,9 @@ namespace eCAL
         bool                                     npcap_enabled       { false };   //!< Enable to receive UDP traffic with the Npcap based receiver (Default: false)
       
         Network::Configuration                   network;
+        const Local::Configuration               local;
+
+        ECAL_API Configuration& operator=(const Configuration& other);
       }; 
     }
 
--- a/ecal/core/src/ecal_globals.cpp
+++ b/ecal/core/src/ecal_globals.cpp
@@ -54,7 +54,7 @@ namespace eCAL
     /////////////////////
     if (registration_provider_instance == nullptr)
     {
-      registration_provider_instance = std::make_unique<CRegistrationProvider>(eCAL::GetConfiguration().registration);
+      registration_provider_instance = std::make_unique<CRegistrationProvider>(eCAL::GetRegistrationConfiguration());
       new_initialization = true;
     }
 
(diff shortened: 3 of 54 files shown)


ecal/core/src/registration/ecal_registration_sample_applier.cpp
```diff
@@ -26,18 +26,15 @@ namespace eCAL
     //////////////////////////////////////////////////////////////////
     // CSampleApplier
     //////////////////////////////////////////////////////////////////
-    CSampleApplier::CSampleApplier(bool network, bool loopback, const std::string& host_group_name, uint32_t pid)
-      :
-      m_network(network),
-      m_loopback(loopback),
-      m_host_group_name(host_group_name),
-      m_pid(pid)
+    CSampleApplier::CSampleApplier(const eCAL::Registration::Configuration& config_, uint32_t pid)
```
Peguen, line 29: Not sure here, if that is the way to go to have the whole config object stored in the sampleapplier when we only need 3 members of it. Maybe it makes sense to have structs like SReceiveAttr also for this?

ecal/core/src/types/ecal_custom_data_types.cpp
```diff
@@ -80,6 +80,7 @@ namespace eCAL
     IpAddressV4& IpAddressV4::operator=(const std::string& ip_string_)  { this->validateIpString(ip_string_); return *this; };
     IpAddressV4& IpAddressV4::operator=(const char* ip_string_)         { this->validateIpString(ip_string_); return *this; };
     IpAddressV4::operator std::string()                                 { return m_ip_address; };
```
KerstinKeller, line 82: If you have a const version (which is good 😄 ) you can remove the non-const version. You don't need to duplicate the code.

ecal/core/src/registration/udp/ecal_registration_receiver_udp.h
```diff
@@ -26,6 +26,7 @@
 
 #include <memory>
 #include <registration/ecal_registration_types.h>
+#include "io/udp/ecal_udp_receiver_attr.h"
```
KerstinKeller, line 29: The includes can usually give you a hint, if we're dealing with the right level of abstraction here.
In this case it feels a bit strange that we're including from a different part of ecal (e.g. from io/udp, whereas this file is in registration/udp), so maybe we need to reiterate a bit.

ecal/core/src/registration/udp/ecal_registration_receiver_udp.cpp
```diff
@@ -19,34 +19,16 @@
...
-    attr.rcvbuf = UDP::GetReceiveBufferSize();
-    return attr;
-  }
-
-}
-
 using namespace eCAL;
 
-eCAL::CRegistrationReceiverUDP::CRegistrationReceiverUDP(RegistrationApplySampleCallbackT apply_sample_callback)
+eCAL::CRegistrationReceiverUDP::CRegistrationReceiverUDP(RegistrationApplySampleCallbackT apply_sample_callback, const UDP::SReceiverAttr& attr_)
```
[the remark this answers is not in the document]
KerstinKeller: In this case, things feel a bit unsymetrical.  `CRegistrationReceiverSHM` takes `Registration::Layer::SHM::Configuration`, but `CRegistrationReceiverUDP` takes `UDP::SReceiverAttr` and not an `Registration::Layer::UDP::Configuration` object.
Maybe this can be handled in a similar way.

ecal/core/include/ecal/config/registration.h
```diff
@@ -63,14 +64,14 @@ namespace eCAL
...
-                                                                 false = local host only communication (Default: false) */
-      bool                 loopback             { true };   //!< enable to receive udp messages on the same local machine (Default: true)
-      std::string          host_group_name      { "" };     /*!< Common host group name that enables interprocess mechanisms across 
-                                                                 (virtual) host borders (e.g, Docker); by default equivalent to local host name (Default: "") */
+      bool                 network_enabled      { false };                            /*!< true  = all eCAL components communicate over network boundaries
+                                                                                           false = local host only communication (Default: false) */
+      bool                 loopback             { true };                             //!< enable to receive udp messages on the same local machine (Default: true)
+      std::string          host_group_name      { eCAL::Process::GetHostName() };     /*!< Common host group name that enables interprocess mechanisms across
```
KerstinKeller, line 73: I don't like the coupling here to `ecal_process.h`. I understand it, but I don't like it 😄

ecal/core/include/ecal/config/transport_layer.h
```diff
@@ -25,6 +25,7 @@
 #pragma once
 
 #include <ecal/types/ecal_custom_data_types.h>
+#include <ecal/ecal_os.h>
```
KerstinKeller, line 28: Why  is this include necessary? is it for the `ECAL_API` macro?

ecal/core/src/ecal_globals.cpp
```diff
@@ -54,7 +54,7 @@ namespace eCAL
     /////////////////////
     if (registration_provider_instance == nullptr)
     {
-      registration_provider_instance = std::make_unique<CRegistrationProvider>();
+      registration_provider_instance = std::make_unique<CRegistrationProvider>(eCAL::GetConfiguration().registration);
```
KerstinKeller, line 57: This just looks a bit strange. We could think if we want tot have distinct functions like `eCAL::GetRegistrationConfiguration()` or something in that direction.

ecal/core/src/registration/ecal_registration_provider.cpp
```diff
@@ -63,23 +91,22 @@ namespace eCAL
...
     {
-      m_reg_sender = std::make_unique<CRegistrationSenderSHM>();
+      m_reg_sender = std::make_unique<CRegistrationSenderUDP>(CreateUDPSenderAttr(m_config));
     }
     else
     {
-#endif
-      m_reg_sender = std::make_unique<CRegistrationSenderUDP>();
-#if ECAL_CORE_REGISTRATION_SHM
+      eCAL::Logging::Log(log_level_warning, "[CRegistrationProvider] No registration layer enabled.");
```
KerstinKeller, line 107: 👍 
Do you think warning is enough? I would argue maybe even error log would be the appropriate level?
We could also think if this is something that should be validated when validating the configuration.

ecal/core/src/registration/ecal_registration_provider.h
```diff
@@ -60,8 +61,7 @@ namespace eCAL
 
     std::mutex                           m_applied_sample_list_mtx;
     Registration::SampleList             m_applied_sample_list;
-      
-    bool                                 m_use_registration_udp;
-    bool                                 m_use_registration_shm;
+
+    Registration::Configuration          m_config;
```
KerstinKeller, line 65: Here I am a bit undecided. Is it necessary to store the whole config? (Looking at the cpp file, I guess it is, because udp / shm configs are needed later on). If everything was done during initialization, I guess it would not be necessary.

ecal/core/src/registration/ecal_registration_sample_applier.cpp
KerstinKeller: Yes, I think we need partial stucts, something that fits to the objects which are instantiated. These can be created from the actual config objects.

Often it might be a 1:1 translation, but sometimes it might be merged from different subcompontents of the configuration objects.

KerstinKeller: Also I think in this case, the sampleApplier design is not yet optimal, because it does more than one thing. It decides if a sample should be applied (filtering samples), and it applies the sample to everyone who is interested (distributing samples).
Probably the logic if the sample should be applied should reside somewhere else.
Instead the sample applier should be constructed with an `accept_sample` function, so the logic about this can be moved to another entity.

You can also see that the design is not ideal, because you cannot test these two functionalitiess independently (filtering samples and distributing samples).

ecal/core/src/registration/shm/ecal_memfile_broadcast.h
```diff
@@ -79,7 +80,7 @@ namespace eCAL
   public:
     CMemoryFileBroadcast();
 
-    bool Create(const std::string& name, std::size_t max_queue_size);
+    bool Create(const Registration::Layer::SHM::Configuration& config_);
```
KerstinKeller, line 83: I don't think i would pull in the config "that deep" into the internals of shm. This class only needs two parameters, now it gets a lot more.

KerstinKeller (commented):
All in all pretty good. I think that there are a few points that we should discuss a bit further 😄

ecal/core/src/registration/shm/ecal_memfile_broadcast.h
Peguen: Actually in this case its just another boolean value (for enable layer) that is in the config object in addition. I was more worried about overhead in the registration_sample_applier

ecal/core/src/registration/ecal_registration_provider.cpp
Peguen: In case it would stop eCAL from working I would agree with the error level. But isn't it possible to have a configuration without the registration layer, lets say for monitoring purposes? In that case I think warning is enough, as it might not be a common configuration and could also be configured by mistake.

ecal/core/src/ecal_globals.cpp
Peguen: I also thought about that. Lets try that out

ecal/core/include/ecal/config/transport_layer.h
Peguen: Yes, correct

ecal/core/include/ecal/config/registration.h
Peguen: As discussed: we keep this implementation for now, but check for a next PR the logic implementation in where it is used and adapt accordingly

PR #1704, opened 2024-08-09
[config] Adding config object to logging

Peguen:
### Description
tbd


Peguen:
Closed because of problems after merged master. Using https://github.com/eclipse-ecal/ecal/pull/1724 instead.

PR #1708, opened 2024-08-13
[core] feature/id-based-descgate

rex-schilasky:
### Description
Descgate API changed to use separate GetEntityIDs() + GetEntityInfo(id) functions instead of single GetEntities() variants.

### DIFF
--- a/ecal/core/src/ecal_descgate.cpp
+++ b/ecal/core/src/ecal_descgate.cpp
@@ -218,11 +218,7 @@ namespace eCAL
     topic_quality_info.quality = topic_quality_;
 
     const std::unique_lock<std::mutex> lock(topic_info_map_.mtx);
-    auto iter = topic_info_map_.map.find(topic_info_key);
-    if (iter == topic_info_map_.map.end())
-    {
-      topic_info_map_.map[topic_info_key] = topic_quality_info;
-    }
+    topic_info_map_.map[topic_info_key] = topic_quality_info;
   }
 
   void CDescGate::RemTopicDescription(SQualityTopicIdMap& topic_info_map_,
@@ -251,11 +247,7 @@ namespace eCAL
     service_quality_info.response_quality   = response_type_quality_;
 
     const std::lock_guard<std::mutex> lock(service_method_info_map_.mtx);
-    auto iter = service_method_info_map_.map.find(service_method_info_key);
-    if (iter == service_method_info_map_.map.end())
-    {
-      service_method_info_map_.map[service_method_info_key] = service_quality_info;
-    }
+    service_method_info_map_.map[service_method_info_key] = service_quality_info;
   }
 
   void CDescGate::RemServiceDescription(SQualityServiceIdMap& service_method_info_map_,
--- a/ecal/core/src/ecal_descgate.cpp
+++ b/ecal/core/src/ecal_descgate.cpp
@@ -46,54 +46,98 @@ namespace eCAL
   CDescGate::CDescGate() = default;
   CDescGate::~CDescGate() = default;
 
-  Registration::QualityTopicInfoMultiMap CDescGate::GetPublishers()
+  std::set<Registration::STopicId> CDescGate::GetPublisherIDs() const
   {
-    return GetTopics(m_publisher_info_map);
+    return GetTopicIDs(m_publisher_info_map);
   }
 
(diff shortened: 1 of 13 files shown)


ecal/core/src/ecal_descgate.cpp
```diff
@@ -162,69 +206,81 @@ namespace eCAL
...
+                                        const std::string& topic_name_,
+                                        const SDataTypeInformation& topic_info_,
+                                        const Registration::DescQualityFlags topic_quality_)
   {
-    const auto topic_info_key = STopicIdKey{ topic_name_, topic_id_ };
+    const auto topic_info_key = Registration::STopicId{ topic_id_, topic_name_ };
 
     Registration::SQualityTopicInfo topic_quality_info;
-    topic_quality_info.id      = topic_id_;
+    topic_quality_info.id      = std::stoull(topic_id_.entity_id);
```
KerstinKeller, line 217: do we still need this id field here? can we remove it?

ecal/core/src/ecal_descgate.cpp
```diff
@@ -162,69 +206,81 @@ namespace eCAL
...
     topic_quality_info.info    = topic_info_;
     topic_quality_info.quality = topic_quality_;
 
     const std::unique_lock<std::mutex> lock(topic_info_map_.mtx);
-    topic_info_map_.map[topic_info_key] = topic_quality_info;
+    auto iter = topic_info_map_.map.find(topic_info_key);
+    if (iter == topic_info_map_.map.end())
+    {
+      topic_info_map_.map[topic_info_key] = topic_quality_info;
+    }
```
KerstinKeller, line 226: I don't think this is necessarily correct (at least atm) since description information can change at runtime. so I guess we need to overwrite it with every update.

ecal/core/src/ecal_descgate.cpp
```diff
@@ -162,69 +206,81 @@ namespace eCAL
...
+    service_quality_info.id                 = std::stoull(service_id_.entity_id);
     service_quality_info.info.request_type  = request_type_information_;
     service_quality_info.info.response_type = response_type_information_;
     service_quality_info.request_quality    = request_type_quality_;
     service_quality_info.response_quality   = response_type_quality_;
 
     const std::lock_guard<std::mutex> lock(service_method_info_map_.mtx);
-    service_method_info_map_.map[service_method_info_key] = service_quality_info;
+    auto iter = service_method_info_map_.map.find(service_method_info_key);
+    if (iter == service_method_info_map_.map.end())
```
KerstinKeller, line 257: same as above.

ecal/core/src/ecal_descgate.cpp
rex-schilasky: I checked that, yes we can remove it finally.

ecal/core/src/registration/ecal_registration.cpp
```diff
@@ -161,10 +161,68 @@ namespace eCAL
...
+      return g_descgate()->GetClientIDs();
+    }
+
+    bool GetClientInfo(const SServiceId& id_, SQualityServiceInfo& service_info_)
+    {
+      if (g_descgate() == nullptr) return false;
+      return g_descgate()->GetClientInfo(id_, service_info_);
+    }
+
     QualityTopicInfoMultiMap GetPublishers()
```
KerstinKeller, line 212: @rex-schilasky didn't you want to throw out these functions?

ecal/core/src/serialization/ecal_struct_sample_registration.h
```diff
@@ -235,25 +235,6 @@ namespace eCAL
       }
     };
 
-    struct SampleIdentifier
```
KerstinKeller, line 238: I'm not happy here. I would like to keep this type here, even if this means copying at one point. (where it's made available to the user). We're now mixing up internally used datatypes and the datatypes of the public API, which weakens our flexibility (only public API needs to be stable, not the internal one).

ecal/core/src/registration/ecal_registration.cpp
rex-schilasky: Not yet ..

KerstinKeller (approved):
But the next PR, we'll introduce the new type!

PR #1709, opened 2024-08-14
[config] Added config object to registration.

Peguen:
### Description
Added config object to registration layer.
Also added component SAttributes and Attribute Builder to registration layer.

### DIFF
--- a/ecal/core/src/registration/ecal_registration_receiver.cpp
+++ b/ecal/core/src/registration/ecal_registration_receiver.cpp
@@ -42,6 +42,9 @@
 #include <mutex>
 #include <string>
 
+#include "builder/udp_shm_attribute_builder.h"
+#include "builder/sample_applier_attribute_builder.h"
+
 namespace eCAL
 {
   //////////////////////////////////////////////////////////////////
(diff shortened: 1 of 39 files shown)


ecal/core/src/registration/ecal_registration_sample_applier.h
```diff
@@ -59,10 +59,7 @@ namespace eCAL
 
       bool AcceptRegistrationSample(const Registration::Sample& sample_);
 
-      bool                                  m_network;
-      bool                                  m_loopback;
-      std::string                           m_host_group_name;
-      int32_t                               m_pid;
+      SampleApplier::SAttributes                  m_attributes;
```
KerstinKeller, line 63: I am a bit torn weather I think that while you have the attributes in the constructor, individual member variables would be - more readable (?) - than the attributes. But I guess I am still undecided.

Also just a thought: In this case - or in more cases, it could make sense to define the attributes simply as a subclass inside CSampleApplier class?
I think the decision when a separate header file is necessary depends on the context in which the attributes / configs may be used. Simply as types also in a context without the class? Or always tied to the usage of the class.

In this case they are always tied to the class usage, so they might be a subclass.
With the configs, they are used in different places and passed around by the user, so it makes sense to encapsulate them in a separate `.h` file.

ecal/core/src/registration/ecal_registration_receiver.cpp
```diff
@@ -96,19 +98,30 @@ namespace eCAL
     m_timeout_provider_thread = std::make_unique<CCallbackThread>([this]() {m_timeout_provider->CheckForTimeouts(); });
     m_timeout_provider_thread->start(std::chrono::milliseconds(100));
 
-    // receive registration via udp or shared memory
-    m_use_registration_shm = Config::IsShmRegistrationEnabled();
-    m_use_registration_udp = !m_use_registration_shm;
+    m_timeout_provider = std::make_unique<Registration::CTimeoutProvider<std::chrono::steady_clock>>(
```
KerstinKeller, line 101: Did you duplicated this code? it's already in line 87... maybe during merge...

Peguen: Probably some merge conflict, yes. Solved

ecal/core/src/registration/ecal_registration_sample_applier.h
Peguen: That's a good point. I like also the overview directly, what are the member of the struct. I will try this out on the monitoring branch, let's have a look there to make a decision.

PR #1714, opened 2024-08-20
[core] remove monitoring timeout

KerstinKeller:
### Description
Remove the monitoring timeout from Config (+API function) because it is no longer used within eCAL.
Additionally, the default registration timeout is set to 10 seconds as a compromise between a stable connection and up to date display in eCAL Monitor.


rex-schilasky (approved):

Peguen (approved):
Looks good!

PR #1717, opened 2024-08-21
[core] add publisher registration event callback API

rex-schilasky:
### Description
See title ..

### DIFF
--- a/ecal/core/src/ecal_descgate.cpp
+++ b/ecal/core/src/ecal_descgate.cpp
@@ -64,6 +64,22 @@ namespace eCAL
     return GetTopic(id_, m_publisher_info_map, topic_info_);
   }
 
+  Registration::CallbackToken CDescGate::RegisterPublisherEventCallback(const Registration::TopicIDCallbackT& callback_)
+  {
+    std::lock_guard<std::mutex> lock(m_publisher_info_map.mtx);
+
+    Registration::CallbackToken new_token = CreateToken();
+    m_publisher_info_map.cb_map[new_token] = callback_;
+
+    return new_token;
+  }
+
+  void CDescGate::UnregisterPublisherEventCallback(Registration::CallbackToken token_)
+  {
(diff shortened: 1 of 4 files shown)


ecal/core/src/registration/shm/ecal_registration_sender_shm.cpp
```diff
@@ -54,13 +54,13 @@ bool eCAL::CRegistrationSenderSHM::SendSampleList(const Registration::SampleList
 {
   bool return_value{true};
   // serialize whole sample list
-  std::vector<char> sample_list_buffer;
-  if (SerializeToBuffer(sample_list, sample_list_buffer))
+  m_sample_list_buffer.clear();
```
KerstinKeller, line 57: Do we need a mutex now that's a member variable? Any chance that function may be called concurrently?

ecal/core/src/ecal_descgate.cpp
```diff
@@ -278,7 +334,15 @@ namespace eCAL
...
+      service_method_info_map_.id_map.erase(service_method_info_key);
     }
   }
+
+  Registration::CallbackToken CDescGate::CreateToken()
+  {
+    // Atomically increment m_callback_token using fetch_add to ensure thread safety.
+    // fetch_add returns the value before increment, so we add 1 to get the new token value.
+    // memory_order_relaxed is used to optimize performance without additional synchronization.
+    return m_callback_token.fetch_add(1, std::memory_order_relaxed) + 1;
```
KerstinKeller, line 346: Is it really necessary to add one in the end? First call yields 0, second one 1, ...

ecal/core/src/registration/shm/ecal_registration_sender_shm.cpp
rex-schilasky: This is an internal API and is called by the registration send thread only.

Peguen (approved):
looks good!

PR #1719, opened 2024-08-23
[core] Make STopicId available in subscriber callbacks

KerstinKeller:
- Add new Callback Type which includes STopicId (ecal_callback.h)
- Remove topic_name and topic_id from memfile_pool
- Subgate applies Payload::Topic

TODO: take care of PID for TCP and UDP layer.

### Description
We don't yet have the information in callbacks about *who* sent a message, we only receive the topic name. This PR fixes this in a way that callbacks receive an `STopicID` containing both topic name and additional information (unique sender id).

### DIFF
--- a/ecal/core/src/io/shm/ecal_memfile_pool.cpp
+++ b/ecal/core/src/io/shm/ecal_memfile_pool.cpp
@@ -96,7 +96,7 @@ namespace eCAL
     return true;
   }
 
-  bool CMemFileObserver::Start(const std::string& topic_name_, const std::string& topic_id_, const int timeout_, const MemFileDataCallbackT& callback_)
+  bool CMemFileObserver::Start(const int timeout_, const MemFileDataCallbackT& callback_)
   {
     if (!m_created)     return false;
     if (m_is_observing) return false;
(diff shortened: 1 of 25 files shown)


KerstinKeller:
PID is currently not transferred correctly in UDP und TCP layer.
This information can be passed via CUDPReaderLayer, CTCTReaderLayer.

Otherwise things look good. Might consider some renaming for the future, but all to be discussed.

rex-schilasky:
> PID is currently not transferred correctly in UDP und TCP layer. This information can be passed via CUDPReaderLayer, CTCTReaderLayer.
> 
> Otherwise things look good. Might consider some renaming for the future, but all to be discussed.

- Callback function type renamed to ReceiveIDCallbackT
- UDP and TCP layer transport host name and process id as well now

ecal/core/src/io/shm/ecal_memfile_pool.cpp
```diff
@@ -108,11 +108,11 @@ namespace eCAL
...
 
     // start observer thread
-    m_thread = std::thread(&CMemFileObserver::Observe, this, topic_name_, topic_id_, timeout_);
+    m_thread = std::thread(&CMemFileObserver::Observe, this, timeout_);
 
 #ifndef NDEBUG
     // log it
-    Logging::Log(log_level_debug2, std::string("CMemFileObserver started (" + topic_name_ + ", " + topic_id_ + ")"));
+    Logging::Log(log_level_debug2, std::string("CMemFileObserver started."));
 #endif
```
KerstinKeller, line 116: Should we maybe log the memfile name (currently it's not, but we could make it a member variable)

KerstinKeller (commented):
Thanks for making the adaptions.

I'm a bit undecided. The changes that you made will work, but it will introduce an incompatibility, if different eCAL versions now communicate. The callbackID will be "wrong" for anyone receiving messages send by an older eCAL version.
And what is even worse, if you will then want to look up the corresponding descriptors to this ID from the descgate, you will get missmatches, or no info.

I wonder if we can make adjustments based on what we know from registration only.
E.g. on registration layer we already receive the info... tid (our unique identifier) - relates to pid X, topic_name Y, host_name Z. ...
Thus the ecal_reader layers can keep track of the information, and based on the incoming tid fill the rest of the information. Then you do not have to send it over the layers together with the actual data.

rex-schilasky (approved):
Great

PR #1724, opened 2024-08-28
[config] Adding Attributes to logging

Peguen:
### Description
Adapted logging implementation to use attributes instead of config.
Adapted logging to output warnings to console in case an enabled logging mechanism cannot be executed.
Added logging filter methods to change logging filter for each sink on runtime.
Added logging tests.


KerstinKeller (approved):
👍

PR #1727, opened 2024-09-05
[config] Move complete memfile config to publisher

Peguen:
### Description
The memfile settings where split between transport_layer and publisher config. Now all is available only in the publisher config to enable individual configuration for each publisher.


rex-schilasky (approved):
DaNkE !!

PR #1729, opened 2024-09-05
[core] new-datarw-connection-logic

rex-schilasky:
#### Refactor connection state logic for DataReader/Writer (Sub/Pub)

- Corrected connection state handling to trigger IsPublished()/IsSubscribed() and (corresponding connection state events) only when the connection is fully established.
- Using a second registration loop to ensure connections become fully active.
- Send() and Receive() functions remain unaffected by the new state management.
- Added tests to validate correct behavior of IsPublished() and IsSubscribed() methods.


Peguen (approved):
💻 lgtm 👍

PR #1734, opened 2024-09-12
[protobuf] Readded msg_size and field name check to ProcProtoMsg recursion

Peguen:
### Description
To avoid infinite recursive for recursive proto definitions readded the msg_size check and additionally a field name find check. The later enables the msg to show at least one default submsg, even if every field is just a default value.


KerstinKeller (approved):
Sounds good, as long as we avoid infinite recursion 😄

PR #1736, opened 2024-09-13
[core] provide SDataTypeInformation in callbacks

KerstinKeller:
We now have the possibility to provide the DatatypeInformation in Callbacks, which makes it much easier for dynamic subscribers to deserialize data, without needing to look it up elsewhere.

@rex-schilasky 
I now have a few questions:
- We only provide this information for the callbacks. What about the `Receive` method?
- What about callback and receive signatures for protobuf subscribers and dynamic subscribiers?


rex-schilasky (approved):
LGTM. DANKE :-)

PR #1742, opened 2024-09-23
[core] minor performance improvements

KerstinKeller:
### Description
These are some minor performance improvements for eCAL core, to avoid unnessecary copies or indexing into potentially large maps.


Peguen (approved):
Let's speed things up 
🚙 💨

PR #1743, opened 2024-09-24
[python] Bugfix mon_monitoring() crash

Peguen:
### Description
Use correct values for ecal_wrap.cxx connections_loc/ext and message_drops.
Also set ecal_wrap.cxx connections_loc/ext and message_drops to int, otherwise mon_monitoring() function will cause the algo to crash.


KerstinKeller (approved):
👍

PR #1744, opened 2024-09-24
[core] Reduce allocation / deallocations necessary for Registration.

KerstinKeller:
### Description
In bigger, idle eCAL System there is quite a large "Grundlast" in the system, due to how Registrations are processed.
This PR tries to solve that, by providing a `Util::CExpandingVector` which only deallocates members upon destruction, and avoiding any unnessesary instatiations of `eCAL::Sample` instances.


hannemn (approved):
Looks good to me as reviewed together in the meetingv yesterday.

PR #1745, opened 2024-09-26
[core] Introduce CMonitoringFilter class, that takes the task of filtering the topics to be included in the monitoring struct.

KerstinKeller:
### Description
This PR moves the implementation of the filtering of samples which are applied to the monitoring to it's own class, called CMonitoringFilter. It provides methods to set the filters (via an include and exclude list), and a method to check if a given topic name is accepted or filtered out.

At the same time it also provides some performance improvements over the original implementation, where a regex was "compiled" for every incoming sample, which is a very expensive operation. Now the regular expressions themselves are created once, and they are matched every time a sample is applied.

This give significant performance improvement, however, the performance of performing a the matching is still significant.
We should consider either not providing a filtering mechanism for the monitoring, or, to give that mechanism to the user, to be applied when iterating over the data, which might be performed less often than the sample application.
Reverting to faster filtering mechanisms (like startswith..) might also be an alternative.

In a "massive pubsub scenario", with the system idling, there is significant performance spend for the monitoring component.
image
e.g. 7.5% of time spend for monitoring, and 2.7% of all execution time spend only for matching the regex within monitoring. That's quite expensive.

### DIFF
--- a/ecal/core/CMakeLists.txt
+++ b/ecal/core/CMakeLists.txt
@@ -193,6 +193,8 @@ if(ECAL_CORE_MONITORING)
 set(ecal_monitoring_src
     src/monitoring/ecal_monitoring_def.cpp
     src/monitoring/ecal_monitoring_def.h
+    src/monitoring/ecal_monitoring_filter.cpp
+    src/monitoring/ecal_monitoring_filter.h
     src/monitoring/ecal_monitoring_impl.cpp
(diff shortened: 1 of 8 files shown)


hannemn (approved):
Looks good to me from code point of view. I didn't countercheck the implementation on my end, but since you covered the functionalty with test cases I'll assume that it works properly as the previous monitorä filter function (for sure with improved performance).

PR #1746, opened 2024-09-27
[MonGUI] minor performance improvements. 

KerstinKeller:
- Do not get a topic type / encoding from utils if this topic has not set them.
- Avoid copies


Peguen (approved):
🐎

PR #1754, opened 2024-10-01
Fix memory leaks in Python wrapper by adding missing Py_DECREF.

### LINKED ISSUE #1753: Memory leak in eCAL Python binding
### Problem Description

The following simple Python monitoring sample leaks memory:
 
 ```
 import ecal.core.core as ecal_core
import sys
import time

def main():
    ecal_core.initialize(sys.argv, "Python Monitoring")
    ecal_core.mon_initialize()
    
    while ecal_core.ok():
        try:
            topics = ecal_core.mon_monitoring()[1]['topics']
            for topic in topics:
                print(topic['tname']) 
        except Exception:    
            pass
        time.sleep(1)
       
    
if __name__ == "__main__":
    main()
```

### Problem
The problem is that `mon_monitoring` (and also `mon_logging` ) doesn't properly decrease all internal references and thus leaks memory.

### How did you get eCAL?

I don't know

### Environment

master, v5.12, v5.13.

KerstinKeller:
### Description
When writing Python extensions from C++, objects created with `Py_BuildValue`, `Py_....` need their references decremented unless they are returned directly from the functions.
This PR fixes some occurences where the ref counters where not decreased, which leads to memory leaks.

### Related issues
Fixes #1753

### DIFF
--- a/lang/python/core/src/ecal_wrap.cxx
+++ b/lang/python/core/src/ecal_wrap.cxx
@@ -1301,7 +1301,8 @@ PyObject* mon_monitoring(PyObject* /*self*/, PyObject* /*args*/)
     }
   }
 
-  return(Py_BuildValue("iO", 0, retDict));
+  auto* retVal = Py_BuildValue("iO", 0, retDict); Py_DECREF(retDict);
+  return(retVal);
 }
 
 /****************************************/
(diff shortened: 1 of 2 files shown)


hannemn (approved):
Beautiful! :-D

PR #1760, opened 2024-10-02
[python] Call_method in service now allows threads.

Peguen:
### Description
If you used a service client and server in the same process (like described in this issue: https://github.com/eclipse-ecal/ecal/issues/1755), the application locks. With this fix, it does not lock anymore and function as expected.


KerstinKeller (approved):
Somehow we missed that 👍

PR #1764, opened 2024-10-08
[core] fix bug that processes will not be "unregistered" correctly.

### LINKED ISSUE #1763: Processes are not unregistered properly on master
### Problem Description

In the current master build, it seems that processes are not unregistered properly. Neither when they are shut down gracefully, nor when they are killed.
At least they are still present in eCAL Monitor.

### How to reproduce

Start some samples (person_snd, person_rec) and kill them (via ecal_sys, or just close them).
Observe that they are not removed from processes view in eCAL Monitor.

### How did you get eCAL?

I don't know

### Environment

Current master

### eCAL System Information

_No response_

KerstinKeller:
### Description
Correctly keeps processes in CExpiringMap of unregistration provider.
Now their id is the `process_id` instead of `""`, which lead to process entries overwriting themselves in tthe map kept by the unregistration provider.

### Related issues
Fixes #1763


rex-schilasky (approved):
LGTM

PR #1790, opened 2024-11-12
check event callback before execution

rex-schilasky:
### Description
eCAL service client is offering event callback functionality. Like this

```c++
const eCAL::service::ClientSession::EventCallbackT event_callback = []
  (eCAL::service::ClientEventType /*event*/, const std::string& /*message*/) -> void
   {
   };
client_session = client_manager->create_client(protocol_version, endpoint_list, event_callback);
```

If the event callback logic is not relevant for your use case I would expect to just pass a nullptr like this
```c++
client_session = client_manager->create_client(protocol_version, endpoint_list, nullptr);
```

but this leads to a crash.

### Related issues
None

### Cherry-pick to
- _none_

### DIFF
--- a/ecal/service/ecal_service/src/client_session_impl_v0.cpp
+++ b/ecal/service/ecal_service/src/client_session_impl_v0.cpp
@@ -1,6 +1,6 @@
 /* ========================= eCAL LICENSE =================================
  *
- * Copyright (C) 2016 - 2019 Continental Corporation
+ * Copyright (C) 2016 - 2024 Continental Corporation
  *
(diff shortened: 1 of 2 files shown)


KerstinKeller (commented):
Actually, I don't necessarily expect that you can pass a nullptr, you could just pass an empty lambda.
But I guess at least it would need to be documented.

On the other hand, you should't be able to use API incorrectly, so I guess a check is not too bad.

KerstinKeller (approved):

PR #1795, opened 2024-11-14
hotfix/memfile-ack-logic-deadlock

rex-schilasky:
### Description

If the shm data transport layer acknowledgment feature is used and the ack time is set to a large value the publisher send may wait for a this long time on the ack event from the subscriber even the subscriber is no more existing (process is terminated for example). The SHM Datawriter Connect/Disconnect API is locked as long the Datawriter is waiting for the ack signal. That leads to a lock of the Registration layer finally.

Solution:

1. Add `RemoveSubscription` to `CDataWriterSHM` to react on un-registrations from the registration layer in general
2. Decouple `Connect/Disconnect` functionality of `CSyncMemoryFile` from `SyncContent()` to be able to modify the map of the memfile connection events even `SyncContent()` is waiting on a sync event.

### DIFF
--- a/ecal/core/src/readwrite/shm/ecal_writer_shm.cpp
+++ b/ecal/core/src/readwrite/shm/ecal_writer_shm.cpp
@@ -97,7 +97,7 @@ namespace eCAL
     }
   }
 
-  void CDataWriterSHM::RemoveSubscription(const std::string& host_name_, const int32_t process_id_, const std::string& topic_id_)
+  void CDataWriterSHM::RemoveSubscription(const std::string& host_name_, const int32_t process_id_, const std::string&  /*topic_id_*/)
   {
     // we accept local disconnections only
     if (host_name_ != m_attributes.host_name) return;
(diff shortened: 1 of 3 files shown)


ecal/core/src/readwrite/shm/ecal_writer_shm.cpp
```diff
@@ -97,6 +104,47 @@ namespace eCAL
     }
   }
 
+  void CDataWriterSHM::RemoveSubscription(const std::string& host_name_, const int32_t process_id_, const std::string& topic_id_)
+  {
+    // we accept local disconnections only
+    if (host_name_ != m_attributes.host_name) return;
```
KerstinKeller, line 110: I am not sure that this is correct. If we communicate through docker images, we might have different host_name, but identical host_group.
Anyways, who applies the RemoveSubscription?
If you remove only subscriptions from IDs which have previously been subscribed, you don't need to do any checks here.

KerstinKeller: This seems to be a general issue, and will be adressed with  #1804

KerstinKeller (approved):
:+1:

PR #1806, opened 2024-11-26
[config] Remove simpleini as submodule

Peguen:
### Description
Submodule and simpleini dependencies removed.
ecaltime plugin switched to using yaml.


doc/rst/conf.py
```diff
@@ -92,7 +92,7 @@
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
 extensions = [
-    'sphinx_book_theme',
```
KerstinKeller, line 95: @Peguen why did you remove this line? It seems unrelated.

PR #1816, opened 2024-11-27
[submodules] Updated HDF5 to 1.14.3

FlorianReimold:
Updated HDF5 to ~1.14.5~ 1.14.3


FlorianReimold:
The issue is probably not eCAL, but HDF5 1.14.5 itself, which recently broke UTF8 / Unicode support on Windows: https://github.com/HDFGroup/hdf5/issues/5037

So I suggest we downgrade HDF5 again to some non-affected version, until this is resolved.

Peguen (approved):

PR #1831, opened 2024-12-02
[Logging] Introducing Provider/Receiver pattern for logging

Peguen:
### Description
Split of the ecal logging mechanism into a provider and receiver class. Receiver class is an UDP receiver for all log messages. The receiving can be enabled/disabled by config. Additionally, it must be stated explicitly that the receiving component needs to be initialized when eCAL::Initialize(...) is called. Receiving is by default disabled.

Tests and apps adapted towards this logic.


ecal/core/src/ecal_global_accessors.h
```diff
@@ -66,7 +71,8 @@ namespace eCAL
 
   // Declaration of getter functions for globally accessible variable instances
   CGlobals*               g_globals();
-  CLog*                   g_log();
+  Logging::CLogReceiver*  g_log_receiver();
```
KerstinKeller, line 74: Should we make clear from the naming, that this receiver receives only UDP log traffic? (We don't read log files or anything...)

ecal/core/src/ecal_globals.cpp
```diff
@@ -190,9 +190,15 @@ namespace eCAL
     /////////////////////
     if ((components_ & Init::Logging) != 0u)
```
KerstinKeller, line 191: I think it makes sense here now to differentiate Logging and receiving UDP logging traffic. Logging itself should probably be initialized by default, whereas receiving should be off. (Only e.g. monitor applications should activate it).

ecal/core/src/logging/config/attributes/ecal_log_provider_attributes.h
```diff
@@ -0,0 +1,76 @@
...
+      bool        broadcast;
+      bool        loopback;
+      int         sndbuf;
+    };
+
+    struct SProviderAttributes
+    {
+      SProviderUDP           udp;
+      SFile                  file;
+      SConsole               console;
```
KerstinKeller, line 64: Now the naming feels akward. Either all three should be called `SProviderXXX` or none should be. Maybe the structs could be defined within this class, then it's clear that they are all provider attributes and just called `SUDP`, `SFile`, ...

ecal/core/src/logging/config/attributes/ecal_log_receiver_attributes.h
```diff
@@ -0,0 +1,60 @@
...
+      bool        broadcast;
+      bool        loopback;
+      int         rcvbuf;
+    };
+
+    struct SReceiverAttributes
+    {
+      SReceiverUDP           udp;
+
+      SUDPReceiver           udp_receiver;
```
KerstinKeller, line 50: What's the difference between `SReceiverUDP` and `SUDPReceiver`. It feels a bit strange and they both have the `port` attribute.

ecal/core/src/config/builder/logging_attribute_builder.cpp
```diff
@@ -6,9 +6,9 @@ namespace eCAL
 {
   namespace Logging
   {
-    SAttributes BuildLoggingAttributes(const Logging::Configuration& log_config_, const Registration::Configuration& reg_config_, const TransportLayer::Configuration& tl_config_)
+    SProviderAttributes BuildLoggingProviderAttributes(const Logging::Configuration& log_config_, const Registration::Configuration& reg_config_, const TransportLayer::Configuration& tl_config_)
```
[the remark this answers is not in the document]
Peguen: Included in header

ecal/core/include/ecal/config/logging.h
```diff
@@ -69,6 +69,7 @@ namespace eCAL
           bool                enable         { true };               //!< Enable UDP logging (Default: false)
           unsigned int        port           { 14001 };              //!< UDP port number (Default: 14001)
           eCAL_Logging_Filter filter_log_udp { log_filter_default }; //!< Log messages logged via udp network (Default: info, warning, error, fatal)
+          bool                receive        { false };              //!< Enable receiving udp log messages (Default: false)
```
KerstinKeller, line 72: This is the configuration for a UDP File logging sink. Doesn't it make more sense to have a eCAL::Logging::UDPReceiver::Configuration object, where you also provide the port?
And maybe an `enabled` flag which defaults to false.

Or what other information do you need to create a UDP Logging Receiver?

ecal/core/src/config/builder/logging_attribute_builder.cpp
```diff
@@ -6,19 +6,17 @@ namespace eCAL
...
     {
-      SAttributes attributes;
+      SProviderAttributes attributes;
 
-      attributes.network_enabled    = reg_config_.network_enabled;
```
KerstinKeller, line 13: you mentioned something about failing tests. could this also be a reason?

ecal/core/src/logging/config/attributes/ecal_log_provider_attributes.h
```diff
@@ -0,0 +1,74 @@
...
+      struct SUDP
+      {
+        bool                enabled;
+        eCAL_Logging_Filter filter_log;
+      };
+
+      struct SFile
+      {
+        bool                enabled;
+        std::string         path;
```
KerstinKeller, line 41: The path configures the File Sink. Adress, port etc configures the UDP Sink.
However you treat the attributes differently for the different sink types.

If you want to be consistent between for the configurations you could do something like:
```
struct SSink
{
  bool enabled;
  eCAL_Logging_Filter filter_log;
}
```
So to group the common attributes together, and separate the sink specific ones.

ecal/core/src/logging/config/attributes/ecal_log_receiver_attributes.h
```diff
@@ -0,0 +1,49 @@
...
+{
+  namespace Logging
+  {
+    struct SReceiverAttributes
+    {
+      struct SUDPReceiver
+      {
+        std::string address;
+        int         port;
+        bool        broadcast;
```
KerstinKeller, line 36: Does a receiver need a broadcast argument? I thought it's only relevant for sending infos.

KerstinKeller (commented):
I think the PR is in a good state. The only thing that I would change is the Config Object.
It feels wrong that the UDP sink has the property `receive`, I think I would rather split it.

The internal handling could possibly be changed later, also.

ecal/core/src/logging/config/attributes/ecal_log_receiver_attributes.h
Peguen: This is because of the eCAL::UDP::SReceiverAttr in io/udp/ , it expects a value also the broadcast value.

ecal/core/src/config/builder/logging_attribute_builder.cpp
Peguen: Ah no, this was a different reason. There was a type error in the yaml generation, fixed in the commit bcf6d54

ecal/core/include/ecal/config/logging.h
```diff
@@ -41,42 +41,49 @@ namespace eCAL
...
-        Console::Configuration console;
-        File::Configuration    file;
-        UDP::Configuration     udp;
+        Sink                console { true,  log_level_error | log_level_fatal}; //!< default: true, log_level_error
+        Sink                file    { false, log_level_none };                   //!< default: false, log_level_none
+        Sink                udp     { true,  log_filter_default };               //!< default: true, log_filter_default
+        
+        File::Configuration file_config;
+        UDP::Configuration  udp_config;
+
```
KerstinKeller, line 86: Looks very clean like this!

ecal/core/include/ecal/ecal_init.h
```diff
@@ -28,12 +28,12 @@ namespace eCAL
...
-    static const unsigned int Service    =    0x004;
-    static const unsigned int Monitoring =    0x008;
-    static const unsigned int Logging    =    0x010;
-    static const unsigned int TimeSync   =    0x020;
+    static const unsigned int Publisher     =    0x001;
+    static const unsigned int Subscriber    =    0x002;
+    static const unsigned int Service       =    0x004;
+    static const unsigned int Monitoring    =    0x008;
+    static const unsigned int Logging       =    0x010;
+    static const unsigned int TimeSync      =    0x020;
```
KerstinKeller, line 36: you could realign, so you don't get any changes in this file

ecal/core/src/config/configuration_to_yaml.h
```diff
@@ -298,19 +298,19 @@ namespace YAML
...
   */
   template<>
-  struct convert<eCAL::Logging::Sinks::UDP::Configuration>
+  struct convert<eCAL::Logging::Sinks::UDP::ReceiverConfiguration>
   {
-    static Node encode(const eCAL::Logging::Sinks::UDP::Configuration& config_);
+    static Node encode(const eCAL::Logging::Sinks::UDP::ReceiverConfiguration& config_);
 
-    static bool decode(const Node& node_, eCAL::Logging::Sinks::UDP::Configuration& config_);
+    static bool decode(const Node& node_, eCAL::Logging::Sinks::UDP::ReceiverConfiguration& config_);
```
KerstinKeller, line 305: It feels a bit strange that `ReceiverConfiguration` is in the `Sinks` namespace.
Does `eCAL::Logging::Receiver::UDP::Configuration` make more sense?

KerstinKeller (commented):
I just have a very few minor remarks, but I guess then finally ready to be merged 😄

KerstinKeller (approved):

Issue #1832, opened 2024-12-03
eCAL 6 API Review


KerstinKeller:
### Context

We need to review API for eCAL 6.

### Proposal

ecal_util.h
Remove:
- [ ] EnableLoopback (set by Config)
- [ ] PubShareType (set by Config)
- [ ] PubShareDescription  (set by Config)
- [ ] SplitCombinedTopicType
- [ ] CombinedTopicEncodingAndType

DownerCase:
> STopicId -> ID part should probably be private (e.g. not accessible by user)

Could we keep at least the `entity_id` part available? It's useful, for example, to have a registration callback save the entity ID and later use that ID to find the correct publisher/subscriber from the result of `GetMonitoring`. Basically, some way to uniquely identify eCAL entities is useful. (This applies to the server/clients too)

KerstinKeller:
Yes, the id should be there, and we would support e.g. comparing for equality, but I'd like to keep the actual ID private, as it's only a pseudo random number and has no meaning to the user.

DownerCase:
> Yes, the id should be there, and we would support e.g. comparing for equality, but I'd like to keep the actual ID private, as it's only a pseudo random number and has no meaning to the user.

Sounds perfect, thanks.

KerstinKeller:
eCAL 6 C++ API is now frozen until the eCAL 6 release.

PR #1834, opened 2024-12-04
[core] Remove PubShareType / PubShareDescription (set via publisher c…

KerstinKeller:
…onfiguration).

See #1832


rex-schilasky (approved):
LGTM

PR #1835, opened 2024-12-04
remove registration topic/service quality information logic

rex-schilasky:
### Description
The existing 'eCAL::Registration' API contains a variety of functions that provide quality attributes in addition to the DataType information. These quality attributes are also used to filter out information when multiple entities have the same topic or service name.

The eCAL6  Registration API uses identifiers to uniquely distinguish entities. Therefore, the name-based variants are to be removed.


ecal/core/include/ecal/msg/dynamic.h
```diff
@@ -186,7 +150,7 @@ namespace eCAL
      * @param clock_       Message writer clock.
      * @param id_          Message id.
      **/
-    using MsgReceiveCallbackT = std::function<void(const char* topic_name_, const T& msg_, long long time_, long long clock_, long long id_)>;
+    using MsgReceiveCallbackT = std::function<void(const Registration::STopicId& topic_id_, const SDataTypeInformation& topic_info_, const T& msg_, long long time_, long long clock_, long long id_)>;
```
KerstinKeller, line 153: @rex-schilasky  I am not sure if the user callbacks need to be provided the datatype info? The pure subscriber callbacks provide this info. However the dynamic receivers use the datatype information to create prototypes of T and populate them.

KerstinKeller: So, theoretically, we could leave the api as is... But maybe we can talk about it?

KerstinKeller (commented):
We get rid of so much code  😄 
The only thing that is bugging me, is the SDatatypeInformation in the Dynamic Callback. I think it's not necessary, since the information is "processed" within the datatypes which is returned. Maybe we can discuss it tomorrow.

ecal/core/include/ecal/msg/dynamic.h
rex-schilasky: Makes fully sense. The topic id is enough here.

KerstinKeller (approved):

PR #1846, opened 2024-12-06
service protocol 0 removed

rex-schilasky:
### Description
Service protocol v0 implementation removed (client and server side).

PR may conflict with #1842


ecal/core/src/service/ecal_service_server_impl.cpp
```diff
@@ -331,8 +312,8 @@ namespace eCAL
     service.uname       = Process::GetUnitName();
     service.sname       = m_service_name;
 
-    service.tcp_port_v0 = server_tcp_port_v0;
-    service.tcp_port_v1 = server_tcp_port_v1;
+    service.tcp_port_v0 = 0;
```
KerstinKeller, line 315: Is this the internal API?  should it be removed there, too?

ecal/service/ecal_service/src/client_session.cpp
```diff
@@ -76,19 +75,13 @@ namespace eCAL
     }
 
     ClientSession::ClientSession(const std::shared_ptr<asio::io_context>&                   io_context
-                                , std::uint8_t                                              protocol_version
+                                , std::uint8_t                                              /*protocol_version*/
```
KerstinKeller, line 78: doesn't it make sense to remove the variable from the constructor?

ecal/core/src/service/ecal_service_server_impl.cpp
rex-schilasky: This is part of the eCAL service registration sample. We could remove this from the registration sample as well but we have to ensure that it is communicated as 0.

KerstinKeller (commented):
In general very good that we can remove so much code!
Just see the two remarks that I made...

ecal/core/src/service/ecal_service_server_impl.cpp
KerstinKeller: Ok, maybe a separate PR for that, or at least we add it to the TODO list.

ecal/service/ecal_service/src/client_session.cpp
rex-schilasky: This is the service library. Not sure if one day this library will introduce a new protocol again. This "protocol_version" is exiting all over the library implementation ..

KerstinKeller (approved):

PR #1849, opened 2024-12-09
[core] redesign service  server api

rex-schilasky:
### Description
Implement a new service server API (symmetrically to new service client API).


Peguen (approved):
🛸 abgespaced

PR #1851, opened 2024-12-10
[config] Transport layer struct rework

Peguen:
### Description
Introduced MulticastConfiguration struct for local and network configuration.
Introduced for the local udp config the variable as function local static.


rex-schilasky (approved):
LGTM

PR #1853, opened 2024-12-11
[logging] API change for ecal logging

Peguen:
### Description
Removed setter and getter for loglevel. This means, changing of log level is only possible via configuration before initialization.
Removed eCAL::Logging::Log(const std::string&) signature. User now always has to specify the log level to write with.
Deprecated set_loglevel in Python bindings. Python will now always log in "log_level_all" until new logging and API will be introduced.


lang/python/core/ecal/core/core.py
```diff
@@ -148,7 +148,8 @@ def log_setlevel(level):
   :type level: int
 
   """
-  return _ecal.log_setlevel(level)
+  print("[WARNING] 'log_setlevel(int)' is deprecated. All logs will have the log level 'all' until the new API is introduced.")
```
rex-schilasky, line 151: Implement set log level function using global level state variable in ecal python c++ wrapper as intermediate solution.

rex-schilasky (changes requested):

rex-schilasky (approved):
Finally :-)

PR #1854, opened 2024-12-11
[config] API change for ecal_config.h

Peguen:
### Description
Removed the experimental namespace out of ecal_config.h.
Moved some config Get functions to internal space.


rex-schilasky (approved):
all config functions except GetConfiguration, GetPublisherConfiguration and GetSubscriberConfiguration need to be moved to ecal_config_internal.h as well

PR #1862, opened 2024-12-16
[core] Adapted c-api of monitoring in the way to remain c-return value compliant

hannemn:
This is a follow-up on #1859


Peguen (approved):
🍀

PR #1864, opened 2024-12-17
[clang] 2 checks removed

rex-schilasky:
### Description
`misc-include-cleaner` and `readability-avoid-const-params-in-decls` removed from clang-tidy checks


Peguen (approved):
🍘

PR #1872, opened 2025-01-02
[core] Change entity_id type from std::string to uint64_t

KerstinKeller:
Functional change: change map IDs from combined strings to entity_id only (as they are unique within the system)

The serialization format is kept as is for binary compatibility. E.g. while internally being handled as uint64_t, it is converted to a string on the wire, and parsed from the string upon deserialization.

Please review with regard to the following:
Name of EntityID type: `using EntityIdT = uint64_t;` or rather `TEntityID` (similar to `SEntityID` -> bit unhappy about naming here)


hannemn (approved):
lgtm

PR #1875, opened 2025-01-07
[core] get service id implementation

rex-schilasky:
### Description
Implements `GetServiceId` for `CServiceClient` and `CServiceServer`.

### DIFF
--- a/ecal/core/src/service/ecal_service_client_impl.cpp
+++ b/ecal/core/src/service/ecal_service_client_impl.cpp
@@ -368,6 +368,20 @@ namespace eCAL
     return GetRegistrationSample();
   }
 
+  Registration::SServiceMethodId CServiceClientImpl::GetServiceId() const
+  {
+    Registration::SServiceMethodId service_id;
+
+    service_id.service_id.entity_id = m_client_id;
+    service_id.service_id.entity_id = Process::GetProcessID();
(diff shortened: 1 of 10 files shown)


ecal/core/src/service/ecal_service_client_impl.cpp
```diff
@@ -368,6 +368,20 @@ namespace eCAL
...
+  Registration::SServiceMethodId CServiceClientImpl::GetServiceId() const
+  {
+    Registration::SServiceMethodId service_id;
+
+    service_id.service_id.entity_id  = m_client_id;
+    service_id.service_id.process_id = Process::GetProcessID();
+    service_id.service_id.host_name  = Process::GetHostName();
+
+    service_id.service_name = m_service_name;
+    service_id.method_name  = "";
```
Peguen, line 380: Why is the method_name an empty string?

rex-schilasky: The `method_name` field is used in other internal contexts (for example as key in `ServiceIdInfoMap`) but here the method name makes no sense. I will introduce a new SServiceId type additional to the existing SServiceMethodId for now. The new type will cover service and client instance id's and the other one will be used to get ID's of service-methods. Let me try .. ;-)

PR #1878, opened 2025-01-07
[mon_gui] Logging enable, crash fix

Peguen:
### Description
Enable logging receiving in mon_gui.
Check size of rows before adding them in log_model.cpp . Without that size check, in debug the monitor crashed through an assert.

### DIFF
--- a/app/mon/mon_gui/src/widgets/models/log_model.cpp
+++ b/app/mon/mon_gui/src/widgets/models/log_model.cpp
@@ -217,6 +217,8 @@ Qt::ItemFlags LogModel::flags(const QModelIndex &index) const
 void LogModel::insertLogs(const eCAL::pb::LogMessageList& logging_pb)
 {
   int inserted_row_count = logging_pb.log_messages().size();
+  if (inserted_row_count <= 0) return;
+  
   int size_before = logs_.size();
 
   // Remove entries from the top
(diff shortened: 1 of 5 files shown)


app/mon/mon_gui/src/widgets/models/log_model.cpp
```diff
@@ -263,7 +265,8 @@ void LogModel::insertLogs(const eCAL::pb::LogMessageList& logging_pb)
     counter--;
   }
 
-  endInsertRows();
+  if (insert_row)
+    endInsertRows();
```
KerstinKeller, line 269: doesn't it make sense to have "one big if" statement here?
Probably nothing is done if `insert_row == false`, however, it reads weirdly like this.

KerstinKeller: Mhmh but maybe this is a missunderstanding? It only handles the resizing of the display while we always need to copy over logs? I guess it's a bit hard to figure out what the code is doing.

FlorianReimold (approved):
LGTM 😃

app/mon/mon_gui/src/widgets/models/log_model.cpp
Peguen: Removed that again and checked the amount of messages directly at the beginning of the method and return at 0.

PR #1879, opened 2025-01-08
[mon_gui] Change font for new topics in mon_gui

albrechtdennis:
### Description
When new topics appear in the gui, the font is set to bolt.
A variable named "tree_item_counter" is assigned to the new topic, which increases by 2 every monitor update. 
The font stays bolt until the value of the variable exceeds 20, after which the font will return to normal.

### DIFF
--- a/app/mon/mon_gui/src/widgets/models/topic_tree_item.cpp
+++ b/app/mon/mon_gui/src/widgets/models/topic_tree_item.cpp
@@ -297,6 +297,37 @@ QVariant TopicTreeItem::data(Columns column, Qt::ItemDataRole role) const
 
   else if (role == Qt::ItemDataRole::FontRole)
   {
+
+    if ((column == Columns::HNAME)
+      || (column == Columns::HGNAME)
+      || (column == Columns::PNAME)
+      || (column == Columns::UNAME)
+      || (column == Columns::TNAME)
+      || (column == Columns::DIRECTION)
+      || (column == Columns::TENCODING)
+      || (column == Columns::TTYPE))
+    {
+      const QString raw_data = data(column, (Qt::ItemDataRole)ItemDataRoles::RawDataRole).toString(); //-V1016
+      if (raw_data.isEmpty())
+      {
+        QFont font;
+        font.setItalic(true);
+        font.setBold(itemfont.bold());
+        return font;
+      }
+    }
+    else if (column == Columns::TDESC)
+    {
+      const std::string& raw_data = topic_.tdatatype().desc();
+      if (raw_data.empty())
+      {
+        QFont font;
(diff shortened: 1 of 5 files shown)


app/mon/mon_gui/src/widgets/models/topic_tree_model.cpp
```diff
@@ -84,40 +85,65 @@ void TopicTreeModel::monitorUpdated(const eCAL::pb::Monitoring& monitoring_pb)
...
     }
     else
     {
+      auto& tree_entry = topic_tree_item_map_.at(topic_id);
+      auto topic_tree_item = tree_entry.tree_item;
+      if (tree_entry.tree_item_counter < 20)
+      {
+        tree_entry.tree_item_counter++;
+      }
+      else
```
Peguen, line 104: These statements can get united together.

app/mon/mon_gui/src/widgets/models/topic_tree_item.cpp
```diff
@@ -297,41 +297,18 @@ QVariant TopicTreeItem::data(Columns column, Qt::ItemDataRole role) const
...
-      || (column == Columns::TNAME)
-      || (column == Columns::DIRECTION)
-      || (column == Columns::TENCODING)
-      || (column == Columns::TTYPE))
-    {
-      const QString raw_data = data(column, (Qt::ItemDataRole)ItemDataRoles::RawDataRole).toString(); //-V1016
-      if (raw_data.isEmpty())
-      {
-        QFont font;
-        font.setItalic(true);
```
Peguen, line 314: Maybe don't remove the logic completely. Instead set set and unset the member font varible and return it here.

Peguen (changes requested):

app/mon/mon_gui/src/widgets/models/topic_tree_model.cpp
```diff
@@ -84,40 +85,65 @@ void TopicTreeModel::monitorUpdated(const eCAL::pb::Monitoring& monitoring_pb)
...
+      auto fontrole = topic_tree_item->data(1, Qt::ItemDataRole::FontRole);
+      auto font = qvariant_cast<QFont>(fontrole);
+      font.setBold(true);
+      topic_tree_item->setFont(font);
     }
     else
     {
+      auto& tree_entry = topic_tree_item_map_.at(topic_id);
+      auto topic_tree_item = tree_entry.tree_item;
+      if (tree_entry.tree_item_counter < 20)
```
Peguen, line 100: Maybe instead of item counter, you could make it independend of the calls with a time, as the user can specify how fast the monitoring app updates the content.

hannemn:
Handled in #1955

PR #1883, opened 2025-01-09
Use uint64_t for topic id in hdf5

KerstinKeller:
Changes the underlying type of the topic ID (there is no reason for IDs to be of a signed type).
Also add missing license headers.


Peguen (approved):
👍

PR #1885, opened 2025-01-10
[core] Completely remove the ability not to transmit type and descrip…

KerstinKeller:
…tor of a Publisher / Subscriber on the wire.

(Public API already removed previously)


Peguen (approved):
looking good 👍

PR #1887, opened 2025-01-12
[core] service client use datainfo (post fixes)

rex-schilasky:
### Description
Simplified

```cpp
using MethodInfoCallbackT = std::function<int(const SMethodInfo& method_info_, const std::string& request_, std::string& response_)>;
```
using new structure:

```cpp
struct SMethodInfo
{
  std::string              method_name; //!< The name of the method.
  SDataTypeInformation     req_type;    //!< The type of the method request.
  SDataTypeInformation     resp_type;   //!< The type of the method response.
};
```

Removed code duplicates for `SDataTypeInformation` conversion in `CDescGate`.

### DIFF
--- a/ecal/core/src/ecal_descgate.cpp
+++ b/ecal/core/src/ecal_descgate.cpp
@@ -178,8 +178,8 @@ namespace eCAL
     {
       for (const auto& method : sample_.service.methods)
       {
-        SDataTypeInformation request_datatype  = GetDataTypeInformation(method.req_datatype, method.req_type, method.req_desc);
-        SDataTypeInformation response_datatype = GetDataTypeInformation(method.resp_datatype, method.resp_type, method.resp_desc);
+        const SDataTypeInformation request_datatype  = GetDataTypeInformation(method.req_datatype, method.req_type, method.req_desc);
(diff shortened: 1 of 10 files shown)


ecal/core/src/ecal_descgate.cpp
```diff
@@ -178,32 +178,8 @@ namespace eCAL
...
-        {
-          response_datatype = method.resp_datatype;
-        }
-        // otherwise use the old resp_type and resp_desc fields
-        else
-        {
-          response_datatype.name       = method.resp_type;
-          response_datatype.descriptor = method.resp_desc;
-        }
+        const SDataTypeInformation request_datatype  = GetDataTypeInformation(method.req_datatype, method.req_type, method.req_desc);
```
KerstinKeller, line 181: Is this only an intermediate solution, until the datatype information is provided with every sample?
Maybe we should explicitly mark this in the code.

Also it would be faster if we could use the `const&` to the `method.req_datatype` to avoid copies if the data is already set. But I don't know how to do it with the function, because if `GetDatatypeInformation` returned `SDatatypeInformation&` you cannot return a reference to a temporary.

KerstinKeller (approved):

PR #1889, opened 2025-01-13
[config] Modernized config path handling

Peguen:
### Description
Path processing and API changes regarding configuration and log paths.

## API Changes
ecal_util.h
- Renamed:
  - GeteCALCOnfigPath() -> GeteCALDataDir()
  - GeteCALLogPath() -> GeteCALLogDir()
- Removed:
  - GeteCALActiveIniFile() -> use config object GetYamlFilePath() instead
  - GeteCALUserSettingsPath() (cfg path in data path)

## Logic Changes
- GeteCALConfigDir() -> Checking for available path in: env(ECAL_CONFIG_DIR), local user paths (appdata/local, ~/.ecal) or system paths (etc/ecal, ProgramData/eCAL) ==> removed the check for a config yaml file
- GeteCALLogDir() -> Same as GeteCALConfigDir(), but env(ECAL_LOG_DIR) first and as fallback taking system temp path or as last resort /ecal_tmp path
- Path returns will not simply create paths if they are not available. This now needs to be handled by the caller of the functions.

## Misc
- Splitted path handling in new header/source file (previous was in config_initializer)
- Using ecal-util functions for path processing
- Added a functional test for ecal-util
- Added tests for path_processing to ensure proper order of return values
- Enabled GMock for testing
- Log filter default in config/logging.h removed
- eCAL Time fix for reading time yaml (path separator)
- Output of path of the loaded config yaml file to std::cout


ecal/core/src/config/configuration_writer.cpp
```diff
@@ -0,0 +1,64 @@
...
+      }
+
+      file << ss_.str();
+      file.close();
+      return true;
+    }
+    
+    bool dumpConfigToFile(const eCAL::Configuration& config_, const std::string& file_path_)
+    {
+      const std::stringstream stream = getConfigAsYamlSS(config_);
```
KerstinKeller, line 49: Use stringstream from yaml generation.

ecal/core/src/config/ecal_config_initializer.cpp
```diff
@@ -285,133 +113,4 @@ namespace eCAL
...
-      return ini_file;
-    }
-
-    std::string GeteCALDefaultIniFile()
-    {
-      return GeteCALActiveIniFile();
-    }
-  
-  }
-}
```
KerstinKeller, line 417: Rename file `ecal_config.cpp`

KerstinKeller (approved):

PR #1891, opened 2025-01-13
[core] Get config before eCAL::Initialize through eCAL::Init::Configuration()

KerstinKeller:
- `eCAL::Init::Configuration()` to be called by user to receive an initialized (by yaml) configuration object, that they can alter
- `eCAL::GetConfiguration()` to return a `const Configuration&` so the user cannot unintentionally alter things.


Peguen (approved):
👍

PR #1892, opened 2025-01-13
[cmake] Use GNU install dirs variables instead of hardcoded value

DownerCase:
### Description

Removes hardcoded install destinations that aren't derived from the appropriate `CMAKE_INSTALL_<DIR>`, without this the special case `CMAKE_INSTALL_PREFIX` of `/` behaves incorrectly, with files split between (for example) `/usr/include` and `/include` 

Replaces the following uses with their `eCAL_install_<DIR>` equivalents:
- `INSTALL_LIB_DIR`
- `INSTALL_BIN_DIR`
- `INSTALL_INCLUDE_DIR`
- `INSTALL_CMAKE_DIR` (was unused)

Moved the `asioTargets.cmake` from `lib/CMake` to `lib/cmake` consistent with elsewhere.

Gets hdf5 submodule to use the GNU install dirs.

C#: Removed the unused `PUBLIC_HEADER` install and simplified the `install` command to implicitly use the GNU install dirs (if available -- which they are because I added the `include()` for them).

This should only really make any difference at all when `CMAKE_INSTALL_PREFIX` is exactly `/`; which admittedly is rare and unusual. Beyond that, it does mean all the install paths are now user controllable.

### Related issues

- 

### Cherry-pick to
- _none_


KerstinKeller (commented):
Thanks for making the PR!

The only concern I have, that by using the values in the ecal or lang/csharp subfolders, we will no longer be able to build these components on their own.
We could introduce `eCAL_core_xxx` directories, or use the `CMAKE_INSTALL_...` variables directly there.

Or, as I think that specifying the install directories is no longer required in CMake (I think they were before CMake 3.15, or so), maybe it could be deleted altogether.

DownerCase:
> Thanks for making the PR!
> 
> The only concern I have, that by using the values in the ecal or lang/csharp subfolders, we will no longer be able to build these components on their own. We could introduce `eCAL_core_xxx` directories, or use the `CMAKE_INSTALL_...` variables directly there.

Not sure what you mean w.r.t the ecal subfolder as eCAL core already uses the values (eg: https://github.com/eclipse-ecal/ecal/blob/3b22b6ee36c999dbcf5cdcae5a5d1f59593af249/ecal/core/src/ecal_defs.h.in#L37, https://github.com/eclipse-ecal/ecal-core/blob/e91a95c74fefc5e8d97eb19ed0788da022ed5e46/CMakeLists.txt#L154)

The csharp and python subfolders can't currently be built independently as they lack the requisite `find_package(eCAL)` calls (amongst other issues).

> Or, as I think that specifying the install directories is no longer required in CMake (I think they were before CMake 3.15, or so), maybe it could be deleted altogether.

That's pretty accurate. For `install(TARGETS ...)` [the docs note](https://cmake.org/cmake/help/v3.31/command/install.html#targets):

"_For regular executables, static libraries and shared libraries, the DESTINATION argument is not required.
For these target types, when DESTINATION is omitted, a default destination will be taken from the appropriate variable from [GNUInstallDirs](https://cmake.org/cmake/help/v3.31/module/GNUInstallDirs.html#module:GNUInstallDirs), or set to a built-in default value if that variable is not defined._"

For `install(DIRECTORY ...)` (i.e: old style method of install headers) "_Either a TYPE or a DESTINATION must be provided, but not both._" (and it will follow the same logic as above to pick a value)

DownerCase:
> Or, as I think that specifying the install directories is no longer required in CMake (I think they were before CMake 3.15, or so), maybe it could be deleted altogether.

I rebased and changed the C# folder to implicitly use the GNU install dir variables.

KerstinKeller (approved):

KerstinKeller:
Thanks for performing this cleanup!

PR #1909, opened 2025-01-16
[core]    Introduce config_file_path as new field in monitoring.

KerstinKeller:
[mon_gui] Display value in eCALMonGUI (hidden per default) 
Nanpb generator changed to 0.4.9.1

### Description
Introduce a new field `config_file_path` to process registration and monitoring, to be able to display / debug which configuration file has been loaded by eCAL.

It's shared via the registration, and can be accessed via the monitoring.

It can be displayed in eCALMonGUI, though the field is hidden by default.

image

Let's decide if it's worth adding this information.


Peguen (approved):
📎

I like it, nice feature.

PR #1915, opened 2025-01-17
[core] Change type of id in callbacks for server / client from SServiceMethodId to SServiceId.

KerstinKeller:
When registering events to see if a Server / Client has connected, the correct datatype to signal this is SServiceId.


Peguen (approved):
👍

rex-schilasky (approved):

PR #1916, opened 2025-01-17
[test: shm] Add testcase for fast reconnection

### LINKED ISSUE #1914: [eCAL 6] Linux - SHM Subscribers can appear connected but do not receive data
**Status:** Problem understood. When only a single subscriber for a topic is present, unsubscribing and resubscribing in less than the registration timeout period will result in the subscriber using a stale SHM file and be unable to receive new messages.

---

Extracted from https://github.com/eclipse-ecal/ecal/issues/1880#issuecomment-2596344274

### Problem Description

I found that with eCAL monitor applications when inspecting messages from topic A, switching to topic B, and then returning to topic A the application will no longer receive any messages, despite it indicating being connected.

### How to reproduce

- `cmake --preset cli`
- `cmake --build --preset cli`
- Run concurrently:
  - `./out/cli/build/bin/ecal_mon_tui`
  - `./out/cli/build/bin/ecal_sample_person_snd`
  - `./out/cli/build/bin/ecal_sample_minimal_snd`
- In the monitor:
  - Press enter to look at the messages from one topic - Works (messages visible after short delay)
  - Press down to switch to the other topic - Works (messages visible after short delay)
  - **After less time than the registration timeout** press up to go back to the original topic - Broken (messages never populate)
    - The monitor indicates 1 local connection for the topic but no data goes through

### How did you get eCAL?

Built from source

### Environment

- eCAL: Master 5484c2bc6821f0c4df7babf4d7b27803520c78c5
- OS: Linux x86_64 (Arch Linux)
- gcc 14.2.1

### eCAL System Information

```text
Default config
```

DownerCase:
### Description

Will eventually track down and resolve an issue with subscribers connecting but not receiving data.

Current status: Test case refined, problem understood (see issue for details).

### Related issues

- Fixes #1914

### DIFF
--- a/ecal/tests/cpp/pubsub_test/src/pubsub_test_shm.cpp
+++ b/ecal/tests/cpp/pubsub_test/src/pubsub_test_shm.cpp
@@ -216,7 +216,7 @@ TEST(core_cpp_pubsub, SubscriberFastReconnectionSHM) {
     eCAL::CSubscriber subscriber(TOPIC);
 
     const auto receive_lambda =
-        [&](const eCAL::Registration::STopicId & /*topic_id_*/,
+        [&](const eCAL::STopicId & /*topic_id_*/,
             const eCAL::SDataTypeInformation & /*data_type_info_*/,
             const eCAL::SReceiveCallbackData & /*data_*/) {
           {
--- a/ecal/tests/cpp/pubsub_test/src/pubsub_test_shm.cpp
+++ b/ecal/tests/cpp/pubsub_test/src/pubsub_test_shm.cpp
@@ -220,7 +220,6 @@ TEST(core_cpp_pubsub, SubscriberFastReconnectionSHM) {
             const eCAL::SDataTypeInformation & /*data_type_info_*/,
             const eCAL::SReceiveCallbackData & /*data_*/) {
           {
-            std::cerr << "Callback received message\n";
             std::lock_guard<std::mutex> lk(cv_m);
             data_received = true;
           }
@@ -236,7 +235,7 @@ TEST(core_cpp_pubsub, SubscriberFastReconnectionSHM) {
     cv.wait_for(cv_lk, RECEIVE_TIMEOUT,
                 [&data_received]() { return data_received; });
 
-    EXPECT_TRUE(data_received);
+    EXPECT_TRUE(data_received) << "No messages received within the timeout";
   };
 
   // 1 - Subscribe to topic and receive a message
--- a/ecal/tests/cpp/pubsub_test/src/pubsub_test_shm.cpp
+++ b/ecal/tests/cpp/pubsub_test/src/pubsub_test_shm.cpp
@@ -167,13 +167,20 @@ TEST(core_cpp_pubsub, SubscriberFastReconnectionSHM) {
 
   constexpr auto REGISTRATION_REFRESH_MS = 100;
   constexpr auto REGISTRATION_TIMEOUT_MS = 10 * REGISTRATION_REFRESH_MS;
+
+  constexpr auto MESSAGE_SEND_DELAY =
+      std::chrono::milliseconds(REGISTRATION_REFRESH_MS);
+
   // The delay between the destruction of the subscriber and its recreation
   // must be between the registration refresh and the registration timeout
   constexpr auto RESUBSCRIBE_DELAY =
-      std::chrono::milliseconds(2 * REGISTRATION_REFRESH_MS);
+      std::chrono::milliseconds(3 * REGISTRATION_REFRESH_MS);
+
   // Ensure we wait at least one timeout period
   constexpr auto RECEIVE_TIMEOUT =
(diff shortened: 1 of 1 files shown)


DownerCase:
Responding to [@hannemn's comment](https://github.com/eclipse-ecal/ecal/pull/1943#issuecomment-2612472145) from here:
> @KerstinKeller Yes, this definitely needs to be solved more cleanly, even if it is the safest solution for a short term release now. I still had an approach where the validity of the event handle is checked, but I like idea with reference counter that @DownerCase proposed even more. :-) However, this requires most likely a larger test loop.

I explored a few different options targeting either the unlinking of the shm file or interrupting the file observer. I was unable to find a solution for the later option that didn't involve adding something to `named_event_t`, at which point you totally break compatibility with things on eCAL v5.X

So, unless you changed the scheme used for the shm files since v5, trying to use a v6 subscriber with a v5 publisher will have the subscriber try and remove the shm file out from under the publisher as it'll see a 0 refcount. I don't know what your policy is w.r.t compatibility between v5 and v6 (I also haven't tested anything).

Regarding checking the validity of the event handle, that prevents you from doing `pthread_*_destroy` (not that it was being done previously anyway!) because at destructor time you can't know if the resource is in use elsewhere.

KerstinKeller:
Hi @DownerCase, thanks for digging so deep into it.
At the moment, we would like to avoid to break wire compatibility (that's what we call the possibility to communicate with eCAL over different eCAL versions).
And when we do break it, it would be great to put in some additional features, like Shared Locks for reading them SHM files, etc.

@hannemn also tried interrupting the Memfile Observer, this is another area where it might be worth refactoring / restructuring the code a bit.

The current solution, to go back to eCAL 5 behavior, and not to unlink any file until process destruction, also has the benefit that it will work seemlessly with eCAL 5.

However, it would be great, if we could just take your testcase and see if this now works with current master.

DownerCase:
> Hi @DownerCase, thanks for digging so deep into it. At the moment, we would like to avoid to break wire compatibility (that's what we call the possibility to communicate with eCAL over different eCAL versions). And when we do break it, it would be great to put in some additional features, like Shared Locks for reading them SHM files, etc.
> 
> However, it would be great, if we could just take your testcase and see if this now works with current master.

@KerstinKeller Totally understandable, I've rebased onto master and removed the ref counting changes. I believe the test will pass now.

ecal/tests/cpp/pubsub_test/src/pubsub_test_shm.cpp
```diff
@@ -149,3 +150,115 @@ TEST(core_cpp_pubsub, MultipleSendsSHM)
...
+
+  const auto create_subscriber = [&]() -> eCAL::CSubscriber {
+    eCAL::CSubscriber subscriber(TOPIC);
+
+    const auto receive_lambda =
+        [&](const eCAL::STopicId & /*topic_id_*/,
+            const eCAL::SDataTypeInformation & /*data_type_info_*/,
+            const eCAL::SReceiveCallbackData & /*data_*/) {
+          {
+            std::cerr << "Callback received message\n";
```
KerstinKeller, line 223: @DownerCase why are you printing to cerr here?

KerstinKeller (commented):
The testcase looks good, I am just wondering about printing to `std::cerr`. Does this show up in the test logs?

ecal/tests/cpp/pubsub_test/src/pubsub_test_shm.cpp
DownerCase: Force of habit mostly, cerr's unbuffered so don't need to pepper `std::endl/flush`.

Shows up in the logs all the same: https://github.com/eclipse-ecal/ecal/actions/runs/12952082976/job/36128611552#step:11:805

KerstinKeller: I am already not so happy with all the tests logging in the unittests (I know we also do have quite a bunch of tests that do it).

I'd prefer if we only see what's the problem on test failures, e.g. you could log to the google test command:
`EXPECT_TRUE(data_received) << "Missing data from publisher" ;`

DownerCase: Understood, I tried to keep it to a minimum but found the extra messages useful when diagnosing a failure (particularly when finding when internal eCAL debug logs are occurring relative to the test).

As an in-between what about logging via the eCAL logging functionality on the debug4 level (or similar)? So that normally no logs will be visible but they're there when you need to dig into it.

I'll add a descriptive error message to the `EXPECT_TRUE`

KerstinKeller: Hmhm good question. Not so sure using ecal logging for that usescase, feels a bit wrong.
Also you have to go through all the initialization phase for that to work correctly. So you couldn't use it for specific unit tests, etc. Maybe someone else has a better idea?

KerstinKeller (approved):

PR #1917, opened 2025-01-17
[config] Small hotfix for path_processing and logging to file

Peguen:
### Description
Logging to file expected a path separator at the end of the path. This now will be checked beforehand and handled accordingly.

GeteCALLogDirImpl used to check the local user and system path separately. Now its connected to the yaml file path. In case no LOG_VAR, DATA_VAR or config path is set, it tries to log to the folder where the ecal configuration yaml was loaded from.


ecal/core/src/config/ecal_path_processing.cpp
```diff
@@ -230,6 +231,30 @@ namespace eCAL
...
+      {
+        testFile.close();
+        std::remove(testFilePath.c_str());
+        return true;
+      } 
+      else 
+      {
+        return false;
+      }
+    }
```
KerstinKeller, line 249: I am not 100% sure if this is the best way to check if you have write access to a directory.
What happens if you can't delete the file anymore?
I would guess that there are system functions that can check write access to a directory.

KerstinKeller (approved):

PR #1920, opened 2025-01-20
Disable mon plugin and rec addon sdks by default unless building apps

### LINKED ISSUE #380: Guard CMake Include of
Mon and Rec extension points are not part of eCAL core.
They should not be build with the regular settings.

https://github.com/continental/ecal/blob/152145cf5507833cbfb03e5d366a587c2a405a2e/CMakeLists.txt#L315-L321

They should be guarded with either `-DBUILD_APPS` or other options: `BUILD_REC_SDK`, `BUILD_MON_SDK` ???

@FlorianReimold @hannemn

DownerCase:
### Description

Disables building the monitor plugin and rec addon unless `BUILD_APPS` is on.

These options are not part of eCAL core and aren't needed in a minimal configuration.

An option is made for the rec_addon but not the monitor plugin because the monitor app requires it. I presume it is unlikely one would want to use a local build to make a plugin but not build the gui app itself.

### Related issues

- Fixes #380

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,6 +88,7 @@ option(HAS_MESSAGEPACK                         "Platform supports messagepack li
 
 option(BUILD_DOCS                              "Build the eCAL documentation"                                     OFF)
 option(BUILD_APPS                              "Build the eCAL applications"                                       ON)
+option(BUILD_APP_REC_ADDON_SDK                 "Build eCAL rec addon SDK"                               ${BUILD_APPS})
 option(BUILD_SAMPLES                           "Build the eCAL samples"                                            ON)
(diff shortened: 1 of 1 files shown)


DownerCase:
> This might allow to remove the paths in pyproj.toml also:
> 
> https://github.com/eclipse-ecal/ecal/blob/6c9a238d396e274988d5a484110141ce5a32f695/pyproject.toml#L47-L48

Absolutely! Will check if the `app/apps` folder can be sorted too.

KerstinKeller:
We had previously used the app names in eCAL core, and thus always needed that folder. This dependency is no longer there, so it's fine to have that only when we have  `BUILD_APPS` activated.

KerstinKeller (approved):

PR #1923, opened 2025-01-20
[core] Rename Methods regarding Server/Client:

KerstinKeller:
GetServiceIDs -> GetServerIDs
GetServiceInfo -> GetServerInfo

### Description
We need to make the naming a bit more consistent: we haver Servers and Clients, for which we can retrieve information.


hannemn (approved):
👍

PR #1924, opened 2025-01-21
[cmake] add missing enable_testing call.

KerstinKeller:
### Description
Make sure C tests are executed in github actions.


Peguen (approved):
🧪

PR #1925, opened 2025-01-21
[Process] Unit name handling

Peguen:
### Description
Removed SetUnitName from API in C++, C and Python.
Changed UnitName generation in globals to use EcalUtils function.


KerstinKeller (approved):
Looks good!

PR #1930, opened 2025-01-22
eCAL core v5 compatibility modus (part 1)

KerstinKeller:
We want to introduce an eCAL 5 compatibility modus, which will make it easier to migrate to eCAL 6.

This PR is the first step for this.
There are now two targets, `ecal_core` and `ecal_core_v5`. 
They differentiate on how they define the eCAL namespaces:

`ecal_core`:
```
namespace eCAL
{
namespace v5
{
  // deprecated old function
  void old_function();
}
inline namesapce v6
{
  void new_function();
}
}
``` 

and 

`ecal_core_v5`:
```
namespace eCAL
{
inline namespace v5
{
  // deprecated old function
  void old_function();
}
namesapce v6
{
  void new_function();
}
}
``` 

In theory, when users migrate to eCAL 6, and enounter deleted functions, they can first use the compatibility mode.
Later they can switch back to `core`.
We will take the liberty to remove `core_v5` during the lifetime of eCAL 6 (e.g. API stability is promised only for the `core` Target and v6 namespace functions.

Also, the eCAL will use the namespace macros:
`ECAL_CORE_NAMESPACE_V5` and `ECAL_CORE_NAMESPACE_V6` for opening the namespaces.

v5 files / implementations have been moved to v5 include / src folders to allow easier cleanup at a later stage.


rex-schilasky (approved):
THANK YOU!

Issue #1912, opened 2025-01-17
eCAL Sys has broken behaviour for starting apps on different host


Peguen:
### Problem Description

I have a system of multiple hosts. I have a ecal sys configuration like in the picture

![Image](https://github.com/user-attachments/assets/068406eb-97cc-467a-8a69-6174f8f9d4ea)

When I start an application on a different host, e.g. here ecal_mma on ecal-vm-ubuntu-2404, it fails to start.

When I start the same application again, the application does start up. The problem is, next to the wanted application, all other ecal_mma applications on all hosts do also start up. This behaviour is reproducable.

The log of the executing eCAL Sys (on host Ubuntu 2204) client and gui is attached. Steps to generate these logs:
1. try to start ecal_mma on different host -> fail
2. try to start ecal_mma on same different host -> success with also other mmas started
3. turn off the previously started mma
4. turn off all started mmas

[2025-01-16-14-35-21_eCALSysClient_3633.log](https://github.com/user-attachments/files/18453068/2025-01-16-14-35-21_eCALSysClient_3633.log)

[2025-01-16-14-35-50_eCALSysGUI_3646.log](https://github.com/user-attachments/files/18453072/2025-01-16-14-35-50_eCALSysGUI_3646.log)

Sample communication, e.g. from ecal_sample_math_server works also, but also with these warnings:

![Image](https://github.com/user-attachments/assets/15266e1f-9ace-42dd-ac41-abce1667a323)

### How to reproduce

See above

### How did you get eCAL?

Local build from master

### Environment

Ubuntu LTS and latest VMs.

Peguen:
Hostname assignment check and a little wait time until services connected, see PRs, fixed those issues.

PR #1932, opened 2025-01-22
[v5::service_client] Filter eventcalls by hostname in ecal_service_client_v5_impl

Peguen:
### Description
see above

#1912


rex-schilasky (approved):
LGTM

PR #1935, opened 2025-01-23
[all] removed ecal_ prefix of all public header files

rex-schilasky:
### Description
Renaming of all public include files starting with the prefix `ecal_` to remove the prefix, as well as updating all corresponding references in the project. Legacy v5 headers were exluded.


KerstinKeller (approved):
yay 👍

PR #1937, opened 2025-01-23
[core] move SServiceResponse (old API) to v5/ecal_callback.h

KerstinKeller:
### Description
Move deprecated type used for v5 Services to v5/ecal_callback.h

Fix minimal_client sample as it used incorrect values.


rex-schilasky (approved):
👍

PR #1939, opened 2025-01-23
[core-c] removed ecal_ prefix of public API header files

rex-schilasky:
### Description
Removed the `ecal_` prefix of all C language header.


KerstinKeller (commented):
I think, if we're renaming them, we should think about what is consistent.
Personally, I would prefer a completely new include directory, like `ecal_c` and then `ecal_c/publisher.h` etc. 
This naming would emphasize that the c binding is a separate entity altogether.

rex-schilasky:
> I think, if we're renaming them, we should think about what is consistent. Personally, I would prefer a completely new include directory, like `ecal_c` and then `ecal_c/publisher.h` etc. This naming would emphasize that the c binding is a separate entity altogether.

Should I apply this as well?

KerstinKeller:
Let's hear what @FlorianReimold and @Peguen say...

rex-schilasky:
Made a second commit renaming the c language ecal folder to ecalc. Lets decide which way to go.

FlorianReimold:
I would prefer ecal_c (it separates ecal and c better and is more visible). Dropping the cimpl from the header files as well would be nice, too, so we have `ecal_c/publisher.h`

rex-schilasky:
> We decided `ecal_c` as a folder name would be good, because it's easier to spot / read.
> 
> If we do this, we should also
> 
> * remove "_cimpl"
> * match the .c files to the header file names (I know we didn't yet do it for C++, but this always drives me crazy, if .c and .h files are named differently
> * rename `ecalc.h` -> `ecal.h`, `ecalc_export.h` -> `export.h` and `ecalc_...` -> `...`
> * also think about if we want the same folder structure as for C++ (pubsub / client subfolder).
> 
> These are my remarks 😄

That makes fully sense!

lang/c/core/include/ecal_c/core.h
```diff
@@ -18,17 +18,17 @@
...
 
 /**
- * @file   ecal_core_cimpl.h
+ * @file   core.h
  * @brief  eCAL core function c interface
 **/
 
-#ifndef ecal_core_cimpl_h_included
-#define ecal_core_cimpl_h_included
+#ifndef core_h_included
```
KerstinKeller, line 25: Let's keep canonical names here (including e.g. the folder structure), to avoid name clashes with other libraries (e.g. `type_h_included` might be something different libraries might define...

lang/c/core/include/ecal_c/export.h
```diff
@@ -18,12 +18,12 @@
 */
 
 /**
- * @file   ecalc_export.h
+ * @file   export.h
```
KerstinKeller, line 21: in c++ we kept the folder structure here (or did we?), maybe it makes sense here, too. `@file ecal_c/export`

lang/c/core/include/ecal_c/core.h
KerstinKeller: alternatively, we could just do `#pragma once` where the compiler will do the right thing.
not sure how well defined this is for C compilers.

KerstinKeller (commented):
The file structure looks really good, but please see my other two comments.

lang/c/core/include/ecal_c/export.h
rex-schilasky: Next commit ..

lang/c/core/include/ecal_c/core.h
rex-schilasky: Added folder structure (next commit) ..

PR #1942, opened 2025-01-24
[sys_core] Wait for client_service.IsConnected() on remote connections

Peguen:
### Description
When starting a process in ecal_sys, the first call failed. The connection for the clients was not established. With this PR a waiting time for the client connection is introduced to make sure that the client connection is confirmed.

#1912


KerstinKeller (approved):
@Peguen thanks for fixing this 😄 
We could consider adding a comment in the code why it is changed, but in the long run, this needs to be converted to underlying v6 services.

hannemn (approved):
LGTM! 🚀

PR #1943, opened 2025-01-24
[core] Fixed topic resubscription via shared memory

hannemn:
Due to newly created event handles when resubscribing an existing topic, it is no longer possible to receive data via shared memory of the same process and topic on unix operating systems. This commit fixes the problem temporarily until an even more elegant solution is implemented.


KerstinKeller (approved):
Thank you @hannemn 
Let's create an issue to solve this on the subscriber side for eCAL 6.1.

PR #1944, opened 2025-01-24
[core] ported CMsgPublisher/Subscriber to v6

rex-schilasky:
### Description
CMsgPub/Sub porting to v6.

### DIFF
--- a/ecal/core/include/ecal/msg/publisher.h
+++ b/ecal/core/include/ecal/msg/publisher.h
@@ -53,7 +53,7 @@ namespace eCAL
      * @param data_type_info_  Topic data type information (encoding, type, descriptor).
      * @param config_          Optional configuration parameters.
     **/
-    CMsgPublisher(const std::string& topic_name_, const struct SDataTypeInformation& data_type_info_, const Publisher::Configuration& config_ = GetPublisherConfiguration()) : CPublisher(topic_name_, data_type_info_, config_)
+    explicit CMsgPublisher(const std::string& topic_name_, const SDataTypeInformation& data_type_info_, const Publisher::Configuration& config_ = GetPublisherConfiguration()) : CPublisher(topic_name_, data_type_info_, config_)
     {
     }
 
@@ -65,7 +65,7 @@ namespace eCAL
      * @param event_callback_  The publisher event callback funtion.
      * @param config_          Optional configuration parameters.
     **/
-    CMsgPublisher(const std::string& topic_name_, const struct SDataTypeInformation& data_type_info_, const PubEventCallbackT& event_callback_, const Publisher::Configuration& config_ = GetPublisherConfiguration()) : CPublisher(topic_name_, data_type_info_, event_callback_, config_)
+    explicit CMsgPublisher(const std::string& topic_name_, const SDataTypeInformation& data_type_info_, const PubEventCallbackT& event_callback_, const Publisher::Configuration& config_ = GetPublisherConfiguration()) : CPublisher(topic_name_, data_type_info_, event_callback_, config_)
     {
     }
 
@@ -133,7 +133,7 @@ namespace eCAL
 
   protected:
     // We cannot make it pure virtual, as it would break a bunch of implementations, who are not (yet) implementing this function
-    virtual struct SDataTypeInformation GetDataTypeInformation() const { return SDataTypeInformation{}; }
+    virtual SDataTypeInformation GetDataTypeInformation() const { return SDataTypeInformation{}; }
   private:
     virtual size_t GetSize(const T& msg_) const = 0;
     virtual bool Serialize(const T& msg_, char* buffer_, size_t size_) const = 0;
--- a/ecal/core/include/ecal/msg/subscriber.h
+++ b/ecal/core/include/ecal/msg/subscriber.h
@@ -56,12 +56,12 @@ namespace eCAL
     * @param topic_name_  Unique topic name.
     * @param config_      Optional configuration parameters.
     **/
-    CMessageSubscriber(const std::string& topic_name_, const Subscriber::Configuration& config_ = GetSubscriberConfiguration()) : CSubscriber(topic_name_, m_deserializer.GetDataTypeInformation(), config_)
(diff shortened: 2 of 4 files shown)


app/play/play_core/src/measurement_container.cpp
```diff
@@ -219,7 +215,8 @@ bool MeasurementContainer::PublishFrame(long long index)
         {
           timestamp_usecs = std::chrono::duration_cast<std::chrono::microseconds>(frame_table_[index].send_timestamp_.time_since_epoch()).count();
         }
-        frame_table_[index].publisher_info_->publisher_.SetID(frame_table_[index].send_id_);
+        // this is not supported by the eCAL v6 API
+        //frame_table_[index].publisher_info_->publisher_.SetID(frame_table_[index].send_id_);
```
KerstinKeller, line 219: Could this in any way be problematic?
So with eCAL 5, publishers will have the same ID they had when originally recording the measurement. In eCAL 6, they won't.

app/rec/rec_server_cli/src/ecal_rec_server_cli.cpp
```diff
@@ -144,6 +144,9 @@ std::unique_ptr<eCAL::rec_cli::command::Record> record_command;
 int main(int argc, char** argv)
 {
 #ifdef WIN32
+  (void)argc; // suppress unused warning
```
KerstinKeller, line 147: Why surpress it only for Windows? also maybe just comment in the function declaration `int main(int /*argc*/, char** /*argv*/)` (also this has nothing to do with this PR!!! 😏

contrib/ecaltime/simtime/src/ecal_time_simtime.h
```diff
@@ -113,7 +113,7 @@ namespace eCAL
     bool first_message_received;                /**< Whether we received at least one Message (used for the status message)*/
     eCAL::pb::SimTime::eState play_state;         /**< Current state (used for the status message)*/
 
-    eCAL::protobuf::CSubscriber<eCAL::pb::SimTime> sim_time_subscriber; /**< Subscriber for getting simulation timestamps */
+    std::unique_ptr<eCAL::protobuf::CSubscriber<eCAL::pb::SimTime>> sim_time_subscriber; /**< Subscriber for getting simulation timestamps */
```
KerstinKeller, line 116: I kind of don't like that now we're "forcing" code to use pointers.
Do you now have to check the pointers in other locations?

app/play/play_core/src/measurement_container.cpp
rex-schilasky: Its about the topic id for subscriber side filtering. This feature is rarely used and no way to support this in the future, it's removed on both sides (pub and sub). So in v5 I can filter out different "streams" of the the same topic on subscriber side in v6 the connection will just receive all messages and I need to implement an ID myself (for example as part of my protobuf description).

app/rec/rec_server_cli/src/ecal_rec_server_cli.cpp
rex-schilasky: For LINUX platform both arguments are used unfortunately and yes it's not really related but I tried to fix at least some warnings when I touched a file.

contrib/ecaltime/simtime/src/ecal_time_simtime.h
rex-schilasky: Without a Create/Destroy interface and the removed empty default constructors I see no alternative here.

rex-schilasky: We need to think about error handling for all communication entities (pub/sub/server/client) in case construction fails.

ecal/core/include/ecal/msg/publisher.h
```diff
@@ -68,13 +58,14 @@ namespace eCAL
...
      *
-     * @param topic_name_  Unique topic name.
-     * @param config_      Optional configuration parameters.
+     * @param topic_name_      Unique topic name.
+     * @param data_type_info_  Topic data type information (encoding, type, descriptor).
+     * @param event_callback_  The publisher event callback funtion.
+     * @param config_          Optional configuration parameters.
     **/
-    explicit CMsgPublisher(const std::string& topic_name_, const Publisher::Configuration& config_ = GetPublisherConfiguration()) : CMsgPublisher(topic_name_, GetDataTypeInformation(), config_)
+    CMsgPublisher(const std::string& topic_name_, const struct SDataTypeInformation& data_type_info_, const PubEventCallbackT& event_callback_, const Publisher::Configuration& config_ = GetPublisherConfiguration()) : CPublisher(topic_name_, data_type_info_, event_callback_, config_)
```
KerstinKeller, line 68: Why did you remove the `explicit` keyword?

KerstinKeller: Also why the `struct` keywords? They are not necessary and I would consider them bad practice in this case. (I know we have them in various places, but I already remove them when I do see them).

ecal/core/include/ecal/pubsub/publisher.h
```diff
@@ -70,7 +70,7 @@ namespace eCAL
        * @param config_          Optional configuration parameters.
       **/
       ECAL_API_EXPORTED_MEMBER
-        CPublisher(const std::string& topic_name_, const SDataTypeInformation& data_type_info_, const PubEventCallbackT event_callback_, const Publisher::Configuration& config_ = GetPublisherConfiguration());
+        CPublisher(const std::string& topic_name_, const SDataTypeInformation& data_type_info_, const PubEventCallbackT& event_callback_, const Publisher::Configuration& config_ = GetPublisherConfiguration());
```
KerstinKeller, line 73: I came across this, too, and I do wonder if it was on purpose, to copy the callbacks... But wasn't sure.

KerstinKeller (approved):
I just have the few minor remarks... but I guess it looks good!

contrib/ecaltime/simtime/src/ecal_time_simtime.h
KerstinKeller: Yes, but the thing is, construction is unlikely to fail. Failure is more likely to occur during the runtime. E.g. The connections (memory files, sockets, ...) are usually not created during construction, but when we receive monitoring information about other entities in the system.
So we have to find a way to deal with this effectively.
Maybe error/event callbacks would be interesting in that case, too.

PR #1945, opened 2025-01-24
[config] Rename host_group_name to shm_transport_domain

Peguen:
### Description
see above


ecal/tests/cpp/registration_test_public/src/registration_getservices.cpp
```diff
@@ -147,11 +147,11 @@ INSTANTIATE_TEST_SUITE_P(
       return config;
     }() },
     ServicesTestParams{ []() {
-      // shm + host group name
+      // shm + shm transport domain
       eCAL::Configuration config;
-      config.registration.layer.shm.enable = true;
```
KerstinKeller, line 152: Does it make sense to move it into layer.shm ?

KerstinKeller (approved):

PR #1946, opened 2025-01-24
[core] remove multiple structs from Registration namespace. (IDs for pubsub / services).

KerstinKeller:
A few structs (STopicId, SEntityId, SServiceId, ) are moved from `eCAL::Registration` namespace to only `eCAL` namespace.


hannemn (approved):
I hope I haven't overlooked anything. :-D

PR #1947, opened 2025-01-26
[cmake] Optional C bindings

DownerCase:
### Description

Adds a CMake option for including the C language bindings.
I know this might be contentious, as I've seen sentiment that you have too many CMake options already (personally, I think its ok and I hope you never look at VTK :laughing).

It feels like the C bindings are considered important, which is why I kept it default-on and marked as advanced to indicate it probably shouldn't be touched.  
The motivation is that, for example, I don't need it for my Go bindings so would like to be to exclude it from the container build (actually I see there's a `core_v5` target now which is a much larger unused resource for me...).

If the marked-as-advanced option is still too much, I ask at least for an undocumented regular variable.

### Related issues

-


KerstinKeller:
For me an additional option in this case is fine, now that the C bindings are properly split from the rest of eCAL core.

The extra C++ target was just a way to try and deal with all the API changes for eCAL 6 for our users. However in it's current state, it is not so much of a benefit.
We might even remove it again before the eCAL 6 release, or we can also introduce an option here.
Not 100% sure.

KerstinKeller (approved):

PR #1948, opened 2025-01-26
[cmake] Create explicit warnings target

DownerCase:
### Description

Creates a new target with the compiler warnings for eCAL and links it with everything created with the helper functions.

This is part 1 of a refactoring that will eventually remove workarounds like [`ecal_disable_all_warnings`](https://github.com/DownerCase/ecal/blob/dcd73635dfcaec79634ec009060b19bc77e11f86/cmake/helper_functions/ecal_helper_functions.cmake#L52C7-L52C32) arising from the dependencies inheriting the global [`CMAKE_CXX_FLAGS`](https://github.com/DownerCase/ecal/blob/dcd73635dfcaec79634ec009060b19bc77e11f86/CMakeLists.txt#L43) and similar. At the end the warnings (and other flags) for eCAL code may be changed in one location and be guaranteed that no dependencies will inherit them. Meanwhile `CMAKE_<LANG>_FLAGS` will remain untouched for user customization.

Part 2 will resolve the other libraries/executables that do not make use of the `ecal_add_*` helpers (eg: `measurement_hdf5`).  
If there's a reason to not be allowed to use the helper function(s) in any particular portion of the build tree, please say.

Part 3 will finally remove the warning flag configuration in the root CMakeLists (as everything eCAL will be getting it from the warnings target) and the workarounds present in the submodule build wrappers.

Future changes could look at things like increasing the warning level to catch things earlier or being able to enforce warnings as errors (for eCAL code); I certainly won't make any attempt unless desire is shown.

Detailed changes:
- Disabled the "sign-windows-installer" GitHub action step for windows when the repo is not the official eCAL repo.
  - (This step was causing the Windows workflow to always fail for forks, not helpful)
- Adds a new helper file `ecal_compiler_warnings.cmake`
  - This had to be added to installed CMake files because `ecal_add_functions.cmake` now uses it and is also installed :shrug: 
- Created an internal interface target `_ecal_warnings` to attach eCAL project warnings/flags to
- Had all the `ecal_add_*` helpers link `_ecal_warnings`
- Add `PRIVATE`/`PUBLIC` specifiers to _many many_ `target_link_libraries(...)` calls because you cannot mix signatures with and without access specifiers on a target.
  - Only the `*_pb` library targets got `PUBLIC` visibility, everything else (that didn't have specifiers) were plugins or executables.

### Related issues

-

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -194,6 +194,7 @@ jobs:
     # --------------------------------------------------------------------------------------------------
 
   sign-windows-installer:
+    if: github.repository == 'eclipse-ecal/ecal'
     runs-on: windows-2019
     needs: build-windows
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -491,6 +491,7 @@ install(FILES ${eCAL_config} ${eCAL_config_version}
 
 install(FILES
   cmake/helper_functions/ecal_add_functions.cmake
+  cmake/helper_functions/ecal_compiler_warnings.cmake
   cmake/helper_functions/ecal_helper_functions.cmake
   cmake/helper_functions/ecal_install_functions.cmake
   DESTINATION ${${PROJECT_NAME}_install_cmake_dir}/helper_functions
--- a/app/app_pb/CMakeLists.txt
+++ b/app/app_pb/CMakeLists.txt
@@ -68,7 +68,7 @@ target_compile_options(${PROJECT_NAME}
 
 set_property(TARGET ${PROJECT_NAME} PROPERTY POSITION_INDEPENDENT_CODE ON)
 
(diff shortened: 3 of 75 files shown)


KerstinKeller (approved):
This is a good and clean way to globally add the same flags to all of our targets.

KerstinKeller:
Thanks this is great! I will merge it directly.

As for the hdf5 / measurement targets. Just keep in mind, that in general, we aim for eCAL to be a bit more modular than it is today.
This includes being able to build (and maybe version (?)) parts of the repository on its own.
This is especially true for the measurement part, as it has no dependency on the core build of eCAL.

I think this could be the reason why no `ecal_add` functions were used there.

DownerCase:
Thanks, I'll keep that in mind going forwards

PR #1953, opened 2025-01-27
[core] service renaming and comments.

KerstinKeller:
### Description
A little bit of type renaming, and some comments added.


hannemn (approved):
LGTM

PR #1955, opened 2025-01-28
[mon_gui] Change font for new and deleted/disconnected topics in mon_gui

albrechtdennis:
### Description

**For new topics:**
When new topics appear in the gui, the font is set to bolt.
A variable named "tree_item_counter" is assigned to the new topic, which increases by 2 every monitor update.
The font stays bolt until the value of the variable exceeds 20, after which the font will return to normal.

**For deleted topics:**
When a topic is deleted or disconnected, its font changes to strikethrough before it disappears from the GUI.
After the font change, a QTimer starts running in the background, waiting for 10 seconds before finally removing the topic from the list.


hannemn:
@FlorianReimold Can you take a quick look over this PR?

app/mon/mon_gui/src/widgets/models/topic_tree_model.h
```diff
@@ -76,7 +76,7 @@ class TopicTreeModel : public GroupTreeModel
 
   QVariant headerData(int section, Qt::Orientation orientation, int role = Qt::DisplayRole) const override;
 
-  QVector<QPair<int, QString>> getTreeItemColumnNameMapping() const;
+  QVector<QPair<int, QVariant>> getTreeItemColumnNameMapping() const;
```
FlorianReimold, line 79: What is the reason for the QVariant?

FlorianReimold (commented):
- Switching the Group while one topic is striked through causes the main element to remain striked through

    - Set Monitor Group to "Nothing (Plain List)"
    - Start 2 x the same app (e.g. `ecal_sample_minimal_snd`)
    - Terminate one of them and wait for the strikethrough to show
    - Switch the group to "Topic"

    The strikethrough of the groupname now stays there forever (or until you change the grouping again)

    grafik

- A timeouted topic that re-appears takes a long time until it appears as new topic

    - Start an app (e.g. `ecal_sample_minimal_snd`)
    - Wait for the bold font to disappear in Monitor
    - Suspend the app (e.g. with Process Explorer / System Informer on Windows)
    - Wait for the strikethrough to appear
    - Resume the app

    The topic remains stiked through, even though it can be seen by the monitor again and the data clock is going up. After some time, the topic is removed from the monitor and immediately added as new topic in bold, again.

app/mon/mon_gui/src/widgets/models/topic_tree_item.cpp
```diff
@@ -322,16 +323,23 @@ QVariant TopicTreeItem::data(Columns column, Qt::ItemDataRole role) const
...
+    return itemfont;
   }
 
   return QVariant(); // Invalid QVariant
 }
 
+bool TopicTreeItem::setFont(const QFont& font)
+{
+  itemfont = font;
+  return false;
```
FlorianReimold, line 340: why does it always return false?

FlorianReimold (approved):

PR #1956, opened 2025-01-28
[API Review] Renaming of short struct(members)

Peguen:
### Description
s.a.


lang/python/core/ecal/core/core.py
```diff
@@ -321,15 +321,15 @@ def client_destroy(client_handle):
   return _ecal.client_destroy(client_handle)
 
 
-def client_set_hostname(client_handle, host_name):
```
KerstinKeller, line 324: We said we wouldn't change python api, so i would be a bit hesitant to make changes here.

lang/python/core/ecal/core/service.py
```diff
@@ -83,14 +83,14 @@ def destroy(self):
     """
     return ecal_core.client_destroy(self.shandle)
 
-  def set_hostname(self, host_name):
```
KerstinKeller, line 86: We said we wouldn't change python api, so i would be a bit hesitant to make changes here.

lang/python/core/src/ecal_wrap.cxx
```diff
@@ -942,61 +942,61 @@ namespace
       PyObject* topicDict = PyDict_New();
       PyList_Append(topic_list_, topicDict); Py_DECREF(topicDict);
 
-      PyObject* val = Py_BuildValue("i", topic.rclock);
```
KerstinKeller, line 945: We said we wouldn't change python api, so i would be a bit hesitant to make changes here.
Especially in the dictionary here.
I would leave it as is (`PyDict_SetItemString` calls), and properly name with nanopb binding.

KerstinKeller (commented):
I agree with the changes, except for the modified batch script and the python changes.
We should avoid them at this point and do them with nanopb.

hannemn (approved):

PR #1960, opened 2025-01-28
[core] remove ecal/callback.h

KerstinKeller:
### Description
The callback.h file contained only the redundant typedef for 
```
using TimerCallbackT = std::function<void()>;
``` 
which is also contained in `timer.h`.


Peguen (approved):
🗑️

PR #1961, opened 2025-01-28
[core] Limit frequency values on registration layer so they don't bec…

### LINKED ISSUE #1926: Frequency calculation seems to be wrong
### Problem Description

Wenn I start the sample "ecal_sample_latency_snd.exe", the calculation of the frequency seems off:

Image

I would not expect a negative frequency.

### How to reproduce

- see above

### How did you get eCAL?

I don't know

### Environment

eCAL master build on Windows

### eCAL System Information

```text

```

KerstinKeller:
…ome negative.

### Description
Limit frequencies to max values

### Related issues
Fixes #1926


Peguen (approved):
🛑

PR #1963, opened 2025-01-29
[core] Removed tlayer.h include and moved definitions to TransportLayer configuration

hannemn:


ecal/core/include/ecal/config/transport_layer.h
```diff
@@ -31,6 +31,15 @@ namespace eCAL
 {
   namespace TransportLayer
   {
+    enum eTransportLayer
```
Peguen, line 34: Does it make sense to change here to

`enum class
{
  none = 0,
  udp_mc = 1,
  ...
}`

?

FlorianReimold (approved):

PR #1969, opened 2025-01-29
[core] split `GetTopicNames` into `GetPublishedTopicNames` and `GetSubscribedTopicNames`.

KerstinKeller:
Please see if you like the naming.
To be more in line with the functions for service / client, the function was split into two separate functions.

### DIFF
--- a/ecal/core/src/registration/ecal_registration.cpp
+++ b/ecal/core/src/registration/ecal_registration.cpp
@@ -34,10 +34,11 @@ namespace eCAL
 {
   namespace Registration
   {
-    std::set<STopicId> GetPublisherIDs()
+    bool GetPublisherIDs(std::set<STopicId>& topic_ids_)
     {
-      if (g_descgate() == nullptr) return std::set<STopicId>();
-      return g_descgate()->GetPublisherIDs();
+      if (g_descgate() == nullptr) return false;
+      topic_ids_ = std::move(g_descgate()->GetPublisherIDs());
+      return true;
     }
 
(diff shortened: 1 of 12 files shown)


ecal/core/src/registration/ecal_registration.cpp
```diff
@@ -106,16 +106,23 @@ namespace eCAL
...
       const std::set<STopicId> pub_id_set = GetPublisherIDs();
       for (const auto& pub_id : pub_id_set)
       {
         topic_names_.insert(pub_id.topic_name);
       }
+    }
+
+    void GetSubscribedTopicNames(std::set<std::string>& topic_names_)
+    {
+      topic_names_.clear();
```
DownerCase, line 123: Why take these as ref if you're going to wipe the input anyway? Assuming the implementation will reuse the previously allocated memory?

```suggestion
    std::set<std::string> GetSubscribedTopicNames()
    {
      std::set<std::string> topic_names_{}
```

KerstinKeller: I'd be more than fine if they just returned the set. I guess we do have a few other functions which take a reference and then return `bool`.
Personally I don't like the style too much (I'd prefer something like `std::expected`, or `std::variant`, but since the core still uses C++ 14 that's out of the question 😞 

Anyways, I guess the biggest problem here is that we're not consistent. `ECAL_API std::set<SServiceId> GetClientIDs();` here we return a set, but `ECAL_API bool GetClientInfo(const SServiceId& id_, ServiceMethodInformationSetT& service_method_info_);` returns bool and takes a ref, and then lastly the functions to just get a name return  `void`.

PR #1975, opened 2025-01-30
[Topic] Remove remove user defined attributes for topics

Peguen:
### Description
Remove user defined attributes for topics. This includes also removing any functionality in the publisher/subscriber v5 implementation, where these interface calls still exist. They will now log a warning, that this has no impact anymore.

In v6 this already is phased out.

### DIFF
--- a/ecal/core_pb/src/ecal/core/pb/topic.proto
+++ b/ecal/core_pb/src/ecal/core/pb/topic.proto
@@ -50,8 +50,6 @@ message Topic                                      // eCAL topic
   int32               message_drops         = 18;  // dropped messages
                       
   int64               data_id               = 19;  // data send id (publisher setid)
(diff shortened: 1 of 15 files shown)


ecal/core_pb/src/ecal/core/pb/topic.proto
```diff
@@ -50,8 +50,6 @@ message Topic                                      // eCAL topic
   int32               message_drops         = 18;  // dropped messages
                       
   int64               data_id               = 19;  // data send id (publisher setid)
-  int64               data_clock                = 20;  // data clock (send / receive action)
-  int32               data_frequency                 = 21;  // data frequency (send / receive samples per second) [mHz]
-
-  map<string, string> attr                  = 27;  // generic topic description
```
hannemn, line 56: Please mark id 27 as reserved field.

hannemn (approved):

PR #1984, opened 2025-01-31
[core] introduced default timeout parameter for client service method calls

rex-schilasky:
### Description
As the title says. This will simplify the client method call API function usage a remarkable.


KerstinKeller (commented):
What about client_instance.h? We should have the same arguemnt ordering and default arguements there.

rex-schilasky:
> What about client_instance.h? We should have the same arguemnt ordering and default arguements there.

Adapted as well.

hannemn (approved):

PR #1989, opened 2025-01-31
[ecal_sys] Fix for startup bug when applications are not callable

Peguen:
### Description
When starting applications in ecal_sys, it happens that if an application is not available on the host, all other applications that are available on the host are shown as "failed". But they are proper started. You can find and close them through the cloud search.

The problem is, that all calls timeout when setting the timeout parameter too low. With this fix, the timeout is set to infinite, so the call will happen, while the calls to the missing application are proper recognized as "failed".

Increasing this timeout to 6-7 seconds also had the wanted effect.

This seems to be a fix for a symptom, not the cause. As this whole handling needs to be changed to the v6 implementation, it might be different then and for right now we will get the wanted behaviour with this fix.

### DIFF
--- a/app/sys/sys_core/src/connection/remote_connection.cpp
+++ b/app/sys/sys_core/src/connection/remote_connection.cpp
@@ -128,7 +128,14 @@ namespace eCAL
       std::lock_guard<decltype(connection_mutex_)> connection_lock(connection_mutex_);
 
       eCAL::v5::ServiceResponseVecT service_response_vec;
-      constexpr int timeout_ms = 1000;
+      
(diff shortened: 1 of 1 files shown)


Peguen:
In the current installation, we don't provide the python samples. That lead to an anavailable path in the ecal_sys test configuration. If a path is not available, the ecal_sys_client will run into

https://github.com/eclipse-ecal/ecal/blob/f24618fec28891a3eef37a7b2af5a3b08ff0ca3c/ecal/core/src/ecal_process.cpp#L1406

And wait for 5 seconds, which is more than the specified timeout.

PR #1997, opened 2025-02-04
[cmake]: Disable unwanted curl options and put missed targets in correct IDE folder

DownerCase:
### Description

Disables many curl targets from being generated and places missed thirdparty targets (just curl and yaml-cpp) into the correct IDE folder. (`CMAKE_FOLDER sets the default value for the `FOLDER` property)

### Related issues

- Visual studio CMake targets list being filled with junk from curl.


KerstinKeller (approved):

KerstinKeller:
Thank you @DownerCase

PR #2016, opened 2025-02-12
[cmake]: Fix usage after install when built with `BUILD_SHARED_LIBS=ON`

### LINKED ISSUE #2014: eCAL does not install all libraries it generates
### Problem Description

If I were to build and install separately, the `<install location>/lib` seems to be missing some libraries. This results in linker errors when trying to link my own applications. This was noticed in v5.13 and seems to be the case of v6.0.0-rc.1  

### How to reproduce

1. Configure and build eCAL as per the docs. I am on v6.0.0-rc.1 and use the following configuration options: 
    ```
    -DECAL_USE_HDF5=OFF -DECAL_USE_QT=OFF -DECAL_BUILD_APPS=OFF -DECAL_BUILD_SAMPLES=OFF \
    -DECAL_THIRDPARTY_BUILD_PROTOBUF=ON -DECAL_INSTALL_SAMPLE_SOURCES=OFF \
    -DECAL_CPACK_PACK_WITH_INNOSETUP=OFF -DCMAKE_PROJECT_TOP_LEVEL_INCLUDES=cmake/submodule_dependencies.cmake
   ```
2. `make install`
3. Notice that the following libraries are installed in `<install dir>/lib/`
    ```
    libecal_app_pb.so
    libecal_core_c.so
    libecal_core_pb.so
    libecal_core.so
    libecal_ecaltime_pb.so
    libecal_ecal-utils.a
    libecal_protobuf_base.so
    libecaltime-linuxptp.so
    libecaltime-localtime.so
    libecaltime-simtime.so
    libprotobuf-lite.so
    libprotobuf.so
    libprotoc.so
    ```
4. Notice that the following exist in `<build dir>/lib/`. **The last three libraries in the list are missing in the install location**
    ```
    libecal_app_pb.so
    libecal_core_c.so
    libecal_core_pb.so
    libecal_core.so
    libecal_ecaltime_pb.so
    libecal_protobuf_base.so
    libecaltime-linuxptp.so
    libecaltime-localtime.so
    libecaltime-simtime.so
    libprotobuf-lite.so
    libprotobuf.so
    libprotoc.so
    libtcp_pubsub.so
    libThreadingUtils.so
    libEcalParser.so
    ```
5. Create a simple eCAL application program and try to link against eCAL. I get the following linker errors
    ```
    /usr/bin/ld: warning: libtcp_pubsub.so.1, needed by ../../../../external/deploy/lib/libecal_core.so.6.0.0, not found (try using -rpath or -rpath-link)
    /usr/bin/ld: ../../../../external/deploy/lib/libecal_core.so.6.0.0: undefined reference to `tcp_pubsub::Publisher::~Publisher()'
    /usr/bin/ld: ../../../../external/deploy/lib/libecal_core.so.6.0.0: undefined reference to `tcp_pubsub::Subscriber::getSessions() const'
    /usr/bin/ld: ../../../../external/deploy/lib/libecal_core.so.6.0.0: undefined reference to `tcp_pubsub::SubscriberSession::getPort() const'
    /usr/bin/ld: ../../../../external/deploy/lib/libecal_core.so.6.0.0: undefined reference to `tcp_pubsub::Executor::Executor(unsigned long, std::__1::function<void (tcp_pubsub::logger::LogLevel, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)> const&)'
    /usr/bin/ld: ../../../../external/deploy/lib/libecal_core.so.6.0.0: undefined reference to `tcp_pubsub::Publisher::getPort() const'
    /usr/bin/ld: ../../../../external/deploy/lib/libecal_core.so.6.0.0: undefined reference to `tcp_pubsub::Subscriber::addSession(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, unsigned short, int)'
    /usr/bin/ld: ../../../../external/deploy/lib/libecal_core.so.6.0.0: undefined reference to `tcp_pubsub::SubscriberSession::getAddress() const'
    /usr/bin/ld: ../../../../external/deploy/lib/libecal_core.so.6.0.0: undefined reference to `tcp_pubsub::Publisher::Publisher(std::__1::shared_ptr<tcp_pubsub::Executor> const&, unsigned short)'
    /usr/bin/ld: ../../../../external/deploy/lib/libecal_core.so.6.0.0: undefined reference to `tcp_pubsub::Subscriber::Subscriber(std::__1::shared_ptr<tcp_pubsub::Executor> const&)'
    /usr/bin/ld: ../../../../external/deploy/lib/libecal_core.so.6.0.0: undefined reference to `tcp_pubsub::Subscriber::setCallback(std::__1::function<void (tcp_pubsub::CallbackData const&)> const&, bool)'
    /usr/bin/ld: ../../../../external/deploy/lib/libecal_core.so.6.0.0: undefined reference to `tcp_pubsub::Publisher::send(std::__1::vector<std::__1::pair<char const* const, unsigned long const>, std::__1::allocator<std::__1::pair<char const* const, unsigned long const> > > const&) const'
    /usr/bin/ld: ../../../../external/deploy/lib/libecal_core.so.6.0.0: undefined reference to `tcp_pubsub::Executor::~Executor()'
    /usr/bin/ld: ../../../../external/deploy/lib/libecal_core.so.6.0.0: undefined reference to `tcp_pubsub::Subscriber::~Subscriber()'
    clang++: error: linker command failed with exit code 1 (use -v to see invocation)
   ````
    

### How did you get eCAL?

### Environment

- eCAL version: 6.0.0-rc.1
- OS: Ubuntu 24.04
- CMake: 3.31.5
- Compiler: `Ubuntu clang version 21.0.0 (++20250210081934+67b7a2590f39-1~exp1~20250210082102.718)`

### eCAL System Information

```text

```

DownerCase:
### Description

Fixes the build such that the installed result is usable when eCAL is built with `BUILD_SHARED_LIBS=ON`.

- `tcp_pubsub` submodule is now always built statically as it is a strictly internal dependency.
- `ecal-utils` may now be static or shared depending on `BUILD_SHARED_LIBS`
  - It was required to be installed as it was used by other static/shared libraries
  - There was no particular reason for it to always be static so now it aligns with the other non-fixed libraries
- `EcalParser, QEcalParser, CustomQt, and ThreadingUtils` are now **always static** because they were implementation details for applications and application plugins and were not required to be installed.

### Related issues

Fixes #2014

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -377,6 +377,7 @@ jobs:
       working-directory: ${{ runner.workspace }}/_build/complete/_deploy
 
     - name: Determine name of the installer for zip file
+      if: env.IS_DOWNLOAD_AVAILABLE == 'true'
       shell: powershell
       run: |
         $INSTALLER_NAME = "${{ env.ASSET_NAME }}"
--- a/lib/CustomQt/CMakeLists.txt
+++ b/lib/CustomQt/CMakeLists.txt
@@ -66,7 +66,9 @@ set(custom_qt_src
   src/QToolTipMenu.cpp
 )
 
-add_library (${PROJECT_NAME}
+# CustomQt is only used for the implementation of the GUI applications and
+# monitor plugins so it can always be static and never installed/exported
+ecal_add_static_library(${PROJECT_NAME}
   ${custom_qt_includes}
   ${custom_qt_src}
 )
@@ -75,7 +77,7 @@ target_include_directories(${PROJECT_NAME}
   PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/src
   PUBLIC  ${CMAKE_CURRENT_SOURCE_DIR}/include)
 
-target_link_libraries(${PROJECT_NAME}
+target_link_libraries(${PROJECT_NAME} PUBLIC
     Qt${QT_VERSION_MAJOR}::Core
     Qt${QT_VERSION_MAJOR}::Widgets
 )
--- a/lib/EcalParser/CMakeLists.txt
+++ b/lib/EcalParser/CMakeLists.txt
@@ -42,7 +42,9 @@ set(sources
     src/functions/username.h
 )
 
-add_library (${PROJECT_NAME} 
+# QEcalParser is an internal implementation detail for applications so it can
+# always be static and not installed/exported
+ecal_add_static_library(${PROJECT_NAME}
     ${includes}
     ${sources}
 )
--- a/lib/QEcalParser/CMakeLists.txt
+++ b/lib/QEcalParser/CMakeLists.txt
@@ -67,7 +67,9 @@ else()
 endif()
 
 
-add_library (${PROJECT_NAME}
+# QEcalParser is an internal implementation detail for applications so it can
+# always be static and not installed/exported
+ecal_add_static_library(${PROJECT_NAME}
     ${qecalparser_includes}
     ${qecalparser_src}
     ${qecalparser_ui}
--- a/lib/ThreadingUtils/CMakeLists.txt
+++ b/lib/ThreadingUtils/CMakeLists.txt
@@ -33,7 +33,10 @@ set(threading_utils_src
   src/dummy.cpp # Workaround to make this project generate an output. Header only librarys do not work well with Visual Studio
 )
 
-add_library (${PROJECT_NAME} 
+# WARNING: This is only used for the play and recorder applications (CLI and GUI)
+# so it is never needed by the SDK portion of the codebase and thus can be:
+# Internal implementation - not exported/installed
+ecal_add_static_library(${PROJECT_NAME}
   ${threading_utils_includes}
   ${threading_utils_src}
 )
--- a/lib/ecal_utils/CMakeLists.txt
+++ b/lib/ecal_utils/CMakeLists.txt
@@ -38,7 +38,7 @@ set(ecal_utils_src
     src/win_cp_changer.cpp
 )
 
-ecal_add_static_library(${PROJECT_NAME}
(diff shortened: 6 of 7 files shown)


FlorianReimold (approved):

FlorianReimold (changes requested):
I am not happy with hiding the Windows signing issues. It's not a mandatory action and we can merge with the sign job failing, but I would rather have it notify me of failures than report OK despite not working.

DownerCase:
> I am not happy with hiding the Windows signing issues. It's not a mandatory action and we can merge with the sign job failing, but I would rather have it notify me of failures than report OK despite not working.
>
> Edit: or is it just about (not) signing from forks? I am struggeling with the script.

Yeah, the change should only affect forks.

The problem is my runs don't have access to your secrets so the `JENKINS_*` environment variables don't get set, which causes the "Sign the installer on Eclipse CI" step to be skipped (because otherwise it would fail from missing auth).
The signing step sets `IS_DOWNLOAD_AVAILABLE` to true and populates `ASSET_NAME` environment variables.

The other option is to do something with GitHub environments and pull_request_target: https://github.com/orgs/community/discussions/50161#discussioncomment-5824158

Alternatively, if you don't like the deployment spam you'd get with that approach, could do something directly with `pull_request_target` and change your repo settings such that checks always require approval _before_ running (like what should happen for first-time contributors).

---

From what I can tell, the workflow will correctly fail if something other than the secrets being unavailable goes wrong, and when the signing is skipped only the "unsigned-setup" artifact will be uploaded.

Seems fine to me, but if you want me to drop the change just say and I'll open an issue to capture this.

FlorianReimold (approved):

PR #2029, opened 2025-02-14
[samples] Latency rec sample fix

Peguen:
### Description
The latency_rec benchmark sample did not start anymore, as the behaviour of eCAL::Ok() changed, so that before an Initialize() eCAL::Ok() will return false (before it returned true).

Now the eCAL::Ok() gets properly called after Initialize(). In addition, the loop of finalize/initialize got removed. The sample will just create everytime a new subscriber.


rex-schilasky (approved):
👍

PR #2030, opened 2025-02-14
[core] move proto message header

rex-schilasky:
### Description

See title ..

`msg/proto/message.h` -> `msg/protobuf/common_message.h`


KerstinKeller:
We will address this differently.

PR #2043, opened 2025-02-21
[cmake] CMake 4.0 compatibility

### LINKED ISSUE #2041: CMake 4.0 Removes compatibility with CMake < 3.5
### Problem Description

CMake 4.0, releasing soon, (currently in RC1) makes good on CMake's promise that compatibility with versions older than 3.5 is deprecated and will be removed. With CMake 4.0 any projects that do not advertise compatibility with version 3.5 or newer will **hard fail**.

Specifically this means any lines of CMake of the form `cmake_minimum_required(VERSION <y>)` or `cmake_minimum_required(VERSION <x>...<y>)` where `y` is less than 3.5 will fail with the following error:

```text
Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
```

To alleviate the problem as projects updates `CMAKE_POLICY_VERSION_MINIMUM` is a new variable that forces CMake to override all `cmake_minimum_required` to be at least the value of `CMAKE_POLICY_VERSION_MINIMUM`.
This should not be set in project CMake code **unless** it is used immediately before an `add_subdirectory` for a third-party project. In all other cases it should be passed in externally.

---

To solve this all `cmake_minimum_required` not specifying support of at least CMake 3.5 must be updated.

I propose using `cmake_minimum_required(<min>..4.0)` (assuming no policies break anything) which should give the project 6 years to increase the max version again according to the new policy transition schedule: https://cmake.org/cmake/help/v4.0/manual/cmake-policies.7.html#transition-schedule

_I volunteer myself_ to work on this because once my distro updates CMake it will become an immediate annoyance (even if trivially solved). Plus it will get rid of the deprecation notices in my console :smile: 

--- 
Links:
4.0RC1 announcement post: https://discourse.cmake.org/t/cmake-4-0-0-rc1-is-ready-for-testing/13606
Download page: https://cmake.org/download/
Removal release note: https://cmake.org/cmake/help/latest/release/4.0.html#deprecated-and-removed-features
`CMAKE_POLICY_VERSION_MINIMUM`: https://cmake.org/cmake/help/v4.0/variable/CMAKE_POLICY_VERSION_MINIMUM.html
CMake Policies: https://cmake.org/cmake/help/v4.0/manual/cmake-policies.7.html

### How to reproduce

1. Obtain a version of CMake 4 (eg: Download the RC from their page: https://cmake.org/download/)
2. Try and run a build with it
3. Immediate failure stating `Compatibility with CMake < 3.5 has been removed from CMake.`

### How did you get eCAL?

I don't know

### Environment

Any CMake 4.0 or newer
All platforms

### eCAL System Information

_No response_

DownerCase:
### Description

Address instances where the project will fail with CMake 4.0

yaml-cpp and termcolor don't have tagged versions for the commits I updated to but you can see the changes are minimal (literally 1 commit forwards for yaml-cpp).

Possible future changes:
- Update samples to include a maximum version? They currently require 3.10 with no max specified, so don't _currently_ issue any warnings but feasibly could on the next update...
- Unify the project build to hard-require a fairly modern version of CMake and use the new language features.
- Push the max versions as far as possible (likely 4.0) and cleanup redundant instances of `cmake_minimum_required`

### Related issues

Fixes #2041 

### Cherry-pick to
None

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -18,6 +18,12 @@ jobs:
       run: |
           echo Killing XProtect.; sudo pkill -9 XProtect >/dev/null || true;
     
+      # TODO: Remove once CMake 4 is included in the runner image
+      # TODO: Waiting for 4.0.0-rc1 to be added...
+    - uses: lukka/get-cmake@5f6e04f5267c8133f1273bf2103583fc72c46b17 #3.31.5
+      with:
+        cmakeVersion: 4.0.0-rc1
+
     - name: Install latest Python3 via brew
       run: brew install python3 || brew link --overwrite python@3.12
 
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -19,10 +19,9 @@ jobs:
           echo Killing XProtect.; sudo pkill -9 XProtect >/dev/null || true;
     
       # TODO: Remove once CMake 4 is included in the runner image
-      # TODO: Waiting for 4.0.0-rc1 to be added...
-    - uses: lukka/get-cmake@5f6e04f5267c8133f1273bf2103583fc72c46b17 #3.31.5
+    - uses: lukka/get-cmake@28983e0d3955dba2bb0a6810caae0c6cf268ec0c # 4.0.0
       with:
-        cmakeVersion: 4.0.0-rc1
+        cmakeVersion: 4.0.0
 
     - name: Install latest Python3 via brew
       run: brew install python3 || brew link --overwrite python@3.12
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -21,6 +21,12 @@ jobs:
       
     steps:
 
+      # TODO: Remove once CMake 4 is included in the runner image
+      # TODO: Waiting for 4.0.0-rc1 to be added...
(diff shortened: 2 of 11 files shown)


KerstinKeller:
Thanks for doing the changes.
- Two remarks: there is no need to actually upgrade to 4.0.0 for master. We can test if it works, but leave the actions as they are for now.
- The doc samples will be removed in the upcoming days. We can now make these changes, but they will be obsolete.

We will make a new PR and add you as co-author on the commit.
For the submodules we will for now stay with the tags, and apply the same fix you applied for protobuf.

DownerCase:
> Thanks for doing the changes.
> 
> * Two remarks: there is no need to actually upgrade to 4.0.0 for master. We can test if it works, but leave the actions as they are for now.
> * The doc samples will be removed in the upcoming days. We can now make these changes, but they will be obsolete.
> 
> We will make a new PR and add you as co-author on the commit. For the submodules we will for now stay with the tags, and apply the same fix you applied for protobuf.

Sure thing! No problem with that.

DownerCase:
Reimplemented in #2103

PR #2046, opened 2025-02-25
Feature/msg api improvements

KerstinKeller:
### Description
Rework the general `CMessagePublisher` / `CMessageSubscriber`.
Both classes no longer inherit from `CPublisher` / `CSubscriber`

All serialization frameworks need to implement a generic `Serializer` class, which then allows them to easily support all of eCALs framework.

String:
- String classes are no longer `templatized`.
- Datatype information for String publishers are changed. `type=string`, `encoding=utf8`.

Flatbuffers:
-  API is now typesafe and more eCAL like. Support different classes depending on the flatbuffer type:
`CObjectPublisher`, `CObjectSubscriber` and `CFlatSubscriber`

Protobuf:
- Remove DynamicPublisher (Questionable usage).

In a following PR, these Serializer classes will be used to support Measurement read / write access.


serialization/capnproto/capnproto/include/ecal/msg/capnproto/publisher.h
```diff
@@ -24,155 +24,32 @@
...
   namespace capnproto
   {
     /**
-    * @brief eCAL capnp publisher class.
-    *
-    * Publisher template  class for capnp messages. For details see documentation of CPublisher class.
-    *
+     * @brief  eCAL capnp subscriber class.
+     *
+     * Subscriber template  class for capnp messages. For details see documentation of CSubscriber class.
```
hannemn, line 47: Typo: Publisher instead of Subscriber

hannemn (approved):

PR #2047, opened 2025-02-26
[core] Changed return types of entity class getters to const references.

hannemn:


hannemn:
I wille create another pull request that handles these adaptions.

PR #2055, opened 2025-02-28
[core] Removed reference counting for eCAL::Initialize()

hannemn:


KerstinKeller (commented):
I am a bit afraid of the changes, and I wonder how we can test in broader scope.
Should we als remove multiple `Initalize()` / `Finalize()` calls in the different apps, or do that in a separate PR?

hannemn:
There is no app or sample in eCAL which is calling Initialize()/Finalize() multiple times where reference counter necessarily needs to be taken into account. Only a simple Initialize/Finalize test had to be adapted, but this was already part of this pull request.

Peguen (approved):
👍

PR #2071, opened 2025-03-07
[core] Adapted API in order to return const references if possible

hannemn:


ecal/core/src/pubsub/ecal_subscriber.cpp
```diff
@@ -29,6 +29,8 @@
 #include "config/builder/reader_attribute_builder.h"
 #include "ecal_config_internal.h"
 
+#include <iostream>
```
hannemn, line 32: remove include

PR #2073, opened 2025-03-10
[core] Removed dead API function Process::GetTaskParameter() and changed argument type of Process::SetState() to std::string

hannemn:


ecal/core/src/ecal_process.cpp
```diff
@@ -289,11 +278,11 @@ namespace eCAL
       #endif
     }
 
-    void SetState(eCAL::Process::eSeverity severity_, eCAL::Process::eSeverityLevel level_, const char* info_)
+    void SetState(eCAL::Process::eSeverity severity_, eCAL::Process::eSeverityLevel level_, const std::string& info_ /*= "" */)
     {
       g_process_severity = severity_;
       g_process_severity_level = level_;
-      if (info_ != nullptr)
+      if (!info_.empty())
```
KerstinKeller, line 285: I am not a fan of the default parameter here.
Also, you should remove this check! and always assign info.

Because now you can missuse the function:

`SetState(error, level1, "I feel sick");`
`SetState(healthy, level1);`

Now, The state would be `healhty` but the message still `I feel sick`...

hannemn: You're right. This can lead to inconsitent combination when setting only the enums. I will threat the info parameter as mandatory.

KerstinKeller (approved):

PR #2074, opened 2025-03-11
[lang] New C API for eCAL 6

hannemn:


lang/c/core/src/time.cpp
```diff
@@ -70,23 +67,25 @@ extern "C"
...
   {
-    if (max_len_ == ECAL_ALLOCATE_4ME || max_len_ > 0)
+    assert(error_ != NULL);
+
+    std::unique_ptr<std::string> status_message;
+    if (status_message_ != NULL)
     {
-      std::string status_message;
-      eCAL::Time::GetStatus(*error_, &status_message);
+      if (*status_message_ != NULL) return;
```
Peguen, line 77: Here it should be an assert

Peguen (approved):
😱

PR #2076, opened 2025-03-12
[doc] Collect doc samples according to "repo standard" and enable build

Peguen:
### Description
Previously, sample sources in the documentation were stored in a relative src file to the document. This is now changed in order to maintain an equivalent directory structure like in the other components of eCAL.

In addition, previously in integration tests it was noticed, that the sample sources could not be build. Now there is the option to enable their build with ECAL_BUILD_DOC_SAMPLES turned to "ON", which is now the default setting in the documentation workflow.


doc/rst/getting_started/hello_world.rst
```diff
@@ -53,7 +53,7 @@ Somewhere on your hard drive create an empty directory and create a file :file:`
 
 * |fa-file-alt| :file:`CMakeLists.txt`:
 
-  .. literalinclude:: src/hello_world/hello_world_snd/CMakeLists.txt
+  .. literalinclude:: ../source_code_samples/cpp/string/hello_world_snd/CMakeLists.txt
```
KerstinKeller, line 56: Is it possible to provide "absolute" path, e.g. absolute to doc path, in case we move the hello_world.rst file to a different folder?

doc/samples/cpp/configuration/hello_config/main.cpp
```diff
@@ -0,0 +1,57 @@
...
+  eCAL::Configuration custom_config;
+
+  // Use the ecal.yaml file of the system, local data or current folder if available
+  custom_config.InitFromConfig();
+
+  // In case you decide to specify an own .yaml file to use.
+  // Configuration based on previous ini file will be overwritten.
+  // The new Configuration will use default values if the file is not found
+  // in the default directories (ECAL_DATA, cwd, %LOCALAPPDATA%/eCAL, $HOME/.ecal)
+  custom_config.InitFromFile("./ecal_custom.yaml");
```
KerstinKeller, line 18: I would rather show the default way to get the configuration, which would be `auto config = eCAL::Init::GetConfiguration()`.
We could still mention that you can init from a file, but the first option would be the preferred way to retrieve the config.

KerstinKeller (commented):
Looks good, but please see my comments...

doc/samples/cpp/configuration/hello_config/main.cpp
Peguen: Lets mention that in the documentation

doc/rst/getting_started/hello_world.rst
FlorianReimold: Is that path still correct? Shouldn't it be `/../samples/cpp/string/hello_world_snd/CMakeLists.txt`? Or am I missing something?

Peguen: The build happens with the copied src files in the _sources subfolder in the build. Now when we speficy /source_code_samples/... it takes the _sources subfolder as root and with that its not relative to the *.rst file anymore.

KerstinKeller (approved):

FlorianReimold:
The sample code has changed, so now the documentation around it also has to change.

What is the reason why we don't need the `set(CMAKE_FIND_PACKAGE_PREFER_CONFIG ON)` anymore?

Peguen:
Docu adaptions will be done in a separate PR I am working on.

Regarding the CMAKE_FIND_PACKAGE_PREFER_CONFIG refer to https://github.com/eclipse-ecal/ecal/pull/2009 .

PR #2092, opened 2025-03-26
[bugfix] Allow to store empty strings for DataTypeInformation in measurement

KerstinKeller:
Currently, when the user stores empty strings for any field in DataTypeInformation in the measurement, they are read back from hdf5 as "\0". 
This is problematic, as it does not allow to properly compare the DatatypeInformation for usecases like determining if serialization will work successfully.
This PR fixes this behavior.


KerstinKeller:
We could consider to cherry pick this manually to a possible 5.14, but not neccessarily.

Peguen (approved):
💡 
lgtm

PR #2095, opened 2025-03-28
[dependencies] Add support for ftxui 6.0.0.

KerstinKeller:
ftxui has released a new major version.
This branch tries to ensure compatibility with the new release and all older releases.

Todo: It seems that 6.0.0 breaks the ability to select an item from the list of topics. This needs to be investigated.
image
(background ftxui 6.0.0, foreground ftxui 5.0.0 - the elements are no longer selectable).

(This PR does *not* update the submodule, so the current build is still using 5.0.0)

@brakmic-aleksandar There is no chance that you might quickly look into this? I am not quite sure where to start.

### DIFF
--- a/app/mon/mon_tui/src/tui/view/component/scroller.hpp
+++ b/app/mon/mon_tui/src/tui/view/component/scroller.hpp
@@ -32,6 +32,8 @@
 #include "ftxui/dom/requirement.hpp"
 #include "ftxui/screen/box.hpp"
 
+#include <tui/ftxui_version_compatibility.hpp>
+
 namespace ftxui
 {
 class ScrollerBase : public ComponentBase
@@ -41,16 +43,12 @@ class ScrollerBase : public ComponentBase
   ScrollerBase(Component child) : selected_{selected_dummy} { Add(child); }
 
  private:
-#if FTXUI_VERSION_MAJOR >= 6
-  Element OnRender() final
-#else
-  Element Render() final
-#endif
+  Element FTXUI_COMPATIBILITY_RENDER() final
   {
     auto focused = Focused() ? focus : ftxui::select;
     auto style = Focused() ? inverted : nothing;
 
-    Element background = ComponentBase::Render();
+    Element background = ComponentBase::FTXUI_COMPATIBILITY_RENDER();
     background->ComputeRequirement();
     size_ = background->requirement().min_y;
     auto el =  dbox({
--- a/app/mon/mon_tui/src/tui/view/component/scroller.hpp
+++ b/app/mon/mon_tui/src/tui/view/component/scroller.hpp
@@ -1,6 +1,6 @@
 /* ========================= eCAL LICENSE =================================
  *
- * Copyright (C) 2016 - 2019 Continental Corporation
+ * Copyright (C) 2016 - 2025 Continental Corporation
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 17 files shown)


ArthurSonzogni:
Some context about why `Render` is now split into `OnRender` (users) and `Render` (ftxui) side.

**Context** I had to split this function in two:
- `Render` controlled by FTXUI.
- `OnRender` controlled by the users.

`Render()` allows me to tag element as "active" or "inactive" depending on the Component state. This allows to select the focused element in a hiearchical manner. This replaces the previous `focus`/`select` decorator where only a 2-level global focus was previously applied.

See `Render` [implementation](https://github.com/ArthurSonzogni/FTXUI/blob/main/src/ftxui/component/component.cpp#L122-L140)

KerstinKeller:
@ArthurSonzogni I did just what you explained, however, it leads to a different behavior and the app is broken for ftxui 6.0.0. I can no longer select elements from the list in the given application.
As I am not the author of the code, I don't fully grasp  the concept. However, from what I have seen, `OnRender` of some components call `Render` of other components, see 

https://github.com/eclipse-ecal/ecal/blob/5859ba612d2b6ae6ba449ddb81452de731a4d3a4/app/mon/mon_tui/src/tui/view/component/scroller.hpp#L44-L65

Is this allowed?
From quickly looking through the code, it was the first thing that stuck out.

app/mon/mon_tui/src/tui/view/component/scroller.hpp
```diff
@@ -41,7 +41,11 @@ class ScrollerBase : public ComponentBase
...
 
  private:
+#if FTXUI_VERSION_MAJOR >= 6
+  Element OnRender() final
+#else
   Element Render() final
+#endif
   {
     auto focused = Focused() ? focus : ftxui::select;
     auto style = Focused() ? inverted : nothing;
```
ArthurSonzogni, line 51:
```cpp
    Element background = ComponentBase::Render();
```
must be:
```
    Element background = ComponentBase::OnRender();
```

Or else the `background` won't be focused when active.

KerstinKeller:
@ArthurSonzogni I had made the changes that you suggested, however, the problem I described on top remains.
Basically I have some kind of list, where one element should be selected, and then additional information for that element is to be displayed.
Updating ftxui to 6.0 breaks this functionality.

Since I haven't authored the code, I have a hard time locating the source of the issue. Maybe the element of choice is no longer selected correctly (regarding the "focused" issue?)

Anything I can pay attention to in particular?

ArthurSonzogni:
I would be happy to debug this. The problem is that the last time I tried, there was too many obstacle to build the project.

If you can provide some instructions for me to try, I will figure this out.

KerstinKeller:
So in general, it should not be too difficult to build.
You will need a fairly recent CMake version (>3.25) and clone the repo including submodules (recursively), from this branch ([feature/update-ftxui-6](https://github.com/eclipse-ecal/ecal/tree/feature/update-ftxui-6)). Alternatively you can download the fat sources archive from the github actions (https://github.com/eclipse-ecal/ecal/actions/runs/14757653830/artifacts/3038617643) 

You don't need to build everything, i'd recommend samples & console apps (no gui), so you will have mon_tui and some samples (ecal_sample_person_send, ecal_sample_blob_send) to test with.

However, since I just tried a complete clean rebuild with which I though were good options for you, and failed, this might have to wait a little longer :/ 

I will let you know! And thanks so much for offering to help.

ArthurSonzogni:
I found the issue.

This is due to how the `dbox` has been updated in:
https://github.com/ArthurSonzogni/FTXUI/commit/ff305147ca920293d0f8e569bb5ff5d84086de2

In order to support color alpha and "blend" the bgcolor of the top layer on top of the layers below, we blend its bgcolor. Outside of those, we replace everything including the modifier like `inverted` as long as the top layer contains a non empty character.

In your case the "scroller" is implemented by adding an `inverted` empty line. Since this is empty, this has no effect.

What you can do:
1. Copy-paste the `dbox` implementation from ftxui5 and use it.
2. Try to set a black background and use an transparent bgcolor on the selected line. Example:


https://github.com/user-attachments/assets/754e4939-3af1-447b-917f-a583de56b16d

3. I reverted the change to `dbox`. I think I would have liked to use a differement name to avoid breaking users.

PR #2096, opened 2025-03-28
renamed send/receiver for c samples

KaiKosch:
### Description

- Renamed the hello samples for c
- Set frequency to 500ms in c and cpp hello_send samples
- Added a counter to hello_send.c

### DIFF
--- a/lang/c/samples/pubsub/string/hello_send/src/hello_send.c
+++ b/lang/c/samples/pubsub/string/hello_send/src/hello_send.c
@@ -49,7 +49,7 @@ int main()
     snprintf(snd_s, sizeof(snd_s), "HELLO WORLD FROM C (%d)", ++cnt);
 
     // send content
-    if(!eCAL_Publisher_Send(publisher, snd_s, sizeof(snd_s), NULL))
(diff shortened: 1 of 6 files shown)


hannemn:
@KaiKosch Please add a counter to the C sample of the published text message "HELLO WORLD FROM C (XXX)", just like it was done inthe C++ sample.

lang/c/samples/pubsub/string/hello_send/src/hello_send.c
```diff
@@ -44,14 +45,17 @@ int main()
   // send updates
   while(eCAL_Ok())
   {
+    // create message
+    snprintf(snd_s, sizeof(snd_s), "HELLO WORLD FROM C (%d)", ++cnt);
+
     // send content
     if(!eCAL_Publisher_Send(publisher, snd_s, sizeof(snd_s), NULL))
```
hannemn, line 52: As `sizeof(snd_s)` as size parameter is going to send the entire char buffer, it is more efficient in this case to determine string length via `strlen()` and pass it to the function.

hannemn (changes requested):
See comment

hannemn (approved):
Perfect, i think we can merge it now :-)

PR #2099, opened 2025-04-01
[setup] Set eCAL Launcher to be linked on the userdesktop instead of common desktop

Peguen:
### Description
s.a.

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -147,13 +147,13 @@ jobs:
       run: ctest -C Release -V
       working-directory: ${{ runner.workspace }}/_build/complete
 
-    - name: Pack SDK
-      run: cpack -C Debug
-      working-directory: ${{ runner.workspace }}/_build/sdk
-
     - name: Pack complete setup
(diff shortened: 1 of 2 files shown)


.github/workflows/build-windows.yml
```diff
@@ -147,13 +147,13 @@ jobs:
...
-      run: cpack -C Debug
-      working-directory: ${{ runner.workspace }}/_build/sdk
-
     - name: Pack complete setup
       run: cpack -C Release
       working-directory: ${{ runner.workspace }}/_build/complete
+    
+    - name: Pack SDK
+      run: cpack -C Debug
+      working-directory: ${{ runner.workspace }}/_build/sdk
```
KerstinKeller, line 157: With @FlorianReimold's explanation, I suspect that we need to pack the SDK prior  to packing the complete setup?

FlorianReimold: Yes, packing the complete build assumes the SDK to be available in a staging directory.

hannemn (approved):

PR #2100, opened 2025-04-01
Fixing description of string publisher class

aaroon-5:
### Description
Updated description of string publisher class


hannemn (approved):
LGTM

PR #2101, opened 2025-04-01
[cmake] Small adaption due to new cmake toolchain in GH actions

hannemn:


DownerCase:
@hannemn #2043 is ready to solve this.

PR #2103, opened 2025-04-02
[bugfix] Allow compatibility with CMake 4.0

### LINKED ISSUE #2041: CMake 4.0 Removes compatibility with CMake < 3.5

KerstinKeller:
Allow eCAL to successfully build with CMake 4.0

closes #2041

### DIFF
--- a/cmake/helper_functions/ecal_helper_functions.cmake
+++ b/cmake/helper_functions/ecal_helper_functions.cmake
@@ -67,3 +67,11 @@ macro(ecal_restore_warning_level)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS_OLD}")
   endif()
 endmacro()
+
+macro(ecal_variable_push var_name)
+  list(APPEND ecal_${var_name}_old "${${var_name}}")
+endmacro()
+
+macro(ecal_variable_pop var_name)
+  list(POP_BACK ecal_${var_name}_old "${var_name}")
+endmacro()
--- a/doc/samples/cpp/configuration/hello_config/CMakeLists.txt
+++ b/doc/samples/cpp/configuration/hello_config/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.16)
 
 project(hello_config)
 
--- a/doc/samples/cpp/configuration/publisher_config/CMakeLists.txt
+++ b/doc/samples/cpp/configuration/publisher_config/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.0)
+cmake_minimum_required(VERSION 3.16)
 
 project(publisher_config)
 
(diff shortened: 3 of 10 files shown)


KerstinKeller:
@hannemn Should we cherry pick this to 5.13 or do we rather downgrade CMake in the workflow? (Cherry pick might not be possible - we didn't have the providers in CMake 5.13)

Peguen (approved):
🧾 ✏️

FlorianReimold:
> @hannemn Should we cherry pick this to 5.13 or do we rather downgrade CMake in the workflow? (Cherry pick might not be possible - we didn't have the providers in CMake 5.13)

I think we should cherry-pick (or re-implement) the same approach for eCAL 5.13 and 5.12. However, I think we should omit the raised cmake requirement (to 3.16) for those branches.

hannemn:
Yes, we should apply the adaptions in the same way for the support branches as well, making sure that the builds run through independent from getting a cmake 4.x or 3.x runner.

PR #2108, opened 2025-04-06
CI: Use shallow submodules and more system packages

DownerCase:
### Description

Aims to speed up CI times in two ways:

1. By doing shallow clones of the submodules and main repo
2. By building less vendored code, instead using system packages were suitable.

Effect:

This speeds up CI runs by up to roughly 2 minutes (there is a _lot_ of variance). Less than I was hoping, but is ~10% in the best case.

Detailed changes:

- The eCAL repo is now checkedout using `--filter=tree:0`. This is like a less aggressive shallow clone; the commits are fetched but their contents are fetched on-demand. This lets the tag-based versioning to continue to work but minimises the data transferred.
- More unneeded submodules are de-initialized.
- Ubuntu:
  - package lists are now split into common and specific to reduce repetition.
  - `spdlog` and `gmock` have been added to the installed packages.
  - `spdlog`, `gtest`, and `yaml-cpp` are no longer built from the submodules.
    - The `gtest` and `yaml-cpp` packages were being installed and then ignored and built from source...
  - Packages are not installed without "recommendations" to reduce implicit dependencies.
- Added missing compatability function call for `yaml-cpp` in `linuxptp` (exposed by above changes)
- Added missing includes to shm test (exposed by above changes)

### Related issues

-

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -30,8 +30,16 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v4
       with:
-        submodules:  'true'
+        submodules:  'false'
         fetch-depth: 0
+        filter: 'tree:0' # Download all commits but blobs and trees on demand
+
+    - name: Update / download Submodules (selected ones)
+      run: |
+        cd $GITHUB_WORKSPACE
+        git submodule init
+        git submodule deinit thirdparty/hdf5/hdf5 thirdparty/protobuf/protobuf
+        git submodule update --single-branch --depth 1
 
     - name: Display Python version
       run: python -c "import sys; print(sys.version)"
--- a/.github/workflows/build-ubuntu.yml
+++ b/.github/workflows/build-ubuntu.yml
@@ -23,24 +23,38 @@ jobs:
 
     - name: Install Dependencies
       run: |
-        sudo apt update
+        common_packages=(
+          ninja-build
+          doxygen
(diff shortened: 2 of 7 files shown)


KerstinKeller (approved):

DownerCase:
> Hi @DownerCase thanks for making this PR, it looks all good.
> 
> For the future, after the 6.0 Release I did want to look into bringing down the build times. In general I would prefer to provide our dependencies with Conan, however I have yet to find a good way to cache these dependencies, as we don't have access to an artifactory instance, but not all of the dependencies that we need are pre-build on Conan Center Index. Integrating CCache (or an alternative which works for Windows, too), is a second step (or first, as both are independent of each other).
> 
> This way we could get down the build times significantly, my aim would be to achieve build times < 10 min for all platforms.

Yes, Conan sounds good! Even if you can get some pre-builts from Conan it will be an improvement over the current situation. The rest is just a case of working with the Conan community to fill the gaps I guess...

CCache was absolutely on my list! Just need the time to do it, but that should help significantly.

10 minutes sounds like a good target, but I expect Windows will put up a tough fight, seeing as so much is being done there (Debug, Release, Docs, C#...)

---

Re: Conan. How are you thinking of balancing with use of system packages on the Linux side? I guess the `.deb`s won't be able to make use of it for fear of ending up having both Conan and system packages ending up duplicated and mixed together in an end product.

PR #2110, opened 2025-04-07
core/minor-fixes-2025-04-07

rex-schilasky:
### Description

Two minor fixes:

- fixed log issues in CMakeLists.txt
- log_level_info level default active for logging sink console (as it was set in 5.x)

### DIFF
--- a/ecal/core/include/ecal/config/logging.h
+++ b/ecal/core/include/ecal/config/logging.h
@@ -60,7 +60,7 @@ namespace eCAL
 
       struct Configuration
       {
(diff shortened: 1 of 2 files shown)


ecal/core/include/ecal/config/logging.h
```diff
@@ -60,7 +60,7 @@ namespace eCAL
 
       struct Configuration
       {
-        Sink                console { true,  log_level_warning | log_level_error | log_level_fatal };                   //!< default: true, log_level_warning | log_level_error | log_level_fatal
```
KerstinKeller, line 63: @Peguen does this need to be fixed in multiple places (aka .yaml file)

hannemn (approved):

PR #2112, opened 2025-04-08
[fix] Mutex and test ODR violations

### LINKED ISSUE #2111: ODR violation with `::named_mutex` on Linux when robust_clocklock is enabled
### Problem Description

There is an ODR violation of the symbol `::named_mutex` between `ecal_named_mutex_impl.cpp` and `ecal_named_mutex_robust_clocklock_impl.cpp`

https://github.com/eclipse-ecal/ecal/blob/aee20697ce4f1c13bc22f63be6c816173c520ec5/ecal/core/src/io/mtx/linux/ecal_named_mutex_impl.cpp#L38-L44

https://github.com/eclipse-ecal/ecal/blob/aee20697ce4f1c13bc22f63be6c816173c520ec5/ecal/core/src/io/mtx/linux/ecal_named_mutex_robust_clocklock_impl.cpp#L36-L40

Both files are included in the build:
https://github.com/eclipse-ecal/ecal/blob/aee20697ce4f1c13bc22f63be6c816173c520ec5/ecal/core/CMakeLists.txt#L99-L104

Which causes an ODR violation, meaning the program is ill-formed and as a result **the entire operation of eCAL is undefined**

### How to reproduce

1. cmake --preset core --fresh -DCMAKE_CXX_FLAGS="-flto -Werror=odr" -DBUILD_SHARED_LIBS=OFF
2. cmake --build --preset core
3. Linking will fail (`ecal/ecal/core/src/io/mtx/linux/ecal_named_mutex_impl.cpp:38:19: error: type ‘struct named_mutex’ violates the C++ One Definition Rule [-Werror=odr]`)

(Nb: LTO is required for the linker to find ODR violations)

### How did you get eCAL?

### Environment

- eCAL: Master
- OS: Linux

### eCAL System Information

_No response_

DownerCase:
### Description

Fixes two instances of ODR violations:

1. From `::named_mutex`
2. From test fixtures

### Related issues

- 
Closes #2111

### DIFF
--- a/ecal/core/src/io/mtx/linux/ecal_named_mutex_impl.h
+++ b/ecal/core/src/io/mtx/linux/ecal_named_mutex_impl.h
@@ -27,7 +27,6 @@
 #include <cstdint>
 #include <string>
 
-typedef struct named_mutex named_mutex_t;
 
 namespace eCAL
 {
@@ -51,6 +50,7 @@ namespace eCAL
 
     bool Lock(int64_t timeout_) final;
     void Unlock() final;
+    struct named_mutex_t;
   private:
     named_mutex_t* m_mutex_handle;
     std::string m_named;
--- a/ecal/core/src/io/mtx/linux/ecal_named_mutex_impl.cpp
+++ b/ecal/core/src/io/mtx/linux/ecal_named_mutex_impl.cpp
@@ -35,13 +35,13 @@
 #include <cstdint>
 #include <string>
 
-struct alignas(8) named_mutex
+struct alignas(8) eCAL::CNamedMutexImpl::named_mutex_t
 {
   pthread_mutex_t  mtx;
   pthread_cond_t   cvar;
   uint8_t          locked;
 };
-typedef struct named_mutex named_mutex_t;
+using named_mutex_t = eCAL::CNamedMutexImpl::named_mutex_t;
 
 namespace
 {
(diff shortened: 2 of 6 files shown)


ecal/core/src/io/mtx/linux/ecal_named_mutex_impl.h
```diff
@@ -51,9 +50,10 @@ namespace eCAL
 
     bool Lock(int64_t timeout_) final;
     void Unlock() final;
+    struct named_mutex_t;
```
KerstinKeller, line 53: I was not aware that we can forward declare structs inside a class. It feels a bit sketchy. (So does the other thing, that has been done previously, and as you found out, violates odr - It was very old school, C-like).

KerstinKeller (commented):
Thanks for finding those. I wonder that we didn't get any warnings.
As stated, a bit confused that forward declarations are possible inside a class declaration, because I though you'd need to know the size for the class layout.
Maybe we should take a look  why it is defined in the cpp file.

For the testcases though, very clear, makes a lot of sense.

DownerCase:
> Thanks for finding those. I wonder that we didn't get any warnings. As stated, a bit confused that forward declarations are possible inside a class declaration, because I though you'd need to know the size for the class layout. Maybe we should take a look why it is defined in the cpp file.
> 
> For the testcases though, very clear, makes a lot of sense.

You didn't get any warnings simply because the standard says a diagnostic is not required and the toolchain generally assumes a well formed program. :shrug: 

Forward declaration inside a class is unusual sure; but I believe its valid. The difference is instead of meaning there will by a type `::named_mutex` its now saying there is a type `ecal::CNamedMutexImpl::named_mutex_t`, and the rest of the normal rules for visibility and forward declaration is used.

Specifically, we don't need the size because its only being used via a pointer (and the size of a pointer is always known) as per usual rules for forward declaration.

KerstinKeller (approved):

PR #2120, opened 2025-04-10
added tests for c api

KaiKosch:
### Description
- added test for pubsub
- added test for services
- added test for c config.cpp


lang/c/tests/pubsub_test/src/pubsub_test.cpp
```diff
@@ -0,0 +1,211 @@
...
+{
+  char snd_s[256];
+  int cnt = 0;
+  int callback_count = 0;
+
+  // add callback
+  EXPECT_EQ(0, eCAL_Subscriber_SetReceiveCallback(subscriber, OnReceive, &callback_count));
+
+  // sleep here because otherwise it takes to long to set the receive-callback
+  eCAL_Process_SleepMS(8000);
```
hannemn, line 113: too much sleep ;-)

lang/c/tests/pubsub_test/src/pubsub_test.cpp
```diff
@@ -0,0 +1,207 @@
...
+  EXPECT_EQ(0, eCAL_Subscriber_SetReceiveCallback(subscriber, OnReceive, &callback_count));
+
+  // sleep here because otherwise it takes to long to set the receive-callback
+  eCAL_Process_SleepMS(2000);
+
+  // send messages
+  for(int i = 0; i < 10; i++) 
+  {
+    // create message
+    snprintf(snd_s, sizeof(snd_s), "HELLO WORLD FROM C (%d)", ++cnt);
```
Peguen, line 131: You don't need to construct the message from the sample here. Just go with a constant message as you are only interested in the counter.

lang/c/tests/pubsub_test/src/pubsub_test.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+    // send content
+    if(!eCAL_Publisher_Send(publisher, snd_s, strlen(snd_s), NULL))
+      printf("Published topic \"Hello\" with \"%s\"\n", snd_s);
+    else
+      printf("Sending topic \"Hello\" failed !\n");
+
+    // sleep 500 ms
+    eCAL_Process_SleepMS(500);
```
Peguen, line 140: Sleep way less ... maybe around 10ms, so the test does not take so long

lang/c/tests/pubsub_test/src/pubsub_test.cpp
```diff
@@ -0,0 +1,207 @@
...
+#include <gtest/gtest.h>
+
+
+class pubsub_test_c : public ::testing::Test {
+
+  protected:
+    eCAL_Subscriber* subscriber;
+    eCAL_Publisher* publisher;
+    struct eCAL_SDataTypeInformation data_type_information;
+    const char *topic_name = "hello";
```
Peguen, line 31: To differentiate between the sample and the test, you can change this here to e.g. to "C-Binding-Test" or something like that

lang/c/tests/config_test/src/config_test.cpp
```diff
@@ -0,0 +1,233 @@
...
+        EXPECT_EQ(configuration0->publisher.layer_priority_local[i], eCAL_GetConfiguration()->publisher.layer_priority_local[i]);
+    }
+
+    for (int i = 0; i < int(configuration0->publisher.layer_priority_remote_length); i++)
+    {
+        EXPECT_EQ(configuration0->publisher.layer_priority_remote[i], eCAL_GetConfiguration()->publisher.layer_priority_remote[i]);
+    }
+}
+
+TEST_F(config_test_c, print)
```
Peguen, line 175: What do you want to achieve with this test?

PR #2125, opened 2025-04-14
fix: Use asio to validate ipv4 address strings instead of regex

### LINKED ISSUE #2124: ecal_generate_config: [IpAddressV4] No valid IP address: 255.255.255.240
### Problem Description

When `ecal_generate_config` tries to create the config file in some build configurations it will fail unexpectedly.

```
[ 99%] Generating /build/ecal/src/build/ecal/core/cfg/ecal.yaml;/build/ecal/src/build/ecal/core/cfg/ecaltime.yaml
terminate called after throwing an instance of 'std::invalid_argument'
  what():  [IpAddressV4] No valid IP address: 255.255.255.240
make[2]: *** [ecal/core/cfg/CMakeFiles/run_ecal_generate_config.dir/build.make:74: ecal/core/cfg/ecal.yaml] Aborted (core dumped)
make[1]: *** [CMakeFiles/Makefile2:825: ecal/core/cfg/CMakeFiles/run_ecal_generate_config.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
``` 

### How to reproduce

```sh
cmake --preset core --fresh -DCMAKE_CXX_FLAGS="-flto" -DCMAKE_CXX_COMPILER="g++"
cmake --build --preset core --target run_ecal_generate_config
```

I also produce the error without LTO as part of the packaging for the AUR which is the real problem configuration.

Using clang does not produce the problem with either libc++ or libstdc++

I cannot create a minimal test case to show the behaviour, that behaves normally.

It seems this is a gcc std::regex bug.

### How did you get eCAL?

### Environment

- eCAL: master
- OS: Arch Linux
- gcc 14.2.1+r753+g1cd744a6828f-1
- glibc 2.41+r9+ga900dbaf70f0-1

### eCAL System Information

```text

```

DownerCase:
### Description

Replaces use of `std::regex` to validate ipv4 address strings with asio.

Does technically remove support of expressing IPv4 addresses in hex but I suspect that didn't work anyway (as at least one usage path leads into an `asio::ip::make_address` call).

Tidies up some whitespace/headers and clang-tidy reports as well.

### Related issues

Closes #2124

### DIFF
--- a/ecal/core/include/ecal/types/custom_data_types.h
+++ b/ecal/core/include/ecal/types/custom_data_types.h
@@ -27,9 +27,6 @@
 #include "ecal/os.h"
 
 #include <string>
-#include <iostream>
-#include <limits>
-#include <stdexcept>
 
 namespace eCAL
 {
@@ -53,13 +50,13 @@ namespace eCAL
       ECAL_API IpAddressV4& operator=(const std::string& ip_string_);
       ECAL_API IpAddressV4& operator=(const char* ip_string_);
       ECAL_API operator const std::string&() const;
-      ECAL_API bool operator==(const eCAL::Types::IpAddressV4& rhs) const;  
+      ECAL_API bool operator==(const eCAL::Types::IpAddressV4& rhs) const;
       ECAL_API friend bool operator==(eCAL::Types::IpAddressV4 lhs, const char* ip_string_);
       ECAL_API friend bool operator==(const char* ip_string_, eCAL::Types::IpAddressV4 rhs);
       ECAL_API friend bool operator==(eCAL::Types::IpAddressV4 lhs, const std::string& ip_string_);
       ECAL_API friend bool operator==(const std::string& ip_string_, eCAL::Types::IpAddressV4 rhs);
 
-    private:            
+    private:
       ECAL_API void validateIpString(const std::string& ip_address_);
 
       std::string m_ip_address{};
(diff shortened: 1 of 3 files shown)


DownerCase:
Annoying corner on `asio::ip::make_address_v4`. The parsing is platform dependent when the IP is not fully specified...

Under the hood asio will use [`WSAStringToAddressW`](https://learn.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsastringtoaddressw) which is notes: _"Any missing components of the address will be defaulted to a reasonable value, if possible. For example, a missing port number will default to zero."_

Unsure if this is acceptable (ideally all platforms will natively parse in the same way, but they don't), though as long as all 4 segments are specified (as is normally done?) there's no problem.

See also:

https://github.com/chriskohlhoff/asio/issues/274

KerstinKeller:
Thanks for finding / fixing this.
I am a bit reluctant, to pull in asio into this part of eCAL, but it's ok since it doesn't leak into the headers.
We should aim for the same behavior on all platforms, for sure, and I feel that it's ok to request a properly specified IP address.
Upside is also that it will easy to extend to IPv6 at a later point.

DownerCase:
> I am a bit reluctant, to pull in asio into this part of eCAL, but it's ok since it doesn't leak into the headers. 

I assumed it would be ok as asio is used already by eCAL core for the udp receiver and is what eventually consumes the data represented in this class anyway.

> We should aim for the same behavior on all platforms, for sure, and I feel that it's ok to request a properly specified IP address. 

Yeah, definitely. I imagine people were already doing that anyway.

> Upside is also that it will easy to extend to IPv6 at a later point.

Yup, that would be trivial for this class (I'd image other parts would need more attention)

KerstinKeller (approved):
👍

PR #2126, opened 2025-04-14
Align service samples across C++, C and C#.

KerstinKeller:
Name the service mirror and add method reverse.
Clients alternate in calling methods echo and reverse.

### Description
Server / Clients are now interoperable between different language bindings.


ecal/samples/cpp/services/mirror_client/src/mirror_client.cpp
```diff
@@ -27,10 +27,10 @@
 int main()
 {
   // initialize eCAL API
-  eCAL::Initialize("minimal client");
+  eCAL::Initialize("minimal client c++");
```
Peguen, line 30: Currently in the pubsub samples we have the naming convention for c++ like the file name. In this case it would be "mirror_client". For other languages you would add the language, e.g. "mirror_client_python".

ecal/samples/cpp/services/mirror_server/src/mirror_server.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+  // The return value here has no actual meaning. It's made available to the caller.
+  return 0;
+}
+
+// main entry
+int main()
+{
+  // initialize eCAL API
+  eCAL::Initialize("minimal server c++");
```
Peguen, line 60: Same with the naming here. Either we align it in these samples or we change it in the pubsub samples.

lang/c/samples/services/mirror_client_c/src/mirror_client_c.c
```diff
@@ -0,0 +1,99 @@
...
+  else
+  {
+    printf("Service / method '%s' not found :-(\n\n", method_name);
+  }
+
+  // Memory of response array needs to be deallocated explicitly.
+  eCAL_Free(response);
+
+  // Sleep a second
+  eCAL_Process_SleepMS(1000);
```
Peguen, line 69: I would not sleep in the function, rather do it after the call in the main if you want to.

lang/c/samples/services/mirror_client_c/src/mirror_client_c.c
```diff
@@ -0,0 +1,99 @@
...
+}
+
+
+int main()
+{
+  // The client handle
+  eCAL_ServiceClient *client;
+
+  // Initialize eCAL API
+  eCAL_Initialize("mirror client c", NULL, NULL);
```
Peguen, line 79: Same about naming here. 

"mirror_client_c"

lang/csharp/Eclipse.eCAL.Core.Samples/service/mirror_client_csharp.cs
```diff
@@ -48,20 +48,22 @@ static void PrintServiceResponse(ServiceResponse serviceResponse)
   static void Main()
   {
     // Initialize eCAL API.
-    Core.Initialize("minimal client csharp");
+    Core.Initialize("mirror client c#");
```
Peguen, line 51: We currently used "C#" as "csharp", so in this case it would be "mirror_client_csharp"

lang/csharp/Eclipse.eCAL.Core.Samples/service/mirror_client_csharp.cs
```diff
@@ -48,20 +48,22 @@ static void PrintServiceResponse(ServiceResponse serviceResponse)
...
     // Create a service client for service "service1"
-    ServiceClient serviceClient = new ServiceClient("service1");
+    ServiceClient serviceClient = new ServiceClient("mirror");
+    string[] methods = new string[] { "echo", "reverse" };
+    int i = 0;
 
     while (Core.Ok())
     {
       // Create the request payload.
       byte[] content = Encoding.UTF8.GetBytes("hello");
```
Peguen, line 63: Let's also name this as request and the content as "Hello" like in the cpp sample.

lang/csharp/Eclipse.eCAL.Core.Samples/service/mirror_server_csharp.cs
```diff
@@ -23,30 +23,51 @@
...
-    Console.WriteLine("Request: " + Encoding.UTF8.GetString(request));
+    byte[] response = request;
+
+    Console.WriteLine("Method  : '" + methodInfo.MethodName + "' called");
+    Console.WriteLine("Request : " + Encoding.UTF8.GetString(request));
+    Console.WriteLine("Response: " + Encoding.UTF8.GetString(response));
+    Console.WriteLine();
 
     // Echo the request data back as the response.
     return request;
```
Peguen, line 36: Wouldn't we want to return the response here? It will not change the outcome, but from a logical point of view I would think its response.

lang/csharp/Eclipse.eCAL.Core.Samples/service/mirror_server_csharp.cs
```diff
@@ -23,30 +23,51 @@
...
+
+    // Echo the request data back as the response.
+    return response;
+  }
 
   static void Main()
   {
     // Initialize eCAL API.
-    Core.Initialize("minimal server csharp");
+    Core.Initialize("mirror server c#");
```
Peguen, line 55: Same for naming here

Peguen (approved):
🕺

PR #2127, opened 2025-04-15
[CI] Integrate ccache for Linux and revamp

DownerCase:
### Description

Integrates ccache to greatly speed up the Linux CI runs and revamps a lot of the workflow.

- Integrates ccache with a 100M limit
  - A new cache does get created per run (as caches are immutable)
  - A new workflow was added to cleanup old caches when a PR gets closed (to reduce the likelihood of exceeding the max total cache size on GitHub (10GB) )
- Uses capnproto and tinyxml2 from Ubuntu repos, rather than building them
- Disables the `man-db` trigger to save ~1 minute on installing dependencies
- Inverts the submodule init logic to be an explicit llist of the ones required (also sidesteps warning spam that was happening)
- Moved(?) the build directory to avoid needing to set the working directory explicitly
- Reworked the `.deb` package naming to have CPack name it correctly itself.
  - The name of the file itself is unchanged from what was originally being done.
  - However, the name of the artifact on GitHub has changed slightly

### Results

Now a fully cached CI run takes 9-10 minutes. Effectively **2x faster** than master (18-20 mins).

### Related issues

- Follow on from #2108


.github/workflows/build-ubuntu.yml
```diff
@@ -47,78 +58,79 @@ jobs:
           )
         fi
 
+        # Disable man-db update
```
KerstinKeller, line 61: How much time does is save to disable the auto update?
Could we instead remove "apt-get update"?

DownerCase: man-db refers to the underlying system for the `man` pages. Without disabling it the CI spends about 1 minute doing stuff for the man pages which will never be used.
I don't know _exactly_ what is happening. But safe to say it doesn't matter if `man` is missing a few pages.

`apt-get update` is totally unrelated, not problematic, and should be kept.

KerstinKeller:
Also we are aware that the tests do take a long time. Basically they are system tests, not unit tests.
There might be a few things that we should do, e.g. increase registration refresh rates so we don't always have to wait for such a long time before being able to start communication.
We wanted to take a close loop at the tests, and we will keep timing in mind for that.

DownerCase:
Totally agree with your comments on the version constant. I'll leave it to you.

No problem with the doc build. Just frustrating how Ninja schedules it. Perhaps split CI (testing) and packaging workflows? Idk

Yeah, I haven't looked at the tests. Just know from the small corner I interacted with there were plain sleep statements which I shunned in favor of a condition variable and a timeout. Being able to run some in parallel would obviously help but nontrivial to do (topic names must be unique for a start)

KerstinKeller (approved):

PR #2134, opened 2025-04-17
[CI] Use ccache on MacOS

DownerCase:
### Description

Integrates ccache into the MacOS workflow:

- The sphinx documentation is now built with multiple threads (have not measured the impact)
- The ccache steps have been extracted into composite actions to prevent duplication
- Building documentation on MacOS was disabled as it was not used and the Ubuntu workflow already builds it to ensure validity.

### Related issues

- Follow on from #2108


KerstinKeller (approved):

KerstinKeller:
The common action looks good!
Will merge this. I wonder if it's also applicable to Windows, where we have the longest build times by far...

PR #2136, opened 2025-04-17
[docs] Sample adaptions

Peguen:
### Description
- Previously the samples of the documentation where separate ones stored in the docs folder. Now we use the repo default samples for improved maintainability.
- Samples are adapted to be
    - basically the same just in a different language, means behaviour and naming got unified
    - more verbose
    - more documented in order to get all descriptions out of the source code instead of comments in the docu
- New sections in the docu are created. Now its devided into functionality and serialization method. That means, that e.g. for a String Publisher you have one page, but you can click through the different languages via tabs.
- Section for Client/Server Samples is prepared

TBD for future PR:
- Add the client/server samples
- Add configuration samples


lang/c/samples/pubsub/string/hello_send/src/hello_send.c
```diff
@@ -25,43 +26,94 @@
...
+      The string will be "HELLO WORLD FROM C (1)", "HELLO WORLD FROM C (2)", ...
+    */
+    snprintf(message, sizeof(message), "HELLO WORLD FROM C (%d)", ++loop_count);
 
-    // send content
-    if(!eCAL_Publisher_Send(publisher, snd_s, strlen(snd_s), NULL))
-      printf("Published topic \"hello\" with \"%s\"\n", snd_s);
+    /*
+      Send the content to other eCAL Processes that have subscribed to the topic "hello".
+      The message is sent as a raw string, so we use the length of the string as the size of the message.
```
KerstinKeller, line 93: I think it makes sense to explicity mention that we are not null terminating the string

eCAL transports messages as a byte array and length, so for interop with other languages, we will send a non-null terminated strings over the wire.

KerstinKeller (approved):

PR #2141, opened 2025-04-22
[CI] Use ccache on Windows

DownerCase:
### Description

Updates the Windows workflow to make use of ccache. In total this brings a CI run down from ~40 minutes to ~20 minutes. The largest gain is speeding up the release build 10x from 20 minutes to **2**. The largest steps are the release testing, debug (SDK) build (because ccache does not work with /Zi), and packaging. 

### Detailed changes

This PR is also ordered into logical commits (with commit messages).

- `install-qt-action` updated to v4
  - `setup-python` was enabled to allow it pick an appropriate python version for us (currently 3.10)
  - The qt version was relaxed to select the most recent 6.6 patch version
  - The `archives` option is set to minimize the qt install
  - Caching is now enabled
- Initializing and updating of the submodules combined to one line
  - (The windows build uses all submodules so the explicit list is ommited)
- `ccache` is used for building
  - Currently it only works for the release build because ccache does not support the `/Zi` flag
  - Using ccache with Windows requires using Ninja as the generator
  - The `msvc-dev-cmd` action is used to select the msvc toolset as if `vcvarsall` was used
  - CMake requires a Visual Studio generator to use the C# language so the C# bindings have been disabled from the SDK build (and handled separately)
  - As udpcap delay loads `wpcap.dll` it is required for consumers to link with `delayimp.lib`. Visual Studio will add this implicitly but for Ninja it must be explicitly added.
- Dedicated steps added for C#
  - The complete release build is installed into a temporary tree for the C# subproject to use
  - Multiple changes were made to `lang/csharp/CMakeLists.txt` to make it standalone:
    - The missing `cmake_minimum_required` was added
    - The project version is set in the same way as for the main eCAL project
    - The handling of protobuf versions (of 4.x or newer) was fixed.
    - Testing was enabled (**note:** No tests are actually registered with CTest so no tests are run)
    - CPack variables are set for packaging similar to the main eCAL project
- The documentation build no longer depends on the Python modules being built
  - Ninja was failing as those targets weren't enabled
  - It is unclear to me why this was present.
- The innosetup script was updated to include the C# bindings being built separately
- Some whitespace was trimmed

### Potential further work

Splitting the workflow into multiple parallel jobs could further reduce the time needed. Such as moving the Debug and documentation to be built in parallel to the Release build critical path and later brought together for final packaging.

Just doing the Debug build in parallel would get things closer to ~16 minutes. But it all comes at increased complexity.

### Related issues

- Follow on from #2108 #2127 #2134

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -64,12 +64,28 @@ jobs:
 
       shell: cmd
 
+      # Using ccache on Windows requires the Ninja generator
+      # This action calls vcvarsall for us to setup the desired toolchain
+    - name: Setup MSVC toolchain
+      uses: ilammy/msvc-dev-cmd@v1
+      with:
+        arch: x64
+        toolset: 14.29 # v141 / vs2019
+
+    - name: Prepare ccache and restore cache
+      id: ccache_cache-restore
+      uses: ./.github/actions/ccache-prepare
+      with:
+        key-root: windows-2019
+        cache-dir: ${{ github.workspace }}/.ccache
+
     - name: CMake SDK
       run: |
         mkdir "${{ runner.workspace }}\_build\sdk\"
         cd "${{ runner.workspace }}/_build/sdk"
 
-        cmake %GITHUB_WORKSPACE% -G "Visual Studio 16 2019" -A x64 -T v142 ^
+        cmake %GITHUB_WORKSPACE% -G "Ninja" ^
+        -DCMAKE_C_COMPILER=cl -DCMAKE_CXX_COMPILER=cl ^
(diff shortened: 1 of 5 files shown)


KerstinKeller:
Yay for this PR! From 40 down to 20 minutes is great! Will review thouroughly the next days.

KerstinKeller:
> The documentation build no longer depends on the Python modules being built
Ninja was failing as those targets weren't enabled
It is unclear to me why this was present.

It is there because we used to have automated documentation generation based on the python extensions (compiled .pyd files). (see [here](https://eclipse-ecal.github.io/ecal/stable/_api_python/ecal.html))
They were imported, and thus all functionality was automatically documented.
However, it seems to be broken on the master branch anyways, so I guess we can tackle it in a different way.

KerstinKeller (approved):

PR #2154, opened 2025-05-07
[csharp] add a callback-based ServiceClient API and a corresponding C# sample

rex-schilasky:
### Description
This PR introduces a new, callback-based variant of the service client call (CallWithCallback) alongside the existing CallWithResponse. It also adds or updates C# sample code to demonstrate the new API.

#### API Changes
- Added a new constructor overload `ServiceClient(String^ serviceName)` allowing creation of a client without explicitly providing a `ServiceMethodInformationList^`.
- Added `ServiceClient::CallWithCallback(String^ methodName, array<Byte>^ request, ResponseCallback^ callback, int timeoutMs)` to the C# wrapper.

#### Sample Added
- mirror_client_callback_csharp.cs: Demonstrates registering a response-callback delegate instead of using CallWithResponse.


lang/csharp/Eclipse.eCAL.Core.Samples/service/CMakeLists.txt
```diff
@@ -24,8 +24,9 @@ find_package(eCAL REQUIRED)
 macro(ecal_add_csharp_core_sample sample_name)
   ecal_add_sample(${sample_name} ${sample_name}.cs)
   target_link_libraries(${sample_name} PRIVATE Eclipse.eCAL.Core)
-  set_property(TARGET ${sample_name} PROPERTY FOLDER samples/csharp/binary/service)
+  set_property(TARGET ${sample_name} PROPERTY FOLDER samples/csharp/core/service)
```
KerstinKeller, line 27:
```suggestion
  set_property(TARGET ${sample_name} PROPERTY FOLDER samples/csharp/services/binary)
```
at least this is analog to C / C++ (Python is still a bit different).

rex-schilasky: Unfortunately it's not that easy to align with the other samples. Currently the C# samples are aligned to the 3 C# modules (core, protobuf, string). So there are C# Core (binary) samples for pubsub and service. For C# Protobuf we have typed pubsub and monitoring. For C# String we have typed pubsub. I would keep it like it is, so the solution structure is reflecting the C# modules structure.

image

PR #2199, opened 2025-06-10
[csharp] add a callback-based ServiceClient API and a corresponding C# sample (2)

rex-schilasky:
### Description
This PR introduces a new, callback-based variant of the service client call (CallWithCallback) alongside the existing CallWithResponse. It also adds or updates C# sample code to demonstrate the new API.

#### API Changes
- Added `ServiceClient::CallWithCallback(String^ methodName, array<Byte>^ request, ResponseCallback^ callback, int timeoutMs)` to the C# wrapper.

#### Sample Added
- mirror_client_callback_csharp.cs: Demonstrates registering a response-callback delegate instead of using CallWithResponse.


lang/csharp/Eclipse.eCAL.Core.Samples/service/mirror_client_csharp.cs
```diff
@@ -112,7 +116,20 @@ Iterate through all responses and print them.
       }
       else
       {
-        Console.WriteLine("Method blocking call failed.");
+        Console.WriteLine("Method call with reponse failed.");
```
Peguen, line 119: Little Typo - response

rex-schilasky: Will fix this 😀

Peguen (approved):

Issue #2246, opened 2025-07-17
Concurrency issue with CNamedEvent on Unix Systems


KerstinKeller:
### Problem Description

The CNamedEvent class in eCAL is used to signal between Publishers and Subscribers on the same system. Users can wait on these events.

The class seems to have a  concurrency issue.
If two `CNamedEvent` objects are created at exactly the same time, they might not be able to signal each other.

https://github.com/eclipse-ecal/ecal/blob/dd3dde0c6aa6541c2369466fb7c9d38b09aacad4/ecal/core/src/ecal_event.cpp#L323-L334

This is the root cause of #2233

### How to reproduce

-

### How did you get eCAL?

I don't know

### Environment

- OS: Ubuntu

### eCAL System Information

- [x] Fix the actual error by checking if creation of the event returned `nullptr`
- [x] Create testcases that trigger the error

hannemn:
Resolved with #2251

PR #2251, opened 2025-07-21
[test] Testcase for opening named events in parallel

hannemn:
This testcase provokes a race condition failure by opening an os specific event in parallel which is described in #2246 and resolved with #2247 .


KerstinKeller (changes requested):
Please remove gGetNativeEventHandle function again.

KerstinKeller (approved):
Please rename commit message when merging PR
[core] More robust event handling on Linux (or something similar)

PR #2289, opened 2025-07-31
Updated color scheme to Eclipse Orange

FlorianReimold:
Switched to the Eclipse Foundation orange as main color. Continental will be AUMOVIO (with new color scheme) in the future and eCAL has been an Eclipse Project for quite some time, so I think it is OK to make that change

Eclipse-orange is `#F79422` / `rgb(247,148,34)` / `hsl(32,93%,55%)`

- [X] Documentation CSS
- [ ] Documentation SVGs with orange color scheme
- [ ] Logos (Installer, logo dir, favicon, documentation)
- [ ] eCAL Launcher


KerstinKeller:
Hmhm I think that then we should make up a new color scheme and adhere to that. Which might include the eclipse orange 😄

PR #2300, opened 2025-08-02
[build] Avoid mirror_server name clash with HDF5

DownerCase:
### Description

HDF5 may provide a CMake target with the name `mirror_server` as they have an executable of the same name.
If you try to configure eCAL with such an HDF5 package, HDF5 support enabled, and the `mirror_server` sample enabled it will fail as the target will already exist.

To get around this I simply appended `_cpp` to the problematic samples (as there is also `mirror_server_c`)

Eg: The Arch Linux HDF5 package has `mirror_server` included: https://archlinux.org/packages/extra/x86_64/hdf5/

### Related issues

-

### DIFF
--- a/ecal/samples/cpp/services/mirror_client/CMakeLists.txt
+++ b/ecal/samples/cpp/services/mirror_client/CMakeLists.txt
@@ -18,7 +18,7 @@
 
 cmake_minimum_required(VERSION 3.15)
 
-project(mirror_client)
+project(mirror_client_cpp)
 
 find_package(eCAL REQUIRED)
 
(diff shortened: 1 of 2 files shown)


KerstinKeller:
@DownerCase 
Would you want to merge this only in master, or the release branches for 6.0?

@hannemn @Peguen 
If we do this, should we be consistent and do it for all cpp samples?
Or should we go ahead and prefix with `ecal_` (which is done anyways for the actual executable names).

KerstinKeller:
Ok, got it, makes sense 😄

hannemn (approved):

PR #2310, opened 2025-08-07
[Licenses] Added python-based SBOM collection script

FlorianReimold:
Added a python-based SBOM script. Each submodule describes its own values in a file called `sbom.py`. The python structure makes it easy to convert to other formats, e.g. a directory of License files, the debian License file, the Documentation, the NOTICE.txt file etc.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,8 +6,8 @@ license = "Apache-2.0"
 license-files = [
   "LICENSE.txt",
   "NOTICE.md",
-  "licenses_python_binding/**/*",       # Matches all "main" licenses in subdirs
-  "licenses_python_binding/**/**/*"     # Matches all licenses from "thirdparty" sub-subdirs (unfortunatelly, the ** is not recursive here)
+  "LICENSES/LICENSES_python_bindings/**/*",       # Matches all "main" licenses in subdirs
+  "LICENSES/LICENSES_python_bindings/**/**/*"     # Matches all licenses from "thirdparty" sub-subdirs (unfortunatelly, the ** is not recursive here)
 ]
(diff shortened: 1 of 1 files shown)


FlorianReimold:
Ready for review.

Current state:
- License information can be collected with a script. The script is not executed automatically. The generated output has to be committed 
- Some external licenses (-> not from submodules) can be pre-collected with another script. That script requires the Qt and Pcap++ Sources to be available.
- A "licenses" and a "licenses_python_binding" dir is populated in the repo root
- The python wheel now contains all thirdparty licenses from the licenses_python_binding dir, which should be everything that is being used

FlorianReimold:
> @FlorianReimold Quite impressive!

Thanks!


> Could we instead have a licenses folder, and subfolders `collector`, `complete`, `python` / `python_whl` ?

Sure, that is possible. However, it is pretty standard to have a `LICENSE` (in capitals!) dir in the project root containing all licenses, that's why I did it this way. But we can restructure it to a not-so-common structure; the subdirs in the `LICENSE` dir are already uncommon.

> Also is it really necessary to have all these files checked in and not generated? What if I am doing a build with non-submodules for example.

The reason you gave is the exact reason why those files need to be checked in. If you didn't check out submodules / build from fat-source etc. you may not be able to retrieve license information anymore.

>  Don't the incorrect license files get included then?

You mean that too much license information are included? Yes, probably. But that way around it doesn't hurt. I didn't re-implement a dependency management, only a very rough filtering. If you have non-standard builds and need to ship licenses, you may have to go over them individually and select which license you need. The license dir tries to cover the attribution and license shipping by just including everything.

PR #2364, opened 2025-09-05
[core] Make State variables atomic.

### LINKED ISSUE #2262: eCAL::Process::SetState() is not threadsafe
E.g. internal variables are not threadsafe

Image

They might be read when getting process information. E.g. setting / reading the variables should be protected by a mutex / lock.

Peguen:
### Description
- making enums atomic (g_process_severity, g_process_severity_level)
- introducing handling for strings to be handled atomic in a separate header (g_process_info)
    - Pro: no lock for reading
    - Con: Setting always generates a new string
    - Open for discussion (different approach: mutexes)

### Related issues
- Fixes #2262


Peguen:
Makes sense, I will close this PR and open another one for the changes.

PR #2412, opened 2025-11-06
[Samples] Fixed compatibility with Qt 5.15

FlorianReimold:
Replaced QTimer::Start(chrono::milliseconds) with QTimer::Start(int) call in RecClientServiceGUI


KerstinKeller (approved):
:+1:

PR #2458, opened 2026-01-08
[Core Services] Callbacks are now executed in separate threads

FlorianReimold:
Important service Callbacks are now executed in separate threads in order to not block the net-code. Threads are only created when needed, are stored in a dynamic thread pool and are re-used for latter callback executions.

This fixes service deadlock issues that would occur when the user blocked service callbacks with blocking code.

Callbacks that are executed from this dynamic threadpool are:
- Server service callbacks (i.e. the user-provided code that actually defines what the service does)
- Client service-call callbacks (i.e. the user provided code that gets executed upon service-call-completion or errors)

Event callbacks are still being executed synchronously in the netcode.

### DIFF
--- a/ecal/core/src/service/ecal_service_singleton_manager.cpp
+++ b/ecal/core/src/service/ecal_service_singleton_manager.cpp
@@ -19,6 +19,12 @@
 */
 
 #include "ecal_service_singleton_manager.h"
+#include "dynamic_threadpool/dynamic_threadpool.h"
+#include "ecal_service/client_manager.h"
+#include "ecal_service/logger.h"
+#include "ecal_service/server_manager.h"
+
+#include <asio.hpp>
 
 #include <cstddef>
 #include <ecal/log.h>
--- a/ecal/dynamic_threadpool/dynamic_threadpool/include/dynamic_threadpool/dynamic_threadpool.h
+++ b/ecal/dynamic_threadpool/dynamic_threadpool/include/dynamic_threadpool/dynamic_threadpool.h
@@ -76,13 +76,13 @@ public:
    * @brief Get the current number of threads in the pool.
    * @return The number of threads in the pool.
    */
-  size_t GetSize() const;
+  [[nodiscard]] size_t GetSize() const;
 
   /**
     * @brief Get the current number of idle threads in the pool.
     * @return The number of idle threads in the pool.
     */
-  size_t GetIdleCount() const;
+  [[nodiscard]] size_t GetIdleCount() const;
 
   /**
    * @brief Shuts down the thread pool, allowing all queued tasks to finish.
--- a/ecal/dynamic_threadpool/dynamic_threadpool/src/dynamic_threadpool_impl.cpp
+++ b/ecal/dynamic_threadpool/dynamic_threadpool/src/dynamic_threadpool_impl.cpp
@@ -25,8 +25,7 @@
 #include <mutex>
 #include <thread>
 
-DynamicThreadPoolImpl::DynamicThreadPoolImpl()
-{}
+DynamicThreadPoolImpl::DynamicThreadPoolImpl() = default;
 
 DynamicThreadPoolImpl::DynamicThreadPoolImpl(size_t max_size_)
   : max_size(max_size_)
--- a/ecal/dynamic_threadpool/tests/src/atomic_signalable.h
+++ b/ecal/dynamic_threadpool/tests/src/atomic_signalable.h
@@ -17,7 +17,6 @@
  * ========================= eCAL LICENSE =================================
 */
 
-#include <atomic>
 #include <chrono>
 #include <condition_variable>
 #include <mutex>
--- a/ecal/core/src/service/ecal_service_client_impl.cpp
+++ b/ecal/core/src/service/ecal_service_client_impl.cpp
@@ -43,6 +43,7 @@
 
 #include <chrono>
 #include <cstdint>
+#include <functional>
 #include <memory>
 #include <mutex>
 #include <string>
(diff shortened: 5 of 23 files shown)


ecal/dynamic_threadpool/dynamic_threadpool/src/dynamic_threadpool_impl.cpp
```diff
@@ -0,0 +1,132 @@
...
+  // Add the task to the queue
+  tasks.push(task);
+
+  // If there are idle workers, notify one of them
+  if (idle_count > 0)
+  {
+    cv.notify_one();
+  }
+  // If no idle workers and we can create more, spawn a new worker
+  else if (max_size <= 0 || (workers.size() < max_size))
```
KerstinKeller: We can now use C++ 17 😄 I think this could be a great usage for a `std::optional` 

```
std::optional<std::size_t> max_size = std::nullopt; // no limit
// or
std::optional<std::size_t> max_size = 8;            // limit is 8

else if (!max_size || workers.size() < *max_size)
// or
else if(!max_size.has_value() || workers.size() < max_size.value())
```

ecal/dynamic_threadpool/tests/src/atomic_signalable.h
```diff
@@ -0,0 +1,207 @@
+/* ========================= eCAL LICENSE =================================
```
KerstinKeller: This is a copy from somewhere else in eCAL?
Probably it would be very worth it having a "testing" library, which can be used throughout the tests, we have various common testing functionality which would be worth sharing

ecal/dynamic_threadpool/dynamic_threadpool/include/dynamic_threadpool/dynamic_threadpool.h
```diff
@@ -0,0 +1,106 @@
...
+   * available, or create a new thread if the maximum size is not reached, yet.
+   * If the maximum size is reached and all threads are busy, the task will be
+   * queued
+   *
+   * If shutdown has been called on the thread pool, posting a new task will fail.
+   *
+   * @param task_ The task to be executed by the thread pool.
+   * @return True if the task was successfully posted; otherwise, false.
+   */
+  bool Post(const std::function<void()>& task_);
```
KerstinKeller: shouldn't `Post` also be ``[[nodiscard]]` ? Otherwise probably nobody will check if the task has actually been posted.

ecal/dynamic_threadpool/tests/src/dynamic_threadpool_test.cpp
```diff
@@ -0,0 +1,280 @@
...
+    EXPECT_EQ(thread_pool->GetIdleCount(), 0);
+    thread_pool.reset();
+  }
+}
+
+#if 1
+// Create a theadpool and let it execute a task
+TEST(DynamicThreadPool, BasicExecution)
+{
+  DynamicThreadPool thread_pool;
```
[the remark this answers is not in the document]
KerstinKeller: Is github-actions incorrect here? First I thought `Post` might be const, but it isn't.

ecal/dynamic_threadpool/tests/src/dynamic_threadpool_test.cpp
```diff
@@ -0,0 +1,280 @@
...
+  // Create and delete an empty thread pool
+  {
+    auto thread_pool = std::make_unique<DynamicThreadPool>();
+
+    EXPECT_EQ(thread_pool->GetSize(), 0);
+    EXPECT_EQ(thread_pool->GetIdleCount(), 0);
+
+    thread_pool.reset();
+  }
+
```
KerstinKeller: These two scopes are totally independent tests, so I would actually make them two independent testcases 😄

ecal/dynamic_threadpool/tests/src/dynamic_threadpool_test.cpp
```diff
@@ -0,0 +1,280 @@
...
+    thread_pool.Post([&finished_tasks, delay]()
+                      {
+                        std::this_thread::sleep_for(delay);
+                        finished_tasks++;
+                      });
+
+  }
+
+  // Wait a short time, so that the tasks have not finished yet
+  std::this_thread::sleep_for(delay / 2);
```
KerstinKeller: I wonder if there is a different way to check this without sleeping for fixed delays, which also increases total execution time.

I don't know if this is checked in, but for test purposes I implemented synchronization points. So you can synchronize all the threads and the main thread. There you know that everyone is still executing, the main thread can check everything and then you release the execution again.

This way, the test will still execute, but execution time can be way faster, which is only 500ms here, but can accumulate over time.

KerstinKeller (commented):
The threadpool is a great improvement 👍 

Even though it's already merged, i do have a few remarks:
- Please check the license headers, they contain the wrong year.
- One recommendation using a std::optional for max_size
- Some recommendation for the tests, to speed up execution time and check used thread ids instead of using heuristic to check which threads have been used.

I didn't review the actual usage in ecal_service.

ecal/core/src/service/ecal_service_singleton_manager.cpp
```diff
@@ -1,6 +1,7 @@
 /* ========================= eCAL LICENSE =================================
  *
  * Copyright (C) 2016 - 2025 Continental Corporation
+ * Copyright 2025 AUMOVIO and subsidiaries. All rights reserved.
```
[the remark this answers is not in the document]
FlorianReimold: Well it was developed in 2025, so I put 2025 there.

ecal/dynamic_threadpool/dynamic_threadpool/src/dynamic_threadpool_impl.cpp
FlorianReimold: Yes, using a C++17 optional would work, too. But I don't really see the advantage here over having a plain integer type.
The `<=` includes a `==`, so having that is fine and in this case even equivalent

PR #2462, opened 2026-01-13
[core] Handle globals accessor in globals class

Peguen:
### Description
Removed the creation of globals shared ptr within the globals class.


KerstinKeller (approved):
This makes a whole lot of sense 👍

PR #2471, opened 2026-01-20
[Core / Config] Expose Destructor / Copy / Move

FlorianReimold:
Explicitly declared and exported Destructor / Copy / Move  in the header file and moved the default implementation to the cpp file

It is usually advised to explicitly implement Destructor / Copy / Move for library APIs, as re-generating those functions implicitly by the consuming compiler can lead to ABI issues.

### DIFF
--- a/ecal/core/include/ecal/config/configuration.h
+++ b/ecal/core/include/ecal/config/configuration.h
@@ -65,6 +65,13 @@ namespace eCAL
 
     ECAL_API Configuration();
 
+    ECAL_API ~Configuration();
+
+    ECAL_API Configuration(const Configuration&);
+    ECAL_API Configuration(Configuration&&) noexcept;
+    ECAL_API Configuration& operator=(const Configuration&);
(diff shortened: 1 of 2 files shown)


FlorianReimold:
> I am unsure if that really fixes the problems, but maybe it does. We still have all substructures defined in the header only (TransportLayer::Configuration, ...). So we still have a mix of usage in the library and in the user code.

We probably have, as this PR only fixes one of those occurrences. However, your specific example is header-only, so it is 100% compiled by the consuming compiler anyways.
This fixed code however initially had the Constructor compiled into the library and the Destructor compiled by the the consumer.

> What I would do, is to make this structure header only, and have `InitFromConfig()` and `InitFromFile()` be factory functions. (However, which we cannot do within 6.0, only deprecate).

This fix is for 6.1 anyways, as it may break ABI compatibility. Providing factory functions is a differnt topic. They don't address ABI issues between pre-compiled constructors and consumer-compiled destructors.

PR #2602, opened 2026-04-07
[config] Update config generator

Peguen:
### Description

Adds two new CLI flags to `ecal_generate_config` and modernizes the code to C++17.

## New features

**`--overwrite` / `-o`**
Silently overwrites existing `ecal.yaml` / `ecaltime.yaml` without prompting. Useful for automated/scripted setups.

**Interactive overwrite confirmation**
When existing config files are detected and `--overwrite` is not given, the tool lists the affected files and asks before proceeding:
```
The following configuration file(s) already exist:
  /path/to/ecal.yaml
Overwrite? [y/N]:
```

**`--no-wait` / `-n`**
Suppresses the `Press Enter to continue...` pause on exit, making the tool safe to call from scripts.

### DIFF
--- a/ecal/core/cfg/generate_configuration_yaml.cpp
+++ b/ecal/core/cfg/generate_configuration_yaml.cpp
@@ -1,7 +1,8 @@
 /* ========================= eCAL LICENSE =================================
  *
  * Copyright (C) 2016 - 2025 Continental Corporation
- *
+ * Copyright 2026 AUMOVIO and subsidiaries. All rights reserved.
+ * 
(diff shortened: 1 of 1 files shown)


Peguen:
> ```
> The following configuration file(s) already exist:
>   ecal.yaml
> Overwrite? [y/N]:
> ```
> 
> for the following usecase, should we print the complete filepath?

Yes, it will print the whole path. I will add it in the description. Only in the case of -d (dumping in the cwd) it does not show the whole path. Shall it?

KerstinKeller (approved):

PR #2608, opened 2026-04-10
[core] Refactor `ecal_path_processing.cpp` to C++17 `std::filesystem`

Peguen:
### Description
Replaces the custom `EcalUtils::Filesystem` / `EcalUtils::String` wrappers in `ecal_path_processing.cpp` with standard C++17 `<filesystem>` equivalents and fixes several bugs found during the review.

### DIFF
--- a/ecal/core/src/config/ecal_path_processing.cpp
+++ b/ecal/core/src/config/ecal_path_processing.cpp
@@ -24,9 +24,10 @@
 #include "ecal_path_processing.h"
 
 #include "ecal_def.h"
-#include "ecal_utils/ecal_utils.h"
-#include "ecal_utils/filesystem.h"
 #include "ecal/config.h"
+
+#include <filesystem>
+#include <system_error>
 #include "ecal/util.h"
 #include "util/getenvvar.h"
 
@@ -47,21 +48,17 @@
 
 namespace
 {
-  // get the path separator from the current OS (win: "\\", unix: "/")
-  const std::string path_separator(1, EcalUtils::Filesystem::NativeSeparator());
-
-  // returns empty if str1_ is empty. otherwise returns str1_ + path_separator + str2_
+  // returns empty if str1_ is empty. otherwise returns str1_ / str2_ (native path separator)
   std::string buildPath(const std::string& str1_, const std::string& str2_)
   {
     if (str1_.empty()) return "";
-    return EcalUtils::String::Join(path_separator, std::vector<std::string>{str1_, str2_});
+    return (std::filesystem::path(str1_) / str2_).string();
   }
 
   // checks if the specified file is a proper file
   bool isValidFile(const std::string& full_file_path_)
   {
-    const EcalUtils::Filesystem::FileStatus file_status(full_file_path_, EcalUtils::Filesystem::Current);
-    return file_status.IsOk() && (file_status.GetType() == EcalUtils::Filesystem::Type::RegularFile);
+    return std::filesystem::is_regular_file(full_file_path_);
   }
 
 #ifdef ECAL_OS_WINDOWS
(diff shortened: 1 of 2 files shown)


hannemn (commented):
@FlorianReimold Can you give the PR a short review please? Thanks. :-)

ecal/core/src/config/ecal_path_processing.cpp
```diff
@@ -226,34 +221,36 @@ namespace eCAL
...
+      const std::string test_file_name = "ecal_write_test_" + std::to_string(std::hash<std::thread::id>{}(std::this_thread::get_id())) + ".tmp";
+      const std::filesystem::path test_file_path = std::filesystem::path(path_) / test_file_name;
+      std::ofstream test_file(test_file_path);
+
+      if (test_file)
       {
-        testFile.close();
-        std::remove(testFilePath.c_str());
+        test_file.close();
+        std::filesystem::remove(test_file_path);
```
FlorianReimold: I feel like exceptions should be caught or the error_code should be passed, so that a failure while removing the dummy test file does not crash the application.

ecal/core/src/config/ecal_path_processing.cpp
```diff
@@ -265,7 +262,7 @@ namespace eCAL
     #ifdef ECAL_OS_WINDOWS
       
       char unique_path[MAX_PATH];
-      if (!GetTempFileNameA(tmp_dir.c_str(), "ecal", 0, unique_path) != 0)
+      if (GetTempFileNameA(tmp_dir.c_str(), "ecal", 0, unique_path) == 0)
```
FlorianReimold: Using hardcoded ANSI functions it probably is fine in 99.9% of all cases, as the tmp dir will most likely only consist of 7-bit ASCII characters anyways. However, it it doesn't, this will fail, as the ANSI output is later interpreted as UTF-8 (I think). 
Thus, the proper way would be to use the hardcoded Unicode equivalent and convert that to UTF-8.

I found hardcoded ANSI functions in the following functions:

`getTempDir`
- `GetTempPathA`

`DirProvider::uniqueTmpDir`
- `GetTempFileNameA`
- `DeleteFileA`
- `CreateDirectoryA`

FlorianReimold (changes requested):
This is my first review. I will perform another review regarding Unicode correctness. Please stay tuned 😉

FlorianReimold:
Ok, so I did a small test and come to the conclusion that the current implementation does not handle Unicode paths in Windows.

Take the following code for example:

https://github.com/eclipse-ecal/ecal/blob/92f997304a80b564ca89cab2019224c19a62b82f/ecal/core/src/config/ecal_path_processing.cpp#L356-L360

When any eCAL App is being started from a path with unicode characters that don't appear in the currently configure codepage (in my test `D:\testdir\Beer🍺`), the current working directory cannot be represented with 8 bit characters (_Which would be ANSI on Windows so depending on the Codepage! But those legacy codepages don't have emojies._) and therefore `.string()` throws an exception internally and returns an empty string.
The solution for this problem is to call `.wstring()` on windows to get the UTF-16 path as a string and convert it to UTF-8 for passing it through the code.

Therefore, a deeper analysis has to be done to replace all non-widestring calls with widestring equivalents on windows.

FlorianReimold:
I conducted a manual review of the PR changes and didn't find any issues. I think, it should now handle Unicode data properly. I will however continue with a small review of the non-changed codebase as well so see if the surrounding code handles it properly.

FlorianReimold (approved):
OK, looks good for me. Widestring and UTF 8 was used consistently everywhere I checked. An AI review told me the same. The AI review mentioned that we are currently not supporting long file paths for the temp directory (it uses `MAX_PATH`) and that there are ways to also support long paths there. But I think the current state is fine.

PR #2625, opened 2026-05-12
Extract CI failure summary into reusable workflow

FlorianReimold:


.github/workflows/build-macos.yml
```diff
@@ -1,10 +1,7 @@
 name: Build macOS
 
-on:  
-  push:
-  pull_request:
-    branches:
-      - master
+on:
+  workflow_dispatch:
```
KerstinKeller, line 4: This means that GH actions for macos will run only on demand. Do we want this?

.github/workflows/build-windows.yml
```diff
@@ -1,10 +1,7 @@
 name: Build Windows
 
 on:
-  push:
-  pull_request:
-    branches:
-      - master
+  workflow_dispatch:
```
KerstinKeller, line 4: We definately don't want this here

.github/workflows/build_wheels.yml
```diff
@@ -1,8 +1,6 @@
 name: Build Wheels
 
 on:
-  push:
-  pull_request:
   workflow_dispatch:
```
KerstinKeller, line 4: let's revert this

KerstinKeller (commented):
In addition to cleaning up the failed test cases we for sure need to reactivate the workflows properly.

hannemn:
Will be handled #2632

PR #2630, opened 2026-05-20
Feature/tracing v1.0

KerstinKeller:
### Related issues

-


ecal/core/src/tracing/tracing.h
```diff
@@ -0,0 +1,129 @@
...
+      topic_direction  direction;       // publisher or subscriber
+    };
+
+    struct SPublisherSpanData
+    {
+      operation_type op_type;
+      uint64_t       entity_id;
+      uint64_t       process_id;
+      size_t         payload_size;
+      long long      clock;
```
KerstinKeller: why are we using long long here?

ecal/core/src/tracing/tracing.h
```diff
@@ -0,0 +1,129 @@
...
+    struct SSubscriberSpanData
+    {
+      operation_type op_type;
+      uint64_t       entity_id;
+      uint64_t       topic_id;
+      uint64_t       process_id;
+      size_t         payload_size;
+      long long      clock;
+      uint64_t       layer;
+      long long      start_ns;       // start timestamp in nanoseconds
```
KerstinKeller: Would make more sense to provide start & duration (start in synchronized eCAL Time). instead of start & end.

PR #2632, opened 2026-05-26
CI: restore default workflow triggers and fix windows reusable-summary reference

YannikSpaeth:
### Description
Reverts temporary CI trigger restrictions and fixes the Windows reusable summary workflow reference.


FlorianReimold (changes requested):
Currently I see 2 remaining issues with this PR:

1. The action fails for PRs that come from forks. This PR is a good example by itself for that. The reduces permissions of the github token for fork PRs are by design and should not be changed. However, the action should also not go to a failed state for every fork PR. An acceptable solution would be to simply skip the action, as it cannot post comments anyways.

2. The action is triggered by pushes (i.e. triggers that are non-PR related), but don’t react on that. It always tries to obtain the PR number and operate with it. I am not sure what happens in that case (fail / PR number 0 / maybe it is handles somewhere else?)

PR #2673, opened 2026-07-07
[CMake eCALConfig] only find protobuf if eCAL was configured with protobuf

### LINKED ISSUE #2636: With -DECAL_USE_PROTOBUF=OFF, find_package(eCAL..) in downstream consumers still fail due to test for protobuf
### Problem Description

Although eCAL version 6.1.1 allows building `core`  preset without protobuf, CMake projects that consume eCAL still fail on it because `find_package(eCAL...)` calls `find_package(Protobuf...)`. The attached patch fixes it for me, but I am not sure it's the correct fix in general.

[ecal-6.1.1.patch](https://github.com/user-attachments/files/28916398/ecal-6.1.1.patch)

### How to reproduce

- configure with `cmake --preset core -DECAL_USE_PROTOBUF=OFF`, then build and install eCAL
- Have a test application's CMakeLists.txt call `find_package(eCAL REQUIRED)`. 
- Notice that this fails if protobuf is not installed system-wide

### How did you get eCAL?

### Environment

- eCAL 6.1.1
- OS: Linux (ubuntu 24.04)
- CMake 4.3.3
- gcc-16.0.1

### eCAL System Information

```text

```

FlorianReimold:
Fixes #2636

### DIFF
--- a/cmake/eCALConfig.cmake.in
+++ b/cmake/eCALConfig.cmake.in
@@ -30,12 +30,14 @@ set(eCAL_VERSION_STRING @eCAL_VERSION_STRING@)
 set(CMAKE_MAP_IMPORTED_CONFIG_MINSIZEREL Release "")
 set(CMAKE_MAP_IMPORTED_CONFIG_RELWITHDEBINFO Release "")
 
-# Try and find protobuf in Config mode first
-# (Primarily for Windows where we provide protobuf as part of eCAL)
-find_package(Protobuf CONFIG)
-if(NOT Protobuf_FOUND)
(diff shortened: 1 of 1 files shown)


KerstinKeller (approved):
This will work, if you try to consume a version which was *not* build against protobuf.
Ideally, we tie this to required components instead. This will require a bit of refactoring and might be done for 6.2.

The idea is, that a user requests:
``` 
find_package(eCAL REQUIRED COMPONENTS core proto)
```
and then depending on the components requested, protobuf is found and ecal proto targets are included.
In order for this to work properly, we should export targets into different export sets, and export those sets into different files, so we can include them depending on which component has been requested.

PR #2679, opened 2026-07-08
Migrate project from C++17 to C++20

bastienstefani:
Build eCAL with C++20 by default and fix the sources that relied on behavior that was broken/removed in C++20:

- CMakeLists.txt: default CMAKE_CXX_STANDARD to 20 (still overridable from the command line or a toolchain file)
- ecal_path_processing.cpp / path_processing_test.cpp: in C++20 std::filesystem::path::u8string() returns std::u8string (char8_t) instead of std::string, and std::filesystem::u8path() is deprecated. Convert via iterator ranges / char8_t-aware construction that works with both C++17 and C++20.
- InterruptibleThread.h, interruptible_thread.h, ecalsys_gui.cpp, group_edit_widget.cpp, group_widget.cpp: replace implicit 'this' capture via [=] (deprecated in C++20) with explicit [this] / [this, action] captures.
- build-macos.yml: build with -DCMAKE_CXX_STANDARD=20
- doc: update compiler requirements and CMake sample to C++20

Claude-Session: https://claude.ai/code/session_01BYnx5rJRJZiJYe5BWmkwQ3

### Related issues

-


KerstinKeller (commented):
@bastienstefani is there a reason you closed the PR again? While we will still have to keep eCAL compilable with C++17, I have no objections to make it compatible with newer standards, just not as default.
The other changes ([=] -> [this]) and the path handling seem valid.

bastienstefani:
@KerstinKeller to be honest that was an error, i made Claude make the changes in my fork, and he created also this PR. I haven't tested it yet so didn't wanted to create  a PR with untested code.  My apologies for this.

On a side note, why "again" ? i closed it only once right after it was opened.

PR #2680, opened 2026-07-09
[core] Gate publisher sends by active connection layer

KerstinKeller:
This change tracks publisher connections and selected transport layers using atomic counters and explicit connection states. Writes now skip transport layers without active subscribers, reducing unnecessary data copies and network traffic while preserving zero-copy SHM where possible. Connection counts are updated incrementally instead of recalculated from the connection map. Transport counters are reset when writers are destroyed.

### DIFF
--- a/ecal/core/src/pubsub/ecal_publisher_impl.cpp
+++ b/ecal/core/src/pubsub/ecal_publisher_impl.cpp
@@ -104,6 +104,64 @@ namespace
 
 namespace eCAL
 {
+  void CPublisherImpl::SSendLayerConnectionCounters::Increment(TransportLayer::eType layer_)
+  {
+    switch (layer_)
+    {
+    case TransportLayer::eType::udp_mc:
+      udp.fetch_add(1, std::memory_order_relaxed);
+      break;
+    case TransportLayer::eType::shm:
+      shm.fetch_add(1, std::memory_order_relaxed);
+      break;
+    case TransportLayer::eType::tcp:
+      tcp.fetch_add(1, std::memory_order_relaxed);
+      break;
+    default:
+      break;
+    }
+  }
+
+  void CPublisherImpl::SSendLayerConnectionCounters::Decrement(TransportLayer::eType layer_)
+  {
+    switch (layer_)
+    {
+    case TransportLayer::eType::udp_mc:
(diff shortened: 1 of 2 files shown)


FlorianReimold (approved):

ecal/core/src/pubsub/ecal_publisher_impl.h
```diff
@@ -150,12 +151,16 @@ namespace eCAL
...
       SDataTypeInformation data_type_info;
       SLayerStates         layer_states;
+      TransportLayer::eType selected_layer = TransportLayer::eType::none;
       bool                 state = false;
     };
     using SSubscriptionMapT = std::map<SSubscriptionInfo, SConnection>;
     mutable std::mutex                     m_connection_map_mutex;
     SSubscriptionMapT                      m_connection_map;
     std::atomic<size_t>                    m_connection_count{ 0 };
+    std::atomic<bool>                      m_udp_send_enabled{ false };
```
KerstinKeller: instead of an std::atomic<bool> we should use connection counters, as we can increase / decrease them upon registration / unregistration which eliminates the need to iterate over the map every time.

KerstinKeller:
This is already better than what we had previously. However we still have a problem.
In theory we  have a state machine for a connection: `PENDING` -> `ESTABLISHED` -> `CLOSED`. After the first regisration refresh a connection is pending, after the second refresh it's currently counted as establed. This is a very bad heuristic.

With this PR, we currently won't send out data for pending connections, which is a different behavior than previously.
We should consider modeling pending conditions as well, and already send data even if the connection has not yet been confirmed.

Anyways the logic is brittle, and what we need is a true handshake, as we get e.g. in TCP, instead of a heuristic via the monitoring.

ecal/core/src/pubsub/ecal_publisher_impl.cpp
```diff
@@ -155,18 +214,22 @@ namespace eCAL
     // get payload buffer size (one time, to avoid multiple computations)
     const size_t payload_buf_size(payload_.GetSize());
 
+    const bool udp_send_enabled = m_send_layer_connection_counters.UdpEnabled();
```
KerstinKeller: already && with udp_writer.

hannemn (approved):

PR #2685, opened 2026-07-16
[GH Action] change toolchain action vendor

FlorianReimold:
The currently latest v2.0.6 action has an issue that ./index.ps1 is not found.


KerstinKeller:
@FlorianReimold  how do you feel about switching to https://github.com/step-security/msvc-dev-cmd instead? there is a company behind its maintenance, and it's free to use for public repositories.

KerstinKeller (approved):

PR #2688, opened 2026-07-23
Fix: compiler warnings

hasan-htp:
### Description
I used eCAL in a project via CMake FetchContent.  -Werror is enabled globally in my project, and I encountered several CMake warnings. This PR addresses those gcc warnings.

### DIFF
--- a/serialization/common/common/include/ecal/msg/omeasurement.h
+++ b/serialization/common/common/include/ecal/msg/omeasurement.h
@@ -55,11 +55,7 @@ namespace eCAL
         // The way we handle Publishers requires us to do a two pass serialization;
         size_t message_size = m_serializer->MessageSize(entry_.message);
         buffer.resize(message_size);
-        if (message_size == 0)
-        {
-          m_serializer->Serialize(entry_.message, nullptr, 0);
-        }
-        else
+        if (message_size > 0)
         {
           m_serializer->Serialize(entry_.message, static_cast<void*>(&buffer[0]), buffer.size());
         }
--- a/serialization/common/common/include/ecal/msg/omeasurement.h
+++ b/serialization/common/common/include/ecal/msg/omeasurement.h
@@ -28,8 +28,6 @@ namespace eCAL
     template <typename T, typename Serializer>
     class OMessageChannel
     {
-      friend OMessageChannel CreateChannel(OMeasurement& meas_, const std::string& channel_name_);
-
     public:
       // Should those be private?
       using SerializerT = Serializer;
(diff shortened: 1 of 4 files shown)


KerstinKeller:
hi @hasan-htp, thanks for opening the PR.
Some of the changes that you made are ok, but the ones in hdf5 now ignore important error conditions, and the taken approach for the serializers is also questionable.

I will check this and see what we can do!

hasan-htp:
> hi @hasan-htp, thanks for opening the PR. Some of the changes that you made are ok, but the ones in hdf5 now ignore important error conditions, and the taken approach for the serializers is also questionable.
> 
> I will check this and see what we can do!

hi @KerstinKeller
thanks for the reply and the feedback, the reason of that change is `-Wtype-limits`, because the `size>=0 `condition is always true for `size_t` type.

serialization/common/common/include/ecal/msg/omeasurement.h
```diff
@@ -28,8 +28,6 @@ namespace eCAL
     template <typename T, typename Serializer>
     class OMessageChannel
     {
-      friend OMessageChannel CreateChannel(OMeasurement& meas_, const std::string& channel_name_);
```
KerstinKeller, line 31: Why does this issue a warning? I think we need this friend declaration.

serialization/string/string/include/ecal/msg/string/serializer.h
```diff
@@ -62,7 +62,7 @@ namespace eCAL
 
         static bool Serialize(const T& msg_, void* buffer_, size_t size_)
         {
-          if (msg_.size() <= size_)
+          if (msg_.size() <= size_ && msg_.size() > 0)
```
KerstinKeller: with the addtional check in omeasurement.h, this check should not be necessary, as this function will never be called with `size_ == 0`

serialization/common/common/include/ecal/msg/omeasurement.h
hasan-htp: this function is declared in line 80 as a template function. but this friend declaration itself is not template, this is what giving the warning.
The reason of removing the friend declaration is that the function is only using the constractor of `OMessageChannel`, which is public.

or am I missing something?

serialization/string/string/include/ecal/msg/string/serializer.h
hasan-htp: oh yes, thanks, I will remove it

hasan-htp: done

KerstinKeller (approved):

PR #2702, opened 2026-08-28
[python] Include protozero

Fabien-B:
### Description
Add protozero to sdist.only-include

Python bindings won't build without it.


KerstinKeller:
@Fabien-B you're absolutely right. Thanks for creating the PR. Would you mind signing the ECA so that we can merge this?

Fabien-B:
Hello,
For such a small change, I'm kinda too lazy to create an Eclipse account :smile: .
Can you close this PR and do this change on your side ?

